PR #1, opened 2018-01-17
Add skeleton of meta and dataset modules.

tsalo:
Also, add ALE and SCALE code from [pyale](https://github.com/NBCLab/pyale). It won't work though.


nimare/meta/ibma/ibma.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+"""
+Image-based meta-analysis estimators
+"""
+from ..base import MetaEstimator
+
+
+class MFX_GLM(MetaEstimator):
```
tyarkoni: Will probably inherit from `IBMAEstimator`

nimare/meta/ibma/ibma.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    The gold standard image-based meta-analytic test. Uses contrast and standard
+    error images.
+    """
+    def __init__(self, dataset, n_iters=10000, voxel_thresh=0.001, clust_thresh=0.05,
```
tyarkoni: I think we'll almost certainly need another layer on top of `dataset` that represents the actual component of a database/dataset that can be submitted for analysis. So I still lean strongly towards making `dataset` an argument of `fit` and not `init`. Otherwise we'll end up in a weird situation where both the `Dataset` and the `MetaEstimator` classes are implementing methods that filter/mutate the data.

nimare/dataset/dataset.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def has_data(self, dat_str):
+        dat_str = dat_str.split(' AND ')
+        for ds in dat_str:
+            try:
+                self.data.get(None)
+            except:
+                raise Exception('Nope')
+
+    def get(self, search='', algorithm=None):
```
tyarkoni: Given the scope of the project and the variety of different data types this object is going to store, I think we may want more explicit accessor methods (e.g., `get_studies`, `get_metadata`, etc.). Fine to leave this as a placeholder, but ultimately we'll probably replace it with a bunch of more specific methods.

nimare/meta/cbma/mkda.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+Coordinate-based meta-analysis estimators
+"""
+import warnings
+
+from ..base import MetaEstimator
+from ...utils import intersection, diff
+
+
+class MKDA(MetaEstimator):
```
tyarkoni: Should inherit from `CBMAEstimator`

nimare/dataset/dataset.py
tsalo: I'll add in some more empty methods, just so there's a blueprint.

tyarkoni:
I don't think I can push commits to your repo, so I'll merge as-is and then create a new branch with edits.

PR #4, opened 2018-03-08
Skeleton expansion

tyarkoni:
Adds some (suggested) commits to #2. @tsalo sorry for opening another PR to master instead of to your branch, but I'm short on time and didn't want to spend time fiddling with git. If these changes look good to you, maybe close #2 and we can work from here.


tsalo:
Everything looks good and I like the `extract` idea. The next thing I was planning to do was add in a new submodule in `meta.cbma` for modeling activation maps from coordinates, since I know the goal is to keep that separate from the meta-analytic algorithms. I was thinking of naming it `kernel`, with a base `KernelEstimator` and algorithm-specific estimators (`ALEKernel`, `MKDAKernel`, and `KDAKernel`, for now). Once that's working we'll at least be able to work with the test dataset I generated from 10 studies in Neurosynth, although that may need to be adjusted.

Once I get that working, should I open a PR to skeleton-expansion?

nimare/dataset/dataset.py
```diff
@@ -88,3 +222,47 @@ def get_images(self):
 
     def get_coordinates(self):
         pass
+
+    def save(self, filename):
```
tsalo: Addressing concerns raised in #2, I replaced the filename-based check with a `compress` argument in my new PR (#5).

nimare/dataset/dataset.py
```diff
@@ -88,3 +222,47 @@ def get_images(self):
...
+        """
+        if filename.endswith('z'):
+            with gzip.GzipFile(filename, 'wb') as file_object:
+                pickle.dump(self, file_object)
+        else:
+            with open(filename, 'wb') as file_object:
+                pickle.dump(self, file_object)
+
+    @classmethod
+    def load(cls, filename):
```
tsalo: Addressing concerns raised in #2, I replaced the filename-based check with a `compressed` argument in my new PR (#5).

nimare/decode/discrete.py
```diff
@@ -0,0 +1,44 @@
...
+Methods for decoding subsets of voxels (e.g., ROIs) or experiments (e.g., from
+meta-analytic clustering on a database) into text.
+"""
+from .base import Decoder
+from ...due import due, Doi
+
+
+@due.dcite(Doi('10.1371/journal.pcbi.1005649'),
+           description='Citation for GCLDA decoding.')
+class GCLDADiscreteDecoder(Decoder):
```
tsalo: Is it worth it for me to try to fit GCLDA (including the dataset and model modules) into NiMARE or should I just try to make GCLDA NiMARE-compatible after NiMARE is functional?

nimare/dataset/base.py
```diff
@@ -1,64 +1,84 @@
...
+        self.connectomes = []
+        self.surfaces = []
+
+        if images:
+            self.add_images(images)
+
+        if not isinstance(activations, ActivationSet) and activations is not None:
+            raise ValueError('Input activations must be ActivationSet or None')
+        else:
+            self.coordinates = activations
```
tsalo: Addressing a concern raised in #2, I switched to using `coordinates` and `CoordinateSet` in my fork for consistency.

nimare/dataset/dataset.py
```diff
@@ -1,68 +1,187 @@
...
+    for i in xrange(0, len(l), n):
+        yield l[i:i + n]
 
-class Contrast(object):
-    """Container for contrasts (aka experiments or comparisons) nested within
-    studies.
-    Should store an arbitrary number of ConnMatrices, Images, and Surfaces,
-    along with at most one ActivationSet.
+
+def download_combined_database(out_dir, overwrite=False):
```
tsalo: Addressing an issue brought up in #2, I moved `download_combined_database` into `nimare.dataset.extract` in my fork. I can move it into something like `nimare.dataset.download`, but the related downloading functions are now in `extract`, so I figured this function belonged there too.

nimare/dataset/dataset.py
```diff
@@ -1,68 +1,187 @@
...
+    Parameters
+    ----------
+    dataset_file : :obj:`str`
+        Json file containing dictionary with database information.
+    """
+    def __init__(self, database_file):
+        with open(database_file, 'r') as fo:
+            self.data = json.load(fo)
+
+    def get_dataset(self, search='', algorithm=None, target=None):
```
tsalo: If we push transforming images and, more importantly, coordinates until Dataset generation, then I don't think we can do searches using ROIs in the Database. At least the way I've been thinking about it, generally speaking one would only transform a subset of a Database into a Dataset, based on some criteria (e.g., necessary data for a specific estimator, tagged with some CogPO or CogAt label, etc.), which means searching the Database to get a list of Contrast IDs before feeding those into a `get_dataset` method (or something similar). However, without the coordinates all in the same space as some ROI, we can't search based on their locations (unless we transform on the fly in the search method, which I think would be wasteful). I'm fine with allowing metadata-based searching within Databases, but restricting location-based searching to Datasets, but it's probably something worth discussing.

PR #5, opened 2018-03-12
Add CBMA kernels and move kernel calls into meta-analysis calls.

tsalo:
We want to keep the CBMA activation modeling step (e.g., convolving coordinates with a Gaussian or a binary sphere) separate from the meta-analysis, so that users can use whichever combination they prefer. However, keeping them completely separate made the pipeline (as I interpreted it) very complicated, since some CBMA algorithms (e.g., ALE and SCALE) require access to both the dataset and the kernel estimator within the analysis.

My solution is to use the kernel estimator as an argument for the CBMA meta-analytic methods. Arguments for the kernel estimator (e.g., `r` for `MKDAKernel`) can be included as keyword arguments to the meta-analysis with the prefix `kernel__`. I think `sklearn` does something similar with its pipelines.


tsalo:
Here's how I was envisioning the script for example ALE and MKDA meta-analyses on the `AX-CPT` task:

```python

import nimare

database = nimare.dataset.download_combined_database()

## ALE
# get_studies should return studies with coordinates and sample sizes
all_ids = database.get_studies(algorithm=nimare.meta.cbma.ALE)

# dataset contains both task-positive and task-negative studies matching
# algorithm's criteria in target space
dataset = nimare.dataset.Dataset(database, all_ids, target='MNI152_2mm')
task_pos_ids = dataset.get(cogatlas='trm_4c40d2a93ea15')  # AX-CPT
meta = nimare.meta.cbma.ALE(dataset, ids=task_pos_ids,
                            kernel_estimator=nimare.meta.cbma.ALEKernel)
meta.fit(n_iters=1000, corr='FWE', voxel_thresh=0.001, clust_thresh=0.05)
meta.save_results(prefix='/out/dir/axcpt_ale_')

## MKDA
all_ids = database.get_studies(algorithm=nimare.meta.cbma.MKDA)
dataset = nimare.dataset.Dataset(all_ids)
task_pos_ids = dataset.get(cogatlas='trm_4c40d2a93ea15')  # AX-CPT
task_neg_ids = sorted(list(set(all_ids) - set(task_pos_ids)))
meta = nimare.meta.cbma.MKDA(dataset, ids=task_pos_ids, ids2=task_neg_ids,
                             kernel_estimator=nimare.meta.cbma.MKDAKernel,
                             kernel__r=6)
meta.fit(n_iters=1000, corr='FWE', voxel_thresh=0.001, clust_thresh=0.05)
meta.save_results(prefix='/out/dir/axcpt_mkda_')

```

tsalo:
The CBMA methods require a dataset of some kind, but the IBMA methods could take in lists of images or 2D contrast-by-voxel arrays. It seems likely that users will want to be able to run image-based meta-analyses without initializing a database/dataset. I was thinking it might be a good idea to have a way to do it with the simplified inputs, but I'm not sure what the best way is. Would a `fit_arrays` class method that takes in arrays and returns a `MetaResults` object make any sense?

tsalo:
Closing this to merge directly into master.

Issue #8, opened 2018-04-03
Meta-analytic algorithms


tsalo:
Here's a list of the methods we plan to support and their current statuses in my branch:

- Kernel-based coordinate-based meta-analyses:
  - [X] ALE
  - [X] SCALE
  - [X] MKDA Density Analysis
  - [X] MKDA Chi2 Analysis (Neurosynth speed-optimized version)
  - [X] MKDA Chi2 Analysis with empirical null
  - [X] KDA
- Model-based coordinate-based meta-analyses:
  - [ ] [BHICP](http://www.nisox.org/Software/BHICP/) Available implementation only in C++
  - [ ] [HPGRF/BHPGM](http://www.nisox.org/Software/BHPGM/) Available implementation only in C++
  - [ ]  [SBLFR](https://github.com/silphimon/SBLFRM-Coordinate-based-Meta-analysis)
  - [X] ~~SBR~~ Cannot currently be run on a full brain. No source code available.
- Image-based meta-analyses:
  - [X] Fisher's
  - [X] Stouffer's FFX
  - [X] Stouffer's RFX with theoretical null
  - [X] Stouffer's RFX with empirical null
  - [X] Weighted Stouffer's
  - [X] FFX GLM
  - [X] RFX GLM with theoretical null
  - [X] RFX GLM with empirical null
  - [X] MFX GLM

tyarkoni:
Nice! Sorry I haven't had time to keep up with all the changes; will try to get to it later this week.

tyarkoni:
Also looping @nicholst in in case he can think of other methods that are relatively straightforward to implement and are missing from the list. Tom, NiMARE is the better-named successor to the "PyCIBMA" package we proposed in our grant. @tsalo is a grad student working with Angie who's taking the lead on developing this.

tsalo:
[...]

PR #12, opened 2018-05-06
Add tail argument to IBMAs and use FSL for FFX GLM

tsalo:
Based on discussion in #8, IBMAs can be either one- or two-tailed, the FFX GLM can be done with FSL, and SE maps from NIDM results should be squared before being used by FSL.


tsalo:
I think that this PR is ready to merge. Unless anyone has any concerns about it, I plan to merge it later today.

PR #17, opened 2018-05-18
[ENH] Add decoders

tsalo:
The way I'm setting up the discrete decoders assumes that foci and annotations (i.e., features) are stored in separate pandas DataFrames.


nimare/decode/continuous.py
```diff
@@ -15,4 +20,40 @@ def gclda_decode_roi(model, roi, topic_priors=None, prior_weight=1):
...
+            target_image='specificity_z', **kwargs):
+        # Check that input image is compatible with dataset
+        assert np.array_equal(img.affine, self.dataset.mask.affine)
+
+        meta_args = {k.split('meta__')[-1]: v for k, v in kwargs.items() if
+                     k.startswith('meta__')}
+        kernel_args = {k: v for k, v in kwargs.items() if
+                       k.startswith('kernel__')}
+        kwargs = {k: v for k, v in kwargs.items() if not
+                  (k.startswith('kernel__') or k.startswith('meta__'))}
```
tsalo: This whole approach is a little awkward, but I'm not sure if there's a better way to initialize a well specified meta-analysis estimator _inside_ the decoder.

tsalo: After moving the `ids` and `ids2` arguments from the CBMA initialization to `fit`, this problem is resolved.

PR #40, opened 2019-01-14
WIP new json format examples

Shotgunosine:
building out new example jsons based on the format proposed at coco2019.


adelavega:
Do we need the `_id` key? Especially since we decided that names should be fairly static?

nimare/tests/new_json_format/dset_study1.json
```diff
@@ -0,0 +1,75 @@
...
+        {
+          "@id": "cc_id1",
+          "Name": "household items"
+        },
+        {
+          "@id": "cc_id2",
+          "Name": "animals"
+        }
+      ],
+      "@id": 9593960,
```
tyarkoni: I think the @id fields will eventually be populated by the central clearinghouse, and shouldn't be user-assigned. For objects that we think need some user-provided identifier, we should introduce explicit fields for that. E.g., schema.org provides an `identifier` property at the `Thing` level, and the working definition of `Point` includes a `label`. For the examples, I think we can just drop all `@id` properties.

nimare/tests/new_json_format/dset_study1.json
```diff
@@ -0,0 +1,75 @@
...
+        {
+          "Images": {
+            "Elements": [
+              {
+                "Path": "orig/9593960_1_beta.nii.gz",
+                "Value": "Beta"
+              },
+              {
+                "Path": "orig/9593960_1_se.nii.gz",
+                "Value": "StandardError"
```
tyarkoni: We might need to break this up into multiple components—for things like mean, sd, etc., the computation is independent of the unit. We could maybe use `Quantity` and `Unit` or something like that. Alternatively, we could just have the controlled vocabulary include a bigger list of compound strings (e.g., `Beta`, `BetaStandardError`, etc.).

nimare/tests/new_json_format/dset_study1.json
```diff
@@ -0,0 +1,75 @@
...
+              },
+              {
+                "Path": "orig/9593960_1_z.nii.gz",
```
tyarkoni: What would this be relative to? The JSON file itself? We should probably talk about how to handle file packaging, as this has come up in other places.

nimare/tests/new_json_format/dset_study1.json
```diff
@@ -0,0 +1,75 @@
...
+              {
+                "Path": "orig/9593960_1_se.gii.gz",
+                "Value": "StandardError"
+              },
+              {
+                "Path": "orig/9593960_1_z.gii.gz",
+                "Value": "ZStatistic"
+              }
+            ],
+            "Space": "MNI_2mm"
```
tyarkoni: Probably doesn't matter for example's sake, but is MNI 2 mm a sensible surface space? Doesn't this only work for volumes?

nimare/tests/new_json_format/dset_study1.json
```diff
@@ -0,0 +1,75 @@
...
+          "Name": "household items"
+        },
+        {
+          "@id": "cc_id2",
+          "Name": "animals"
+        }
+      ],
+      "@id": 9593960,
+      "Analyses": [
+        {
```
tyarkoni: Each `Analysis` should probably have a mandatory `"Label"` or `"Name"` property, and a strongly encouraged `"Description"` property.

nimare/tests/new_json_format/dset_study1.json
Shotgunosine: I had imagined these examples as prototypes of the completed json including fields that would not be entered by a human (such as @id).

nimare/tests/new_json_format/dset_study1.json
Shotgunosine: Agreed

tyarkoni:
FYI, I have some code for generating working JSON-LD Schema.org definitions up [here](https://github.com/neurostuff/nimads). It's very incomplete, buggy, and doesn't fully comply with the draft spec, but is already serving JSON-LD schemas (e.g., [http://neurostuff.org/nimads/Study.jsonld). Right now I'm still experimenting with this to get a feel for what it will look like when used in various workflows. It would probably make sense to move at least the initial development of examples to that repo, and then we can return to this PR with more worked examples one we've converged on something we're satisfied with.

tsalo:
Tal is going to open a PR at some point for #218, so I'm going to close this.

PR #69, opened 2019-01-15
updating Dockerfile and create_dockerfiles 

djarecka:
updating Dockerfile and create_dockerfiles:
* using the current version of neurodocker
* installing nimare from master
* removing python dependencies that are not needed
* adding Java and MALLET toolbox to neurodocker command

The current Dockerfile was created with the current create_dockerfiles

**edit**
I've changed the docker, so it copies content of the current file instead of installing master, so it can e used for testing (if you just want to use it for development you should just build in the `NiMARE` folder.


create_dockerfile.sh
```diff
@@ -1,13 +1,16 @@
-docker run --rm kaczmarj/neurodocker:v0.3.2 generate -b neurodebian:stretch-non-free -p apt \
-    --install fsl \
+docker run --rm kaczmarj/neurodocker:0.4.3 generate docker -b neurodebian:stretch-non-free -p apt \
+    --install fsl gcc g++ software-properties-common \
     --user=neuro \
     --add-to-entrypoint "source /etc/fsl/fsl.sh" \
-    --miniconda env_name=nimare \
+    --miniconda create_env=nimare \
```
jdkent:
```suggestion
     --copy . /opt/NiMARE/ \
    --miniconda create_env=nimare \
```

If we want to be similar to [fmriprep](https://github.com/poldracklab/fmriprep/blob/dd0f4d4dfc26b107dfd9073760c2d03940051552/Dockerfile#L170) we can build the container around the current code in the repository as opposed to the master branch. This will make it easier to run tests on circleci with the container, since the image will be built around the user's code.

create_dockerfile.sh
```diff
@@ -1,13 +1,16 @@
...
     miniconda_version=4.3.31 \
-    conda_install="python=3.6 jupyter jupyterlab jupyter_contrib_nbextensions
-                   matplotlib scikit-learn seaborn numpy scipy pandas
-                   statsmodels nibabel nipype" \
-    pip_install="nilearn pybids duecredit pymc3" \
+    conda_install="python=3.6 jupyter jupyterlab jupyter_contrib_nbextensions seaborn" \
+    pip_install="https://github.com/neurostuff/NiMARE/tarball/master" \
```
jdkent:
```suggestion
    pip_install="/opt/NiMARE/" \
```
to install the code we copied in from above.

tsalo:
Merging now!

PR #79, opened 2019-01-16
ADD when ALE is called with 1 core, don't use a pool

### LINKED ISSUE #78: ALE uses a multiproc pool even if one core is specified
It helps debugging and profiling if there's an easy way to run without multiproc pool.

Shotgunosine:
Pass n_cores to run_ale and if n_cores == 1, then don't use a multiproc pool

closes #78


chrisfilo:
Could you abstract this away into a helper function and use everywhere where Pool is used?

tsalo:
Merging but will address @chrisfilo's point in a separate PR.

PR #80, opened 2019-01-16
adding the `conperm` CLI command

chrisfilo:
```
Usage: nimare conperm [OPTIONS] CONTRAST_IMAGES...

  Metaanalysis of contrast maps using random effects and two-sided inference
  with empirical (permutation based) null distribution and Family Wise Error
  multiple comparison correction.

Options:
  --output_dir TEXT     Where to put the output maps.
  --output_prefix TEXT  Common prefix for output maps.  [default: ]
  --n_iters INTEGER     Number of iterations for permutation testing.
                        [default: 10000]
  --help                Show this message and exit.
```

### DIFF
--- a/nimare/workflows/ibma_perm.py
+++ b/nimare/workflows/ibma_perm.py
@@ -24,8 +24,10 @@ def con_perm(contrast_images, output_dir=None, output_prefix=output_prefix_defau
              n_iters=n_iters_default):
     target = 'mni152_2mm'
     mask_img = get_template(target, mask='brain')
+    click.echo("Loading contrast maps...")
     z_data = apply_mask(contrast_images, mask_img)
 
+    click.echo("Estimating the null distribution...")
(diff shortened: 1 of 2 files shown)


nimare/workflows/ibma_perm.py
```diff
@@ -0,0 +1,37 @@
...
+
+    res = rfx_glm(z_data, mask_img, null='empirical', n_iters=n_iters)
+
+    if output_dir is None:
+        output_dir = os.getcwd()
+    else:
+        pathlib.Path(output_dir).mkdir(parents=True, exist_ok=True)
+
+    for name, img in res.images.items():
+        img.to_filename(os.path.join(output_dir, output_prefix + name + ".nii.gz"))
```
tsalo: You should be able to use MetaResult.save_results to do this more cleanly.

chrisfilo: done

Issue #36, opened 2019-01-14
Boilerplate


tsalo:
In order to incentivize use and contributions from developers of new methods, we'll want to generate some kind of boilerplate with citations beyond what duecredit would create.

tsalo:
We will implement the boilerplate in the command line tool.

tsalo:
Should we write out a boilerplate text file to the output directory or print the boilerplate to the terminal?

tyarkoni:
I vote for the latter—text sent to stdout can always be piped to a file.

tsalo:
Okay, that should be easy enough with logging (per #58).

PR #98, opened 2019-01-27
[DOC] Draft boilerplates for ALE and SCALE workflows

tsalo:
Progress on #36.

- Add boilerplates for ALE and SCALE CLI workflows.
- Add Colin MNI template (used by GingerALE) and make default for ALE and SCALE.
- Support both `-h` and `--help` for nimare CLI help.


tsalo:
@62442katieb @mriedel56 @kimberlylray You have written up ALE meta-analyses for publications in the past. Would you mind weighing in on the boilerplates? Is there any more information that we should provide?

62442katieb:
Of course! I’ll take a look later today.

kimberlylray:
Hey Taylor, heres a list of additional details that might be useful: 
1. number of foci contributing to ALE
2. total number of subjects across all studies in ALE
3. does NiMare ask for a minimum cluster volume like gingerALE does? 
4. specify the gray matter template - Im assuming its one of the MNI?

Also - is the contrast between two ALEs going to be an option?

tsalo:
@kimberlylray Thank you!

I can add each of those, since we have all of that information.

Contrast analysis _will_ be supported, but currently is not. Technically, the code is already there, but it's untested and I don't know if it's any good. Once we've tested it, I'll add in an optional `sleuth_file2` argument to the ALE CLI. If you have a boilerplate (or even an example from a publication) for contrast analyses that I can work off of, that would be extremely helpful.

Issue #66, opened 2019-01-15
Refactoring discussion thread


tyarkoni:
Following discussion, I thought it would be good to start a thread for discussion of the proposed refactoring of the codebase. The main suggestion so far is to rework the API so it's centered around a distinction between `Transformer` classes whose `transform` method returns a `Dataset` object, and `Estimator` classes whose `fit` method returns a `MetaResult` object. It sounds like almost every existing class fits fairly comfortably into one of these hierarchies. The package structure might then look something like this:

```
nimare.dataset (`Dataset`, `Study`, other data classes)
nimare.transformers
--nimare.transformers.base (base `MetaTransformer` class etc.)
--nimare.transformers.annotate (annotation tools)
--nimare.transformers.kernel (kernel estimation methods)
--nimare.transformers.decomposition (dimensionality-reducing transformations)
nimare.estimators
--nimare.estimators.base (base `MetaResult` class)
--nimare.estimators.cbma (CBMA estimators)
--nimare.estimators.ibma (IBMA estimators)
--nimare.estimators.decode (decoding methods)
--nimare.estimators.parcellate (parcellation methods)
nimare.utils
--nimare.utils.io (conversion utilities to/from NIMADS format and other formats)
--nimare.utils.stats (stats utilities)
--nimare.utils.cli (Command-Line Iinterface)
nimare.workflows (tools for constructing workflows/pipelines)
```

Thoughts?

tyarkoni:
I think this is all more or less done now, so I'm closing this.

PR #107, opened 2019-03-16
[REF] Reorganize package to incorporate Transformers and Estimators

tsalo:
References #66.

- ~~Replace estimator method `fit` with `predict`.~~
- Move stats, io, and utils into new submodule nimare.utils.
- Further consolidate base classes in nimare.base.
- Consolidate Dataset and Database classes.
- Add basic dataset-querying functions (`get_labels`, `get_studies_by_label`, `get_studies_by_mask`, and `get_studies_by_coordinate`).
- Load image files into DataFrame within Dataset the same way labels and coordinates are loaded.


tyarkoni:
What's the rationale for replacing `fit` with `predict`? I think the semantics of what most of these methods are doing is much closer to fitting than to predicting...

tsalo:
I don't remember the exact reasoning, since this is a cleaned up version of [some refactoring](https://github.com/tsalo/NiMARE/commit/359b69975db9f35e427907c27866862e09e7ecac) I started at Coastal Coding and it's been a little while, but I think I switched to `predict` to better match `sklearn`'s Estimators. Estimators' `fit` method updates the Estimator in place, I think, while `predict` returns a result. I can switch it back to `fit` pretty easily, though.

tyarkoni:
I think we should probably stick with `fit`. It's true that it won't perfectly parallel the sklearn API, but that reflects an underlying divergence in usage patterns. In our case we're almost always going to be returning parameter estimates/summaries from a fitted model, and there will almost never be any prediction per se. So, at least in this respect, statsmodels (which also uses `fit` to fit the model and return a result) is probably a better model for us than sklearn.

tsalo:
Sounds good to me. I'll switch back to `fit`.

tsalo:
@tyarkoni Do you mind if I merge this PR now or would you prefer to look over it first? Behavior should be pretty much the same (except for the consolidation of Database and Dataset classes into a single Dataset class), and tests are passing, so I don't think there are any issues with it except perhaps more high-level conceptual concerns.

tyarkoni:
Fine with me! FWIW, I just skimmed through very quickly and nothing jumped out at me as obviously problematic.

PR #115, opened 2019-04-25
[WIP] API changes

tyarkoni:
This is a WIP that makes major changes to the API. Everything here is up for discussion; I just figured that it would be easier to prototype the changes directly and push them for comment. None of this is tested and it is undoubtedly very broken. Once we decide how to proceed, I'll patch everything so it runs.

To summarize:
* I moved a bunch of stuff around, though most of these moves are not meant as permanent suggestions and I think we should open a separate issue (or pick up from #66) to discuss this.
* I made a whole bunch of changes to the `Corrector` hierarchy. The general idea is to push as much logic into the base class as possible. There is now a `_validate_input` method that verifies that the `MetaResult` passed to the `transform()` method contains all the needed inputs (generally just the `p` values). Also, the internal `_fit` call (or the correction method implemented in an `Estimator` class) is now expected to return a dictionary of new (corrected) maps, rather than making direct changes to the `MetaResult`. This allows us to centralize the naming of maps. (To facilitate sensible naming, there is also now a new `name_suffix` property every `Corrector` must implement.)
* Masking and unmasking are operations that IMO should never take place inside the statistical methods (e.g., `stouffers`), but only inside the wrapping classes (e.g., `Stouffers`). This allows us to use these methods generically for meta-analysis of *any* kind of data that can be represented as arrays, not just for fMRI. In line with that reasoning, I would suggest moving all of the non-imaging-specific meta-analysis functions (which is pretty nearly all of the IBMA methods) into their own module. (Relatedly, I think unmasking should be deferred as long as possible—probably not until the user tries to actually retrieve images from the `MetaResult`).
* Correction should also be separated from estimation. I think we had already talked about this at the NiMARE hackathon, and I worked up an implementation to see what this would look like. In addition to to `Corrector` hierarchy changes described above, this means no arguments like `corr` get passed to estimators any more. I demo this for `Fishers`/`fishers` (I haven't touched any of the other estimators/functions, but they should eventually receive the same treatment).
* Instead of passing a `dataset` to an `Estimator` at initialization, it's now passed to `fit()`, which is the more sklearn-like approach. Conversely, extra arguments that modulate estimation behavior must now be passed in at initialization. The *only* argument `fit()` should receive is a `Dataset` instance.
* The base `Estimator` class adds a `_validate_input()` method that runs when `fit()` is called and makes sure the  current estimator has all the inputs it needs. These are defined in the `_inputs` class attribute. To make this work nicely, `Estimator` subclasses now implement a `_fit()` method rather than `fit()`, which allows standardized processing before and after `_fit`.
* I removed a bunch of classes that weren't doing much of anything (e.g., `MetaEstimator`, `Result`). I also think we should drop all the data modeling classes (`Study`, `Contrast`, etc.), as they seem like unnecessary overhead (I'm aware that I was the one who added them).
* To keep things tidy, images are passed to `MetaResult` in a `maps` argument, rather than as `kwargs`.

I'm probably forgetting a bunch of stuff, but that's a start. Take a look through this and see what you think.


tsalo:
Sorry for the conflicts I just introduced (and for not reviewing earlier). Some thoughts:

- Your points about masking and unmasking make a lot of sense.
- I’m still unsure about completely separating correction from estimation, because some correction methods (e.g., FWE correction as implemented with the CBMA estimators) require the dataset, the kernel estimator, etc. They may even need to be integrated into the estimation procedure (not sure; will need to take a closer look at the code).
- I know that we want to make Datasets immutable, but until we have working NIMADS code (pynimads?) I feel like we need to retain the search methods in order for NiMARE to be useable. What about adding a method to Dataset that will return a smaller version based on requested IDs? That way we can drop the IDs argument from most functions, but the package should still be able to do everything it can now.
     - I’d like to get some pseudo-code written for the less common methods we want to support (e.g., CBP, MACM, and decoding) to make sure that they’re compatible with this refactor.

tyarkoni:
> Sorry for the conflicts I just introduced (and for not reviewing earlier). Some thoughts:

No worries, this was meant just as a proof-of-concept—even in the best case, it would require extensive refactoring of the existing codebase.

> I’m still unsure about completely separating correction from estimation, because some correction methods (e.g., FWE correction as implemented with the CBMA estimators) require the dataset, the kernel estimator, etc. They may even need to be integrated into the estimation procedure (not sure; will need to take a closer look at the code).

I think my suggested approach accommodates this scenario. Take a look [here](https://github.com/neurostuff/NiMARE/pull/115/files#diff-a6a0172b5b2e9f6e7a7864c554937403R70). Note that the (optional) `_correction_method` implemented in an `Estimator` class now receives all arguments that the `Corrector` instance was initialized with. In addition to this, it will obviously also have access to anything else in the `Estimator` instance. Since the `Estimator` is welcome to store the last-fitted `Dataset` internally, I don't think there's anything that the correction method needs but won't be able to access. But in the majority of cases, where no custom correction is required, this approach will save a lot of redundant code.

> I know that we want to make Datasets immutable, but until we have working NIMADS code (pynimads?) I feel like we need to retain the search methods in order for NiMARE to be useable.

FWIW, I currently have a working prototype of a study editor that reads and saves NIMADS JSON-LD data (including the full hierarchy of analyses, contrasts, images, points, etc.). It also ingests automatically from NeuroVault (and soon Neurosynth), and facilitates centralized searching via a REST API. It's not quite ready for public display, but by OHBM, I think it will be nicely dockerized and ready for at least alpha testing. I've started working on a Dataset Editor (i.e., an interface for assembling a collection of analyses that would get fed to NiMARE). So I think we're actually not too far from working code.

That said, I have no objections to temporarily retaining any search methods already written. It's more that I think it's not worth investing further effort beyond what's already done into this, as I think we're not too far from being able to release a centralized search API.

> What about adding a method to Dataset that will return a smaller version based on requested IDs? That way we can drop the IDs argument from most functions, but the package should still be able to do everything it can now.

Fine with me, with the above caveat that if it takes you more than an hour or two, it may not be worth the effort.

> I’d like to get some pseudo-code written for the less common methods we want to support (e.g., CBP, MACM, and decoding) to make sure that they’re compatible with this refactor.

Sounds like a good plan. Note that even if they're not fully compatible, if the number of exceptions turns out to be small, I think it may still makes sense to push forward. It wouldn't be the end of the world if, e.g., there are one or two estimators that need to handle their own correction internally, provided we're still centralizing the correction code for everything else.

> One other concern I have about making Datasets immutable is the memory cost. It's taking more than an hour and a half to load the Neurosynth NiMARE dataset on my laptop. I wonder if it makes more sense to operate on more datasets that would typically be smaller (except when doing reverse inference, MACMs, or training annotation models) or to operate repeatedly on one very large dataset by passing around just the identifiers of experiments within the dataset.

The guiding principle behind the proposed architecture is modularity and separation of concerns. Search and filtering, under this proposal, would not be the job of NiMARE, and would in any case never need to be done in-memory—that's what relational DBs are for! The way I see it, the datasets people feed into NiMARE would pretty much always be the size of a typical CBMA or IBMA analysis. People would use a web interface to search the centralized NeuroStuff DB and construct a static dataset file, which they would then export as JSON-LD. AFAIK nobody really does substantive meta-analyses involving thousands of studies, so there should virtually never be a case where someone has to worry about memory (and if they do, I mean, that's enough of a niche case that they can rent a high-memory node for a day).

FWIW, if someone ever wants to write a local package that reproduces and filters the entire NeuroStuff DB, they're welcome to do that... but it would still make sense to do that using a relational DB, and not in memory. Frankly, the only reason the Neurosynth package still does things in-memory rather than using, say, SQLite was that initially the coordinate database was quite small, so it didn't matter, and later on I was too preoccupied with other projects to spend time refactoring using a DB. But since we're starting from scratch, we may as well do it right this time. :)

tsalo:
I managed to accidentally make some merge commits to this branch, but I think I reverted them all. I was able to play with the new API and everything makes a lot of sense. As long as there are internal means of searching and slicing the dataset (as added in #121), I think the other tools we want to include in NiMARE will be fine.

I still have a couple of questions though:

1. Do you mind if kernel transformers operate on coordinates DataFrames instead of Datasets? If we force kernel transformers to operate on Datasets, then in the permutation-based multiple comparisons correction methods, we'll need to initialize dummy datasets with random coordinates. It's much easier (and probably more light-weight) to work on DataFrames.
    - Maybe we could allow transformers to operate on either Datasets or DataFrames, with the preferred method being to use a Dataset?
2. The same Corrector may require different maps from the MetaResult object. For example, I think MKDA uses the OF value maps and ALE uses both the ALE value and z-value maps. We can rename "OF" and "ALE" value maps to "metric" (or something), which will be good enough most of the time, but there will still be cases like ALE where different inputs are required. Is there any way to set the required inputs to the correction method inside the estimator?

I also realized that I was just doing a terrible job at building the label and image DataFrames when initializing the Datasets. Once I cleaned it up loading Neurosynth only ended up taking 2.5-3 minutes. Not a major concern, as you mentioned, but it still makes things easier.

tyarkoni:
Re: (1), can you walk me through a concrete example? I have no particular objection to operating directly on DataFrames, I just want to make sure I understand the constraints. It's not obvious to me why taking Datasets as inputs is more complicated/intensive; assuming that a `Dataset` instance has a DF containing coordinates stored somewhere, isn't it just a matter of internally assigning `df = dataset.store.coords` or something like that, and then after that it's exactly the same procedure?

Re: (2), the pattern I proposed is that the `Corrector` subclass first checks to see if the current `Estimator` implements a correction method of the same name. So one way to handle the scenario you mention within the current framework is to implement the correction method separately for each `Estimator` that needs to use different inputs from the `MetaResult`. But I agree that this isn't ideal in a situation where literally the only difference is which inputs are selected, and thereafter the correction procedure is the same.

One approach we could try is to implement a method in `Estimator` classes called `_select_correction_inputs(corrector)`, where `corrector` is either the `Corrector` instance, or a string giving the name of the corrector. Internally, the method would basically just be a lookup table that maps from `MetaResult` to a dict of (possibly renamed) input images, conditional on the correction procedure passed in. It's not the most elegant solution in the world, but it might be a reasonable first pass.

tsalo:
> Re: (1), can you walk me through a concrete example? I have no particular objection to operating directly on DataFrames, I just want to make sure I understand the constraints. It's not obvious to me why taking Datasets as inputs is more complicated/intensive; assuming that a Dataset instance has a DF containing coordinates stored somewhere, isn't it just a matter of internally assigning df = dataset.store.coords or something like that, and then after that it's exactly the same procedure?

I'm mostly thinking of the permutations run in parallel, like here for ALE:
https://github.com/neurostuff/NiMARE/blob/c83fc97fcecc2f6247241cd854d771d672487946/nimare/meta/cbma/ale.py#L502-L510

It would be totally possible to copy `self.dataset` here and to replace `self.dataset.coordinates` with `iter_df`, but it just seems like there could be a bit of overhead if `self.dataset` also contains texts, labels, metadata, etc.

> Re: (2), the pattern I proposed is that the Corrector subclass first checks to see if the current Estimator implements a correction method of the same name. So one way to handle the scenario you mention within the current framework is to implement the correction method separately for each Estimator that needs to use different inputs from the MetaResult. But I agree that this isn't ideal in a situation where literally the only difference is which inputs are selected, and thereafter the correction procedure is the same.
> 
> One approach we could try is to implement a method in Estimator classes called _select_correction_inputs(corrector), where corrector is either the Corrector instance, or a string giving the name of the corrector. Internally, the method would basically just be a lookup table that maps from MetaResult to a dict of (possibly renamed) input images, conditional on the correction procedure passed in. It's not the most elegant solution in the world, but it might be a reasonable first pass.

I'm planning to take a crack at the Correctors once I finish up #155. I think I'll have a better understanding of how that would work once I'm actively working on that code.

tsalo:
I think we've implemented all of the proposed changes with #152, #153, #154, #155, and #157. Do you know if there are any other changes that need to be made?

PR #124, opened 2019-05-27
[ENH] Add texts field to Dataset

### LINKED ISSUE #123: Text storage in Dataset
I think we can add a `dataset.texts` DataFrame that can contain different kinds of text (e.g., abstracts).

tsalo:
Closes #123.

- Add `_load_texts` and `get_texts` functions to Dataset class.
- Get `download_abstracts` function working.

### DIFF
--- a/nimare/annotate/text.py
+++ b/nimare/annotate/text.py
@@ -2,6 +2,7 @@
 Text extraction tools.
 """
 import re
+import logging
 import os.path as op
 
 import nltk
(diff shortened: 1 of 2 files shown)


tsalo:
@tyarkoni The texts-related changes are pretty straightforward, but I am a little less sure about `download_abstracts`. I don't know how well the function fits with the plans for Datasets. On the one hand, I know that we don't want to edit Datasets, but on the other hand, text is often proprietary and can't be shared openly (e.g., in NeuroStuff), which, if I recall correctly, is why Neurosynth doesn't share the abstracts directly. In order to support annotation and decoding based on text, we'll need some way for users to attach texts to Datasets locally. WDYT?

tyarkoni:
Seems fine to me. We can always remove that later if we deem it unnecessary. I don't see a problem including PubMed metadata (including abstracts) in datasets returned from NeuroStuff. I think I didn't do it in Neurosynth mainly because of size—it would have dramatically increased the size of the database. This is much less of an issue when people are doing the study selection online and retrieving only smaller subsets.

tsalo:
Okay, cool. That's good then. I think we'll need to be able to attach text to Datasets in cases where it's more problematic, like full texts, but we can definitely remove `download_abstracts` at a later date if NeuroStuff can store abstracts. 

I'll merge now.

PR #125, opened 2019-05-27
[DOC] Add ecosystem info to documentation

tsalo:
Closes None.

- Add ecosystem figure from OHBM poster.
- Add information about ecosystem to new page to link in with OHBM poster.


tsalo:
@tyarkoni would you mind taking a look at the ecosystem writeup to make sure it makes sense and is accurate?

tyarkoni:
I don't have any specific objections, but it seems a bit premature, inasmuch as most of the tools/packages in question are in early development or don't exist yet. I'd prefer to build the things first, get them working, and then worry about describing them. But I don't have any problem with this as long as you don't mind it changing (probably substantially) over time.

tsalo:
I can definitely tone down the language to make it more clear that this is a tentative plan and to indicate what stage each service is currently in (e.g., planning, building, operating). If there are any you'd like me to cut (e.g., pyNIMADS), I can do that too. I just want to make sure that people can see how NiMARE will fit into everything, especially since we don't plan to make it a one-stop shop and want to emphasize separation of concerns.

PR #136, opened 2019-06-16
[ENH] Use log-P values for FWE maps in CBMA estimators

### LINKED ISSUE #128: Output cluster-level inference maps with cluster log(P) values
We currently output most cluster-level inference maps with Z or metric (e.g., ALE) values, thresholded at some alpha. We can (and should) output them with cluster-level log(P) values and _no_ threshold. This would also eliminate the need for cluster-level threshold arguments.

Originally proposed by @nicholst.

tsalo:
Closes #128.

- For CBMA estimators, output cluster-extent FWE thresholded maps with cluster-level log-P values instead of voxel-level statistic (e.g., ALE) or z values. Also output voxel-level FWE thresholded maps with corrected log-P values.
- Re-run example notebook with updated methods.


tsalo:
The only problem seems to be the ALE subtraction analysis. One step in the analysis is to limit comparisons to significant voxels from each of the individual groups' cluster-level thresholded ALE maps. The actual values in the cFWE maps don't matter, but the index of significant voxels does. Since the cFWE maps are no longer cFWE-thresholded (in fact, they're just thresholded according to the cluster-defining threshold), the index doesn't correspond to "significant" clusters.

@mriedel56 do you have any thoughts on how best to handle this? I'd prefer not to add a cluster-extent threshold argument back into the code, so should we hard-code p<0.05 into the subtraction analysis code specifically?

mriedel56:
When subtraction analysis is called from inputting two datasets into the ALE method, hard-code p<0.05; however, if the subtraction analysis is being called independently, then the user inputs two (presumably thresholded) maps, that define the clusters. Those user-defined maps can be thresholded at any cluster-level FWE threshold. But should there be some kind of check to ensure that the two maps were defined by the same threshold? Like, we don't want one map to be thresholded at p<0.05 and another to be thresholded at p<0.01, right?

tsalo:
That works for me. Regarding the check, I don't think it's possible to check the maps (at least not if we're just accepting maps and not MetaResult objects or something equally involved), so I think we'll just need to trust users.

Issue #130, opened 2019-06-11
Refactor GCLDA model


tsalo:
The GCLDA model implemented here merges both the Dataset and Model classes implemented in the [GCLDA package](https://github.com/timothyrubin/python_gclda). As a result, the code is a bit bloated, and is probably less than efficient. We need to clean up the model and double check that my previous refactor didn't introduce any bugs. If there's any way for us to speed it up, that would be great too.

tsalo:
Just to start, I think we can remove extra methods in the estimator that print out or save summary files. That information will be accessible through the conditional probability tables and the model parameters. Here is a tentative list of those methods:
- `save_model_params`
- `_save_activation_assignments`
- `_save_topic_word_counts`
- `_save_topic_word_probs`

We should also make sure that the estimator inherits from the `AnnotationModel` base class, so that it has access to the `load` and `save` methods without having to implement them itself.

I think there is also an overabundance of attributes. I'm not sure if the peaks, terms, and documents need to be handled as indices, or if we can operate more directly on the DataFrames.

tsalo:
Unneeded methods were removed in #137. It also dealt with inheritance and reduced the number of attributes (by putting them in new dictionary attributes). I'd still like to see if we can eliminate the dataset index attributes though.

tsalo:
I noticed recently that the model is also prohibitively slow, even when run on high-memory nodes on an HPC. To be honest, I'm not entirely sure if the changes I've made since integrating GCLDA into NiMARE have significantly slowed things down or not. By merging the dataset and model classes, I know that I've increased the memory load of the object, but I don't know if memory is the main problem, or if it just takes too long to perform all of the necessary steps. I tried reviewing the code yesterday to figure out ways to streamline things, but I found myself at a loss. There are a few things I could do to reduce the memory load or simplify the class structure, but I don't know enough about the algorithm itself to do anything that would make a real dent.

tsalo:
I'll need to profile the code to see if this makes any difference, but from looking at the code, `GCLDAModel._get_peak_probs()` seems like it would take some time, and we calculate it twice after the attributes that drive the changes in those peak probabilities (namely `mu` and `sigma`) are updated by `GCLDAModel._update_regions()` in each loop. If `_update_regions()` instead calculated the peak probabilities itself and stored them as a model attribute, that would cut the number of times `_get_peak_probs()` is called in half.

tsalo:
[...]

PR #137, opened 2019-06-17
[REF] Refactor GCLDA model

tsalo:
References #130, but does not close. The model could still do with a good review and is still pretty slow. I would like to see if we can optimize it before closing the issue.

- Move several attributes into new `data` and `topics` dictionary attributes. Only done to attributes that the user won't generally care about.
    - I would like to eliminate `data` in the future, if possible, by relying solely on the input DataFrames `counts_df` and `coordinates_df`.
- Remove several methods that were extraneous.
- Move `load` and `save` methods out of Dataset class definition.
- Reorganize base classes so that GCLDAModel and Dataset can inherit.
    - The problem is that Dataset now inherits from NiMAREBase, but NiMAREBase was defined in the same file as Estimator and Transformer, which both need to use Dataset to check inputs. This causes a whole circularity issue that broke everything. I resolved it by moving Estimator and Transformer into `base.estimators` and left NiMAREBase in `base.base`. I also had to add a `base` import to the top-level __init__ file.

### DIFF
--- a/nimare/__init__.py
+++ b/nimare/__init__.py
@@ -17,5 +17,5 @@ with warnings.catch_warnings(record=True) as w:
     from . import utils
     from .version import __version__
 
-    __all__ = ['dataset', 'meta', 'resources', 'io', 'stats', 'utils',
+    __all__ = ['base', 'dataset', 'meta', 'resources', 'io', 'stats', 'utils',
                '__version__']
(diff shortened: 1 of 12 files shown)


tsalo:
A local test with five iterations on the full Neurosynth dataset seems to indicate that functionality is fine, so I'm going to merge this.

PR #144, opened 2019-06-23
[DOC] Add examples gallery

tsalo:
Closes None.

- Add sphinx-gallery for examples. Time-intensive examples shouldn't be run.


tsalo:
I don't know how well this will work, so I will revert it if it causes too many problems with the documentation or takes too long.

PR #146, opened 2019-06-28
[FIX] Support multiline experiment names in Sleuth converter

### LINKED ISSUE #129: Sleuth converter failing on multiline experiment names
The Sleuth-to-json converter is failing on any experiment labels that are more than one line long. This applies both to very long names that are wrapped and to grouped experiments (one of the options for Sleuth exporting). It should be pretty easy to fix.

tsalo:
Closes #129.

- Makes the Sleuth converter more flexible, to allow for multiline experiment names. These occur when the name is too long and ends up wrapped in the text file, or when grouping is applied to consolidate experiments within the same study.


nicholst:
+1

PR #148, opened 2019-07-18
[FIX, TST] Fix CircleCI

tsalo:
Closes None.

- Split FSL-based and Python-based IBMA tests, so we can ignore the FSL-based ones.
- Update CircleCI config file.


tsalo:
Things finally seem to be working (for the most part). Will see what happens regarding Codecov when I merge.

PR #152, opened 2019-07-21
[REF] Update MetaResult and move general IBMA functions into new module

tsalo:
Closes None. This implements a small part of #115.

- Replace MetaResult with @tyarkoni's version.
    - The only difference is that it takes in “mask” instead of “dataset”, since retaining “dataset” at the point where MetaResults are generated within the estimators would take a bit of refactoring at this point. I can do it later though.
- Remove mask argument from most IBMA functions. We still need it for the functions using FSL, but for the pure Python functions we're operating on, and returning, numpy arrays, not images.
- Move non-FSL-dependent IBMA functions into new ESMA (effect-size meta-analysis) submodule. The IBMA estimators are still in the IBMA submodule, because they require img-like inputs.


tsalo:
I'd also like to use this PR to move the IBMA statistical functions into their own module (per @tyarkoni's suggestion in #115). @tyarkoni do you have any thoughts on what the module should be called? I guess they all fall under effect size meta-analyses?

tsalo:
I'm going to merge this but we can revisit the ESMA in the future.

Issue #74, opened 2019-01-16
stouffers and rfx_glm do not use parallelization


chrisgorgo:

tsalo:
Maybe we can use functors for this and #73. I don't know if it's possible to make a class callable without creating an instance, but that would allow us to have a private function for the permutations that can be called with multiprocessing, the same way we do it for ALE and other CBMA algorithms.

tyarkoni:
Haven't thought about this at length, but this might be better handled with a `@parallelize` decorator or something like that, assuming the parallelization code is fairly consistent, or can be easily parameterized. It may be worth looking at [deco](https://github.com/alex-sherman/deco).

tyarkoni:
Closing this, as we'll get parallelization for free now once we wrap PyMARE.

Issue #73, opened 2019-01-16
rfx_glm does not use parallelization


chrisgorgo:

tsalo:
Closing this because the permutation step for contrast permutation (RFX GLM) is exactly the same as the one for z permutation (Stouffer's).

PR #157, opened 2019-07-25
[REF] Separate multiple comparisons correction from estimators

tsalo:
Closes None. Originally a part of #115.

- Add @tyarkoni's Corrector classes, with some small edits.
- Use Estimator._fit instead of fit in order to check inputs automatically.
- Add FWECorrector.
- Add _fwe_correction method to all CBMA methods.
- Split ALE subtraction analysis into a separate class (ALESubtraction).
- Draft Dataset.get method for grabbing required inputs within meta-analyses. This method will need to be improved dramatically, but the current version is functional.


nimare/meta/ibma.py
```diff
@@ -34,24 +34,16 @@ class Fishers(IBMAEstimator):
...
+    _inputs = {
+        'z_maps': ('maps', {'type': 'z'})
+    }
 
-    def fit(self, dataset):
-        """
-        Run a Fisher's meta-analysis.
+    def __init__(self, two_sided=True):
+        self.two_sided = two_sided
+        self.__init__()
```
tsalo: This causes `RecursionError: maximum recursion depth exceeded`

nimare/base/estimators.py
```diff
@@ -24,24 +23,44 @@ def __init__(self):
     def transform(self, dataset):
         """Add stuff to transformer.
         """
-        if not isinstance(dataset, Dataset):
+        if not hasattr(dataset, 'slice'):
```
tsalo: Replacing `isinstance` with `hasattr` so I can use fake datasets in tests. If there's a more elegant way to deal with it, I'd be happy to do that instead.

nimare/tests/test_cbma_ale.py
```diff
@@ -68,13 +72,29 @@ def test_ale(testdata1):
     """
     ale_meta = ale.ALE(n_iters=5, n_cores=1)
     ale_meta.fit(testdata1)
+    corr = FWECorrector(method='permutation')
+    result_corr = corr.transform(ale_meta.results, voxel_thresh=0.001)
```
tsalo: I think this is what standard usage would look like. WDYT?

nimare/tests/test_cbma_ale.py
```diff
@@ -66,15 +70,34 @@ def test_ale(testdata1):
     """
     Smoke test for ALE
     """
-    ale_meta = ale.ALE(n_iters=5, n_cores=1)
+    ale_meta = ale.ALE()
     ale_meta.fit(testdata1)
+    corr = FWECorrector(method='permutation')
+    result_corr = corr.transform(ale_meta.results, voxel_thresh=0.001,
+                                 n_iters=5, n_cores=1)
```
tsalo: I think this is what standard usage would look like. WDYT?

tsalo:
@tyarkoni I think I've got a better handle on how the correctors might work now that I've got one of them working. I still have some questions though:

1. How can we use multiple comparisons correction with the ESMA functions? They return dictionaries, while the correctors require MetaResults. Also, all of the current methods use `statsmodels` for multiple comparisons correction, so that's not a problem, but if any methods did have more involved correction methods that would otherwise be implemented _within_ the function, what would be the best course of action there? Would we want to convert the functions to classes in that case?
2. Arguments for the class method `_fwe_correction` are fed into `FWECorrector.transform`, not `FWECorrector.__init__`. I couldn't figure out how to get the latter to work. Is that a problem?
3. The `Estimator._fwe_correction` method is going to be called regardless of which `method` (e.g., bonferroni vs. permutation) is fed into `FWECorrector.__init__`. Does that mean we need to add the `method` to `Estimator._fwe_correction`, or can we make calling the class method `method`-dependent (i.e., use standard `FWECorrector._transform` for bonferroni and call `Estimator._fwe_correction` for permutation)?

nimare/base/estimators.py
tyarkoni: Mocking might be the way to go if we anticipate this kind of thing being an issue routinely, but is probably more trouble than it's worth at this point.

nimare/base/estimators.py
```diff
@@ -24,26 +21,41 @@ def __init__(self):
...
             raise ValueError('Argument "dataset" must be a valid Dataset '
                              'object, not a {0}'.format(type(dataset)))
+        for k, v in self._inputs.items():
+            data = dataset.get(v[0], **v[1])
+            if not data:
+                raise ValueError("Estimator {0} requires input dataset to "
+                                 "contain {1} with filters {2}, but none "
+                                 "were found.".format(self.__class__.__name__,
+                                                      v[0], v[1]))
+            setattr(self, k, data)
```
tyarkoni: I'm now thinking it might not the best practice to set attributes directly on the instance... we should probably rename `_inputs` to something more explicit, like `_required_inputs`, and then store the actual data in an `inputs` dictionary rather than setting them as attributes directly.

nimare/base/estimators.py
```diff
@@ -2,12 +2,9 @@
 Base classes for datasets.
 """
 import logging
-from abc import ABCMeta, abstractmethod
```
tyarkoni: Isn't `ABCMeta` necessary? Haven't double-checked this, but I think `abstractmethod` will only work on classes that have `ABCMeta` as their metaclass.

nimare/correct/correct.py
```diff
@@ -0,0 +1,138 @@
...
+        # MetaEstimator, use it. Otherwise fall back on _transform.
+        if (method is not None and hasattr(est, method)):
+            corr_maps = getattr(est, method)(result, **kwargs)
+        else:
+            self._validate_input(result)
+            corr_maps = self._transform(result)
+
+        # Update corrected map names and add them to maps dict
+        corr_maps = {(k + self._name_suffix): v for k, v in corr_maps.items()}
+        result.maps.update(corr_maps)
```
tyarkoni: See earlier note. We're updating the `.maps` dict, so it definitely needs to be explicitly copied inside `MetaResults.copy()`, otherwise both copies of the `MetaResults` instance will have the same (modified) `maps` dict.

nimare/correct/correct.py
```diff
@@ -0,0 +1,138 @@
...
+    _correction_method = '_fwe_correct'
+
+    def __init__(self, method='bonferroni'):
+        self.method = method
+
+    @property
+    def _name_suffix(self):
+        return '_corr-FWE_method-{}'.format(self.method)
+
+    def _transform(self, result, voxel_thresh=0.001):
```
tyarkoni: I think all arguments should be passed to `__init__` rather than `transform` unless there's some good reason not to. The `voxel_thresh` is IMO something that should be set at the `Corrector` level, not something that varies across `MetaResult` objects. (I assume this is just an oversight, since `voxel_thresh` isn't actually used anywhere at the moment.)

nimare/tests/test_cbma_ale.py
tyarkoni: I like it, with two possible modifications:

1. I think maybe `.fit` should return the `MetaResult` object by default (though it should also set it internally).
2. The arguments to `transform` should probably be passed to the `FWECorrector` at initialization.

There's also an alternative pattern we could consider, which is to make the correction classes functors (i.e., classes that are called like functions). That would allow us to do something like:

```python
ale_meta = ale.ALE()
result = ale_meta.fit(cbma_testdata1)
result_corr = fwe_correct(result, method='permutation', voxel_thresh=0.001, n_iters=5, n_cores=1)
```

Internally, this would actually be a pretty small change from the present implementation. `fwe_correct` would still be defined as a class, and we would have the same class hierarchy. The main difference is we would direct `__call__` to `transform` (or just replace the latter with the former).

It would make for a slightly cleaner API; the main drawback is the loss of consistency with the APIs of the other major class hierarchies. I don't have a strong preference either way (might be worth polling a few other people).

nimare/base/estimators.py
tsalo: I think that both Estimator and Transformer inherit from NiMAREBase, which has ABCMeta as its metaclass.

nimare/correct/correct.py
tsalo: Okay, that makes sense. I was having trouble accessing the arguments from init, but I can just put them in an attribute to access within transform, right? Like this:

```python

class FWECorrector(Corrector):
    """
    Perform family-wise error rate correction on a meta-analysis.

    Parameters
    ----------
    method : `obj`:str
        The FWE correction to use. Either 'bonferroni' or 'permutation'.
    **kwargs
        Keyword arguments to be used by the FWE correction implementation.
    """

    _correction_method = '_fwe_correct'

    def __init__(self, method='bonferroni', **kwargs):
        self.method = method
        self.parameters = kwargs
```

nimare/base/base.py
```diff
@@ -231,3 +232,6 @@ def save_maps(self, output_dir='.', prefix='', prefix_sep='_',
             filename = prefix + prefix_sep + imgtype + '.nii.gz'
             outpath = os.path.join(output_dir, filename)
             img.to_filename(outpath)
+
+    def copy(self):
```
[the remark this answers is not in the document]
tsalo: Would something like the following work?

```python
def copy(self):
    new = MetaResult(copy.copy(self.estimator),
                     copy.copy(self.mask),
                     copy.deepcopy(self.maps))
    return new
```

nimare/tests/test_cbma_ale.py
tsalo: > 1. I think maybe .fit should return the MetaResult object by default (though it should also set it internally).
That's easy enough. I'll do that.

> 2. The arguments to transform should probably be passed to the FWECorrector at initialization.
I think I've got some questions in another comment, but definitely working on this.

I'm fine with holding off on converting the Correctors to functors until a later date, depending on how people feel about using them as they currently are. Once this is merged and the Correctors are all working, I can raise an issue to get folks' opinions.

tsalo: I spent some time today looking into functors to maybe solve #73 and #74, but it looks like they still need to be initialized (i.e., `func_inst = functor(); func_inst('call')` instead of `functor('call')`). I could be missing something though.

nimare/base/estimators.py
tsalo: I changed it a bit and added a really bad first attempt at a `Dataset.get` method, but it seems to be working for images and metadata at least.

tsalo:
I think that everything has been dealt with. It would be nice to see what you think of `Dataset.get` and the handling of required inputs for Estimators, but I'd also be happy to merge now and to circle back to those things in a later PR.

I also think that this finishes up the API changes proposed in #115. Does that seem right or am I missing something?

nimare/base/base.py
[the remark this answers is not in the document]
tyarkoni: I think we probably want to pass the original `estimator` and `mask` by reference... it's just the `maps` that I think need to be copied, as they'll definitely be modified.

tyarkoni:
LGTM! Let's go ahead and merge and any other issues can be addressed in other PRs.

tyarkoni:
Oh, and have not looked at `Dataset.get` and handling of required inputs yet, but that shouldn't hold up merging.

PR #163, opened 2019-08-02
WIP: Add FFX/RFX estimators

tyarkoni:
This PR adds new FFX and RFX estimators and restructures some of the estimation code. Main changes:

* A new `ffx` estimation function that runs a standard fixed-effects meta-analysis.
* A `method` argument to the `FFX` class (renamed from `FFX_GLM`) that allows users to specify `'fsl'` optionally (but defaults to the standard approach).
* Removed the `ffx_glm` and `mfx_glm` functions as they were just wrappers for `fsl_glm` and we now have (or soon will have) alternative methods for `FFX` and `RFX`.

TO DO:
* [X] Add mixed-effect estimation to `stan_mfx`
* [ ] Add a `StanMFX` class that wraps `stan_mfx` and follows the existing API
* [ ] Split `stan_mfx` into separate compilation and estimation steps for efficiency
* [ ] Get PyStan to install correctly on CircleCI
* [ ] Improve returned results (to include credible intervals, pseudo-p values, etc.)
* [ ] Auto-scale priors based on data to avoid initialization failures and improve sampling efficiency


tyarkoni:
@tsalo: I'm now assuming `var_maps` instead of `se_maps`, and haven't touched the code in `fsl_glm`. Presumably that needs to be updated to make sure that we compute `se_maps` 
from the `var_maps` and `sample_sizes`.

tsalo:
Per a conversation with Camille Maumet, [here](https://github.com/neurostuff/NiMARE/issues/8#issuecomment-386651975), we don't scale the se or var maps by sample size. We just square the se maps to make the flameo-compatible var maps. All we should have to do is remove the line squaring the maps.

tyarkoni:
Are we talking about standard error maps, or standard deviation maps? Standard errors are estimates of the SD of the population distribution underlying the observed sample. They depend on the sample size (because the estimate gets more precise the larger the sample). But if FSL expects variance maps, then yes, we just need to drop the sqrt.

nicholst:
+1 on using "SE2" or "StdErr2" or somehow referencing `standard error'
terminology, as that is the only way to be clear about what is intended.
(AFAIK, all of the "var"s that @tsalo mentions are squared standard errors
-- aka varcopes -- and not population variances).

tyarkoni:
Still a WIP, but the `stan_mfx` function should be working, and is fairly flexible. I decided to go with an MCMC implementation to start with, as it's way easier to implement, more flexible, and is likely to be less biased. The obvious cost is computation time, but it's not as bad as I expected—a few hundred models (e.g., for an atlas-based reduction) shouldn't take more than an hour or two on most systems, and even whole-brain analysis should be feasible on a cluster (though we'll probably need to write code that distributes different voxels to different workers to really make that work well).

tyarkoni:
I've generalized `stan_mfx` so that it can now do mega-analysis as well as meta-analysis. Basically you can pass various combination of `estimates`, `variances`, `standard_errors`, and `groups`, and will generally build you the best possible model given what you passed in. Once we merge this and #173, we'll be in a position to do ROI-level Bayesian mixed-effects meta/mega-analysis estimation fairly efficiently... pretty exciting!

tsalo:
Now that PyMARE is up and running, and the PyMARE integration is mostly handled (minus Stan and MCC), should we close this?

PR #166, opened 2019-08-06
[ENH] Make convert_sleuth_to_dataset more flexible

62442katieb:

- `convert_sleuth_to_dataset` now allows for multi-line experiment names (which occur quite frequently in Sleuth output textfiles)

### DIFF
--- a/nimare/io.py
+++ b/nimare/io.py
@@ -141,9 +141,9 @@ def convert_sleuth_to_dict(text_file):
     data = data[1:]
     exp_idx = []
     for i in np.arange(0, len(data)):
-        if data[i].startswith('//') and data[i+1].startswith('// Subjects') and data[i-1].startswith('//') == False:
+        if data[i].startswith('//') and data[i + 1].startswith('// Subjects') and data[i - 1].startswith('//') is False:
             exp_idx.append(i)
-        elif data[i].startswith('//') and i+2 < len(data) and data[i+2].startswith('// Subjects'):
+        elif data[i].startswith('//') and i + 2 < len(data) and data[i + 2].startswith('// Subjects'):
             exp_idx.append(i)
         else:
             pass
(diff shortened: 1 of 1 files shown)


tsalo:
It looks like your text editor replaced some of the four-space tabs with two spaces, which is causing problems with flake8. Would you mind fixing the spacing? It'll be a bit easier for me to review that way too.

62442katieb:
I think I've fixed all the whitespace issues. Sorry about that

tsalo:
There are still a couple of style issues that need to be fixed.

nimare/io.py
```diff
@@ -137,12 +139,18 @@ def convert_sleuth_to_dict(text_file):
...
     data = data[1:]
-    metadata_idx = [i for i, line in enumerate(data) if line.startswith('//')]
-    exp_idx = np.split(metadata_idx,
-                       np.where(np.diff(metadata_idx) != 1)[0] + 1)
-    start_idx = [tup[0] for tup in exp_idx]
+    exp_idx = []
+    for i in np.arange(0, len(data)):
+        if data[i].startswith('//') and data[i + 1].startswith('// Subjects') and data[i - 1].startswith('//') is False:
+            exp_idx.append(i)
+        elif data[i].startswith('//') and i + 2 < len(data) and data[i + 2].startswith('// Subjects'):
```
tsalo: This allows 1-2 line contrast names, but not 3+, right? Shouldn't be a problem the vast majority of the time, but I want to make sure.

62442katieb: I believe that's correct! I can add another `if` to make allow for 3+ line contrast names, if you want?

tsalo:
Tests are passing. The pending test is there because it's required but was added _after_ this PR was opened. You could update from neurostuff master, but I don't think it's worth it. Just going to force the merge through now. Thank you for updating!

PR #169, opened 2019-08-10
[REF] Switch CLI from click to argparse

tsalo:
Closes None. I find that `sphinx-argparse` produces more readable documentation than `sphinx-click`. Also, with click we could not call the workflows as Python functions, but with argparse we can. This way, we can add tests for the workflows and CLI separately, as well as allow users to run the workflows from Python.

- Replace all uses of `click` for CLI with `argparse`.
- Add workflows to API documentation, now that they can be called from within Python.


tsalo:
Nothing in this PR should impact core functionality and any bugs introduce to the command-line interfaces can be fixed as we add workflow-specific tests, I think, so I'm going to merge this.

PR #173, opened 2019-08-15
WIP: changes to IBMAEstimator hierarchy and masking functionality

### LINKED ISSUE #175: Generalize masking to support both voxel-based and ROI-based analysis
Currently we maintain a nibabel volume as a `mask` in `Dataset`, `MetaResult`, etc., and then use nilearn's `apply_mask` and `unmask` utilities to convert to and from arrays. This works fine, but is limiting, as a common use case—particularly once we merge the Stan MFX estimator—is likely to be ROI-based analysis.

The proposal here is to replace most, if not all, of the masking functionality throughout the package with a custom `Masker` class (possibly named something else to avoid conflicts with nilearn) that can seamlessly handle conversion to/from both voxels and ROIs. API changes would be minimal; mainly, we would replace `mask_file` with `mask` everywhere, and users would be able to pass (a) a string, (b) a NiBabel image, or (c) a `Masker` instance. If given the first two, we would internally construct a `Masker` that does voxel-wise extraction. But if given a `Masker`, we would then be able to seamlessly handle ROIs.

Usage example:

```python
# Current usage with MNI152 default still works fine
dataset = Dataset("source.json")

# Specifying custom mask only changes argument name
dataset = Dataset("source.json", mask='mymask.nii.gz')

# Use ROI-level extraction with custom mask
m = Masker("mask.nii.gz", regions=True)
dataset = Dataset("source.json", mask=m)
```

Thoughts? I feel pretty strongly that this is going to make life substantially easier when working with computationally intensive models, but let me know if you have any objections, @tsalo .

tyarkoni:
This PR includes several improvements to the `IBMAEstimator` hierarchy and related classes, as well as the way masking is handled. Key changes:

* Base classes for `IBMAEstimator`, `CBMAEstimator`, and `KernelTransformer` are now located closer to the child classes. The principle I've followed is that each base class should be as low in the hierarchy as possible (e.g., there's no reason to put `IBMAEstimator` at `base/` level if only the stuff in `ibma.py` uses it).
* A new `_preprocess_input` method is now automatically called in `Estimator.fit()`. This is implemented in the `IBMAEstimator` hierarchy to do automatic masking.
* `apply_mask` has been removed from all `IBMAEstimator` subclasses, and is now handled uniformly in `_preprocess_input`.
* Adds an `atlases` folder to resources that contains the Harvard-Oxford probabilistic atlas plus its license (necessary for testing region extraction).
* Per discussion in #175, the `mask` arguments to `Dataset` and `IBMAEstimator` now take a string, a nibabel image, or a nilearn `Masker`. Internally, all masking is done via a nilearn Masker. This allows estimators a lot of flexibility in what they're applied to; e.g., one can easily operate over ROIs instead of voxels by passing a `NiftiLabelsMasker` instead of a nibabel volume or `NiftiMasker`. The reliance on the `.transform()` and `.inverse_transform()` methods also makes it easier to move back and forth between image and array space. (Note that at present I haven't updated the entire database to actually *use* this pattern; there are still a lot of places where `apply_mask` is invoked. I have, however, made sure that these all work properly, so all tests currently pass.)
* From a user-facing standpoint, the only breaking change to the API is that the mask argument is now `mask` and not `mask_file`, which I thought was necessary to reflect the increased flexibility. We could potentially also call it `masker` given that the internal attribute is `.masker`.

Closes #175.


tyarkoni:
@tsalo this is ready for review. I ended up combining the `IBMAEstimator` changes with the `Masker` changes, as it would have been a huge headache to deal with the former without the latter.

tsalo:
I'm happy to spread the base classes back out into their associated submodules, although we'll have to update the API documentation to match.

Long-term, do you foresee any other uses for `_preprocess_input`? I imagine it can serve many purposes, but nothing comes to mind.

nimare/meta/ibma.py
```diff
@@ -14,13 +14,33 @@
...
 from ..stats import p_to_z
 
 LGR = logging.getLogger(__name__)
 
 
+class IBMAEstimator(Estimator):
+    """Base class for image-based meta-analysis methods.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
tsalo: I'm a little confused about this. What are `*args` and `**kwargs` going to be here and why do we need to run the Estimator's init?

tyarkoni: Oh, that's vestigial. I briefly considered moving the masking stuff to the base class, but ultimately decided to leave it in `IBMAEstimator`, since it's not really relevant for CBMA.

tyarkoni: On a related note, we should make a decision about whether to enforce explicit inclusion in init signatures of all arguments that get passed onto superclasses/mixins. The tradeoff is usability vs. maintainability. Having to include `mask` in every `IBMAEstimator` would be a huge pain to maintain (have to update multiple docstrings each time any change is made, etc.), but would make it much clearer to users what functionality is available.

Personally I incline to keep it as-is for the time being, as it's easier to make args explicit later than to drop them in favor of `kwargs`. But am open to the alternative.

tyarkoni:
@tsalo re: `_preprocess_inputs`, I can imagine things like smoothing or transformation of input data being potentially useful. E.g., if your estimator needs `z` maps, and the input has `p`, then you can get what you need via conversion. That could go in `_validate_input`, but it could also be construed as a form of preprocessing.

Regardless, I see this as a form of future-proofing: at some point soon we're probably going to think of other things we want to do prior to estimation, and this seems like a natural place for it. But I'm open to  alternatives if you want to drop this, rename it, etc.

tsalo:
@tyarkoni The changes look good to me. I think it'll be easier to figure out the best place for the data transformation once we try implementing it, but `_preprocess_input` looks it will be really useful regardless.

PR #174, opened 2019-08-15
Save pain test dataset to local directory

### LINKED ISSUE #172: Avoid downloading test data on every run
The use of `tmpdir_factory` in the `download_data` test fixture means the pain dataset is downloaded anew every time tests are run. This is inefficient and can stall tests when no internet connection is available (as I just found out the hard way). Probably best to download the test data to the `tests/data` directory the first time the tests are run, and use them from the local store thereafter.

tyarkoni:
When the tests are first run, the NIDM pain dataset will be saved locally under `~.nidm`, preventing a ~200 MB download each time.

Closes #172.

### DIFF
--- a/nimare/tests/conftest.py
+++ b/nimare/tests/conftest.py
@@ -15,20 +15,16 @@ from nimare.tests.utils import download_nidm_pain, get_test_data_path
 
 
 @pytest.fixture(scope='session', autouse=True)
-def download_data(tmpdir_factory):
+def download_data():
     """
-    Download 21 pain studies from NeuroVault to test IBMA functions.
+    Download and save 21 pain studies from NeuroVault to test IBMA functions.
     """
-    tst_dir = tmpdir_factory.mktemp('tests')
-    out_dir = tst_dir.ensure('resources',
-                             'data',
-                             'neurovault-data',
-                             'collection-1425',
-                             dir=True)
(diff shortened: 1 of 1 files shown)


nimare/tests/conftest.py
```diff
@@ -15,20 +15,16 @@
...
+    Download and save 21 pain studies from NeuroVault to test IBMA functions.
     """
-    tst_dir = tmpdir_factory.mktemp('tests')
-    out_dir = tst_dir.ensure('resources',
-                             'data',
-                             'neurovault-data',
-                             'collection-1425',
-                             dir=True)
+    home = op.expanduser("~")
+    out_dir = op.join(home, '.nimare', 'data', 'neurovault', 'collection-1425')
```
tsalo: Why in home instead of `nimare/tests/resources`?

tsalo:
This looks good to me. We could probably also cache the dataset folder in the CircleCI configuration so it doesn't download on there every time, but I can do that in a separate PR.

nimare/tests/conftest.py
tyarkoni: At 200 MB, it's possible users will want to delete it to save space at some point, and I doubt anyone will think to look in the code directory. Putting large datasets in `~/.{package}` is pretty conventional.

PR #177, opened 2019-08-18
[REF] Distribute base classes into their associated modules

tsalo:
References #173, which started moving base classes away from the base module to their associated modules.

- Move the base classes out of `nimare.base` and into the associated modules.
- Update interlinks within API documentation.
- Improve some docstrings.


tsalo:
> 1. I think the natural place for MetaResult is probably in a separate results.py module at top level (or maybe even under meta/, except for the fact that stuff in correct also needs to import MetaResult).

I'm happy to move MetaResult up a level and into its own module. Now that there are only a few base classes not tied to specific modules, I think it's probably a good idea to move the files up a level (i.e., remove the `base/` folder).

> 2. I'm not sure how much it buys us to inherit from NiMAREBase at the moment. I'm suspicious that the save and load methods may not work properly right now (I don't think they're tested), as I think there are probably unpickleable attributes in many of the estimators. The get_params functionality seems potentially worthwhile, but AFAICT isn't actually called anywhere else in the codebase at the moment. And Dataset probably shouldn't inherit from NiMAREBase in any case, because it only a single class (i.e., no further inheritance) and has only three initialization arguments. I would incline to drop it entirely and then bring it back later if there's a need for it.

I think I'd like to keep NiMAREBase for now. We definitely need `load` and `save`, at least for Dataset, and I'm reticent to add more code to the Dataset class by moving NiMAREBase's contents into Dataset (it's already like 600 lines). I haven't been checking whether most of our classes are pickleable, but I should probably look into that. We definitely need some way to save a number of slow-to-train models (e.g., GCLDA).

I assume that `get_params` is more useful for users than for internal operations, but I guess I haven't thought about it much.

Issue #170, opened 2019-08-12
Support transformations on Estimators' required inputs


tsalo:
As part of `Estimator._validate_input`, we should allow transformations and combinations of data. Some transformations to include:
- Standard error maps to variance maps
- T-statistics and sample sizes to z-values
- Combining compatible values with existing ones (e.g., default to z-values when available, and only convert t-statistics when they're not).

tsalo:
It seems like the best place to do this is `Dataset.get`, since that's where data are identified and compiled. The question, I think, is whether to explicitly describe the acceptable data and transformations in `_required_inputs` or to automatically identify them elsewhere (e.g., in `Dataset.get` and/or a new `transformations` module).

tyarkoni:
I'm pretty strongly against any changes to data being made at `Dataset` level that weren't explicitly and specifically invoked by the user. If a user wants to explicitly convert from t to z via a function call, fine. But anything that involves automatic detection/remediation of missing data types should IMO live only in the `Estimator` (and `MetaResult` if needed). I don't think it's the `Dataset`'s responsibility to know what transformations are possible or to predict what inputs an `Estimator` might need.

I also think we should restrict transformations to only simple, deterministic operations where there's no ambiguity. The examples in the list above seem fine, but for anything more complex than that, I think it becomes the user's job to make sure they have the necessary data.

tsalo:
I think that's reasonable, although I'll have to change how we're doing things now to make it work. Currently, `Dataset.get` grabs the relevant data and reduces based on data availability. When calling it, we end up with a dictionary with keys for each of the requested data types and for IDs. Each value is a list of the data/IDs.

To implement transformations in the Estimator, I think I have to disable the reduction implemented in `Dataset.get`, so that the returned dictionary has `None` when a given contrast doesn't have the requested data. It will also need to grab all transformable data for each requested data type, in order to fill in the gaps, when possible, within the Estimator. I think that means we'll also need to describe the transformations in the `_required_inputs` attribute, which will mean making that dictionary more... expressive, I guess. I'm not really sure how to do that, but maybe you could weigh in on the construction of the `_required_inputs` dictionaries once we get to that step.

tsalo:
What do you think of this:

```python
_required_inputs = {
    'z_maps': (('image', 'z'),
               {'transform': 't_to_z',
                'inputs': {'t_maps': ('image', 't'),
                           'sample_sizes': ('metadata', 'sample_size')}
                }
    }
```

[...]

PR #179, opened 2019-08-19
[REF] Implement _required_inputs in CBMA estimators

### LINKED ISSUE #171: Add coordinate support to Dataset.get
Currently, Dataset.get only works for metadata and images, but we'll need to support coordinates as well, in order to use Estimator._validate_input in coordinate-based algorithms.

tsalo:
Closes #171 and references #170.

- ~~Consolidate some of the loading methods in Dataset to reduce duplicate code.~~ Done separately.
- Add "coordinates" to `Dataset.get`, which returns a list of DataFrames divided by contrast ID.
- Drop "n" column from `Dataset.coordinates` in order to use the metadata field "sample_sizes" instead. Automatically average sample sizes for meta-analyses.

TODO:
- [ ] Allow arguments that override existent data within `_required_inputs`.
- [ ] Fix correctors (esp. FWECorrector) so they use `Estimator._inputs` (or some updated version?) the same was the Estimator's main fit method does.
- [ ] Figure out how the kernel estimators should take in sample sizes. Currently, I'm reorganizing `Estimator._inputs` into a single DataFrame (with a sample size column) in `Estimator._fit` and feeding that DataFrame into the kernel estimator, which just doesn't seem to fit in with the approach we're going for.
- [ ] Add `levels` argument to `Dataset._load_data`, in order to load data from the study level when appropriate (e.g., for metadata and texts, but not for images).


tsalo:
Now that #232 is merged and #233 is opened, I can close this.

PR #182, opened 2019-10-30
[FIX] Fix missing parameter q of FDRCorrector

### LINKED ISSUE #181: Missing q parameter in FDRCorrector class
Hi, I think I found a small error in the `nimare.correct.FDRCorrector` class. The transform method of the class is looking at a parameter `q` https://github.com/neurostuff/NiMARE/blob/e17629fb0793a593a9b244ca4fe9dbbc3412f592/nimare/correct/correct.py#L66 
which does not exist in that class.
https://github.com/neurostuff/NiMARE/blob/e17629fb0793a593a9b244ca4fe9dbbc3412f592/nimare/correct/correct.py#L56-L58

The PR will follow.

alexprz:
Closes #181.

I added the missing parameter `q` in the `__init__`function of the `nimare.correct.FDRCorrector`.

The other alternative would have been to read this parameter from the existing `self.parameters` dictionary but I found it weirder since `q` is a required argument.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -57,6 +57,11 @@
       "name": "Laird, Angela R.",
       "affiliation": "Florida International University",
       "orcid": "0000-0003-3379-8744"
+    },
+    {
+      "name": "Pérez, Alexandre",
+      "affiliation": "Montreal Neurological Institute, McGill University",
+      "orcid": "0000-0003-0556-0763"
     }
   ],
   "keywords": [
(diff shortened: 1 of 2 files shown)


tsalo:
Thank you for catching this! Would you mind changing `q` to `alpha` throughout the class, as that's what we use in other FDR implementations in NiMARE?

alexprz:
Done! I also changed the doc consequently.

nicholst:
+1 for alpha over q  :)

tsalo:
@alexprz Everything looks good to me, but before I merge, do you want to add yourself to the Zenodo file?

alexprz:
It was only a small change but I would appreciate it thanks!

tsalo:
Great! Merging now.

PR #192, opened 2019-12-03
[ENH] Improve Corrector transparency

tsalo:
Closes None.

- Make Estimator correction methods public.
- Expand Estimator docstrings with “Examples” and “See Also” sections.
- Explicitly warn users to only call Estimator correction methods
within Correctors.
- Add “_native_methods” attribute to Correctors to make it easier to
validate settings.
- Add “Notes” section to Corrector docstrings clarifying where to look
for Estimator-native correction methods.


tsalo:
@tyarkoni I realized that it's difficult for users to determine what correction methods are available and where to look for more information on Estimator-specific methods, so I'm proposing these changes to remedy that. Users should have a much easier time with these changes, although the big thing that you might not like is that it means making Estimator-specific correction methods public.

Rather than implementing it all now, I've just altered the ALE estimator and the FWECorrector as examples. Would you mind having a look to see what you think when you have some time?

tyarkoni:
At first glance, that seems fine to me. The main change is just exposing the default correction method publicly, right? I have no objections to that, provided that `correct()` remains an alias that's always guaranteed to use a sensible default.

tsalo:
Yeah, I'd say that's the main change here, although I don't think the Estimator-implemented correction methods are necessarily the defaults in all cases. For MKDAChi2, for example, the standard methods don't work because you end up with both consistency- and specificity-p maps instead of just p maps, so we end up implementing all applicable correction methods in the Estimator.

There isn't a general `correct()` method though. The docstrings just point from each Estimator's `correct_[correction-type]_[method]()` method to the associated Corrector, so users know to use that instead of the Estimator's method directly.

tsalo:
Everything's passing and the docs look good, so I'm merging now.

PR #199, opened 2019-12-14
[ENH] Add extract submodule (again)

### LINKED ISSUE #198: Use nilearn-style data directory for downloaded files
Currently, we default to `nimare/resources/` for downloaded files, but that location (1) isn't very accessible if you install with pypi and (2) is located within the git repository if you're working on a fork (which means these files can be accidentally tracked and pushed).

I think it would be preferable to store those files (including MALLET and any trained models) in a more accessible folder, much like how nilearn and nltk do it. We can add relevant data_dir functions to `nimare.utils`, possibly just copied from nilearn.

tsalo:
Closes #198.

- Add an extract submodule. Unlike before, where we planned to add tools for flexible querying of third-party databases like NeuroVault, Neurosynth, and Brainspell, this version is more limited in scope and more like nilearn's datasets module (i.e., functions to download specific resources).
- Add dataset directory-related functions, taken directly from Nilearn and only adapted minutely.
- Refactor GCLDA and LDA a little bit.


tsalo:
I'm not sure why CodeCov isn't ignoring the files and folders I specified in the config file, but other than that it's good to go. Merging now.

PR #222, opened 2020-05-21
[DOC] Fix annotate API docs

tsalo:
Closes None.

- Clean up conf.py, removing unused extensions and variables.
- Add annotate sub-submodules to API file. They're nested in the annotate submodules, and the module.rst template can't document submodules for some reason.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -93,9 +93,14 @@ API
    :toctree: generated/
    :template: module.rst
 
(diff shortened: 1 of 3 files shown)


tsalo:
Going to merge now. Local doc builds tend to look different from RTD, so I'll revisit whether I prefer `napoleon` or `numpydoc` once I see the real build.

PR #224, opened 2020-05-23
[REF, DOC] Update meta-analysis output map names

### LINKED ISSUE #217: Make available MCC methods transparent
At the moment, it's very difficult to figure out what multiple comparison correction methods are available for a given Corrector and Estimator. We need to fix that.

tsalo:
Closes #217. The main change in this PR is to make the output map names for meta-analyses more consistent. There are also some documentation changes, listed below.

The output map name convention is as follows:

```
[value]_desc-[description]_level-[cluster|voxel]_corr-[FWE|FDR]_method-[method]
```
- value: The data type. This will ultimately have a restricted vocabulary, but I've only taken a first pass at that so far. Values in NiMARE include p, logp, z, t, stat (for "FFX stat" from some IBMAs), of (for MKDA), and cope (for FSL-based IBMAs).
- desc: A freeform description, when necessary. For example, to dissociate consistency and specificity maps in MKDA Chi2.
- level: Either voxel or cluster. For thresholded maps only. (We already had this.)
- corr: Either FDR or FWE. Multiple comparisons correction type. For corrected maps only. (We already had this.)
- method: Multiple comparisons correction method (e.g., `montecarlo`, `bonferroni`). For corrected maps only. (We already had this.)

- More explicit Napoleon keyword use.
- Examples!
- Improve docstring for MetaResult.


tyarkoni:
@tsalo I think that of the names you use, only `p` and `z` are relevant to pymare, and those are also `p` and `z` there. There are some other things we'll probably want to pass along to NiMARE (e.g., heterogeneity stats like `I^2` or `H`), but we can cross that bridge when we get to it.

tsalo:
I'll move forward with this as-is then. Thanks!

PR #234, opened 2020-05-27
[FIX] Mask coordinates before kernel convolution

### LINKED ISSUE #37: ALEKernel masks data after convolution
This means that signal for coordinates outside of the mask will be included. Was this intended?

tsalo:
Closes #37.

- Add step in `MetaEstimator._preprocess_input` to mask coordinates if they fall outside of the analysis mask.

### DIFF
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -12,6 +12,7 @@ import inspect
 import numpy as np
 import pandas as pd
 from six import with_metaclass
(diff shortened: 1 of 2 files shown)


tsalo:
Per @angielaird, GingerALE actually _doesn't_ mask out coordinates before fitting the estimator, so I will close this.

Issue #218, opened 2020-05-19
Add nimads module with classes that mirror the spec


tyarkoni:
FYI @tsalo I've started working on a `nimads` module that provides lightweight data containers that mirror the spec and are backed by SQLAlchemy. The plan is to eventually split this out into its own package, which will interface directly with the central DB, as we've discussed in the past. But for now I'm just going to include it here, and then maybe rewrite the `Dataset` class so a lot of the `get()` methods operate over the `Study`, `Analysis`, etc. classes. This will make it easier to do lazy loading, study filtering, etc.

tsalo:
@tyarkoni is there any chance your `nimads` module code is still around? I'm wondering if it might be good to try working from that instead of starting from scratch.

tyarkoni:
IIRC I don't think I had a separate `nimads` module; what I had was built into the new Neurosynth back-end. I think @jdkent commented most of that code out, but it should still be in there. If you don't find anything that looks right, let me know, and I can check again.

tsalo:
Thanks! I'll follow up with James then.

jdkent:
ah, would this be the JSON-LD representation in neurostore?

Here is a commit with the original definitions:

https://github.com/neurostuff/neurostore/blob/26cd210a5ed6efbc7e8b761ff38dabfe95931524/neurostuff/schemas/schemas.py

jdkent:
this has been completed with the nimad module in NiMARE

PR #270, opened 2020-06-22
[REF] Convert Dataset attributes to properties

### LINKED ISSUE #256: Replace Dataset attributes with properties
Datasets have a number of important attributes that could be changed to properties to provide greater control.

tsalo:
Closes #256. I'm not sure how this will interact with #218, but at least for now properties seem perfect for these DataFrames.

This is a breaking change, as Datasets from <=0.0.3 will be incompatible with the new ones.

- Convert DataFrame attributes to properties.
- Move methods that could be functions to nimare.utils.
- Re-make test Dataset.
- Reorganize base classes.
- Lots of changes to docstrings and API.


tsalo:
@tyarkoni do you have any issues with the proposed change? I know that NIMADS integration will impact how datasets are stored, but properties seem pretty useful while we are still working with DataFrames.

tyarkoni:
Fine with me!

tsalo:
Awesome, thanks. I'll move forward with this then.

PR #272, opened 2020-06-23
[ENH] Add functions to convert image types

### LINKED ISSUE #170: Support transformations on Estimators' required inputs

### LINKED ISSUE #229: Improve handling of image spaces
The Dataset currently can transform coordinates between different spaces, but no effort is made to check or transform between standard spaces for images. 

At minimum, we could:
- Document the limitation in the Dataset docstring
- Check affines and raise warnings when the affines don't match.
- Apply `resample_to_img` to deal with resolution differences between dataset images and the selected template.

I don't think actually warping images is a reasonable goal right now, but we should at least make that clear to users.

tsalo:
Closes #170, closes #229.

- Add functions that can convert image types.
- Convert Dataset.masker to a property.
- Document the handling of image spaces.
- Change 'con' to 'beta' for beta-value images.

TODO:
- [x] Document new functions
- [x] Tests for new functions
- [x] Add example with transform
- [x] Fix shared path determination in image validation. Probably something to do with relative vs. absolute paths.


tsalo:
I just need to add tests, and then I can merge.

tsalo:
I'll add more tests over time, but for now I'm going to merge.

PR #273, opened 2020-06-24
[ENH] Use PyMARE for image-based meta-analyses

### LINKED ISSUE #108: Remove FSL dependency for IBMAs if possible
It would be preferable to limit NiMARE to Python dependencies, but we currently use FSL for the fixed and mixed effects GLM IBMAs. If possible, it would be good to find pure Python tools that we could use for this. Nistats may have the means to do this.

### LINKED ISSUE #211: Integrate PyMARE into image-based meta-analysis estimators
Continues from https://github.com/neurostuff/PyMARE/issues/31#issuecomment-626407883.

@tyarkoni Do you want to get this merged before 0.0.2? I remembered that there's a big bug in 0.0.1 and now that it should be fairly easy to deploy releases, I was thinking it would be good to release 0.0.2 soonish.

tsalo:
Closes #211, closes #108.

- Remove ESMA module.
- Use PyMARE for IBMA methods (except RFX GLM, for now).
- Add a new NiMARE test dataset made from the 21-pain studies NiMARE dataset. Instead of full images, however, we have 10x10x10 maps. Also includes transformed maps like z and varcope.
- Add mask and atlas images for new test dataset.


tsalo:
So far, the main blocker on this has been https://github.com/neurostuff/PyMARE/issues/42, but a simple hack is to just mask out any voxels where any of the input maps has a value of exactly 0. I'm going to push that for now, even though it's a suboptimal solution.

tsalo:
The IBMA tests are passing locally!!!! :tada: But we are going to either need a fixed release for PyMARE or pin the dependency to GitHub.

Also the outputs only grab FE stats, so if we really want the RE stats, that will need to be updated. I also haven't done anything as far as permutation tests of MCC of any kind. The standard NiMARE correctors (e.g., FDRCorrector) should work fine, but it's possible that we'll want some native correction methods implemented for the IBMA estimators (like how we have the `montecarlo` method for most of the CBMA estimators.

Issue #196, opened 2019-12-05
IBMAEstimator class organization


tyarkoni:
Probably a good time to talk about the organization of estimator classes. This applies to both CBMA and IBMA, but let's deal with IBMA first—I suspect there's more to talk about. Also tagging @nicholst and @cmaumet because this relates to the names/meaning of the estimators more than to the code architecture.

Currently we have the following IBMA estimators (some still in progress):
* Fisher's method
* Stouffer's z-score
* Weighted Stouffer's z
* Fixed effects (the canonical inverse-weighed method)
* Random effects (treats study as random, basically a t-test on studies)
* "fixed effects GLM" via FSL—I'm not sure how this differs from the canonical version
* "mixed effects GLM" via FSL—again I'm not sure what the difference is
* Mixed effects via Stan—gold standard, arbitrarily complex mixed-effects estimation

I may be missing something.

At minimum, I think we need to come up with more descriptive names for some of these. E.g., I don't like having `FFX` and `FFX_GLM` classes; this will be completely opaque to users. If the latter is specifically tied to FSL, we could maybe call it `FFX_FSL` (and ditto for `MFX_FSL`).

Alternatively, if the implementation differences are relatively minor, we could parameterize generic `FFX` and `MFX` classes with `method` and/or `estimator` arguments. So it would be somethiing like `est = FFX(method='fsl', ...). I think the choice here depends on just how different the internal model is, so it would be good to have input from Tom or Camille. It's also worth considering just getting rid of the FSL-based methods. I'm sure they do something slightly different, and possibly better, but the overhead is very high, and I'd prefer to encourage people to use standard, well-understood models. Is there any good reason to keep these around? 

Along similar lines, I see no reason to have separate `Stouffers` and `WeightedStouffers` methods/classes; IMO weights should just be an argument to `Stouffers`, much like the conventional fixed-effects models can take arbitrary weights, or weight by variance or sample size. We could even consider pushing Fisher's and Stouffer's into the `FixedEffects` class and specifying those via the `method` argument.

I guess my proposal is to consolidate/refactor into the following classes (I'm less concerned about the methods that the classes are wrapping):

* `Fishers` (maybe)
* `Stouffers` (maybe)
* `FixedEffects`
* `FSLFixedEffects` (maybe)
* `RandomEffects`
* `StanMixedEffects`
* `FSLMixedEffects` (maybe)

[...]

PR #282, opened 2020-07-12
[REF] Rename RandomEffectsGLM to TTest

tsalo:
Closes None, but references #196.

- Rename RandomEffectsGLM to TTest.
- Rename SampleSizeBased to SampleSizeBasedLikelihood.
- Rename rfx_glm to t_test.
- Reorganize tests.
- Bump up CBMA kernel coverage.


tyarkoni:
Sorry I'm late to this. Does the `TTest` estimator accept covariates? If yes, I think we should just call this `GLM`. If not, `TTest` is probably fine (provided it actually returns t statistics).

tsalo:
Not at the moment. I'm considering switching to using [`nilearn.mass_univariate.permuted_ols`](https://nilearn.github.io/modules/generated/nilearn.mass_univariate.permuted_ols.html), assuming it works similarly to the empirical null distribution method we have implemented, in which case covariates would be supported. I'll open an issue about it.

PR #300, opened 2020-08-08
[TST] Add flake8-black to test requirements

### LINKED ISSUE #298: Include black formatting evaluation in style check
I'd like to look into the [black code formatter GitHub Action](https://github.com/marketplace/actions/black-code-formatter). @akimbler is planning to test this out with his own repos, so we'll wait on what he has to say.

akimbler:
Closes #298 
* Adds `flake8-black` to test requirements.
* Adds Adam Kimbler to Zenodo file.


tsalo:
@akimbler Feel free to add a `pyproject.toml` with black config info in this PR. I don't think I'll get #302 working any time soon.

tsalo:
Okay I added `pyproject.toml` with #303. I decided to drop the `nimare/info.py` --> `setup.cfg` migration, since that's what was causing #302 to break.

setup.cfg
```diff
@@ -15,3 +15,12 @@ exclude = *build/,info.py,_version.py,due.py
 putty-ignore =
     */__init__.py : +F401
 ignore = E203,E402,E722,W503
+
+[options.extras_require]
```
tsalo, line 19: you can drop this section. `setup.cfg` is my enemy now.

akimbler: Done

tsalo:
@akimbler do you want to try just updating the flake8 settings to match black's recommendations? Like, ignore W503 but not W504, etc.?

akimbler:
@tsalo Where are the recommendations? We already ignore W503 (Line Break Before Operator) and E203 (No Space Before ':') which are the two I'm most familiar with

tsalo:
If you swap out flake8 with flake8-black in the style check CI job, what does it do again? Does it say _what_ black has a problem with, or just that "black would change this"?

akimbler:
BLK100 (Black would change this) is the only error flake8-black throws other than errors with black installation

tsalo:
I really don't like that. It's so uninformative that I think it would only make things worse. At least if it's run as a replacement for flake8. How about adding a flake8-black job to the CI? So we still have flake8 (and reduce our ignored warnings/errors to better match black) and we also run flake8-black. That way, if a PR passes flake8 but fails flake8-black, they will at least know that it's a more nuanced issue to be handled.

akimbler:
> I really don't like that. It's so uninformative that I think it would only make things worse. At least if it's run as a replacement for flake8. How about adding a flake8-black job to the CI? So we still have flake8 (and reduce our ignored warnings/errors to better match black) and we also run flake8-black. That way, if a PR passes flake8 but fails flake8-black, they will at least know that it's a more nuanced issue to be handled.

flake8-black isn't executable, it's just an additional set of tests for `flake`. What we could do is have a separate check with `flake8 --select BLK100` to limit the report to only `black` errors

tsalo:
Ohhh okay. So flake8-black _adds_ BLK100 to the existing flake8 checks? It doesn't replace them? E.g., if there's a W504, flake8 will still say W504, not BLK100.

tsalo:
Can you commit a couple of style mistakes so I can see how it will look? We can just revert them afterwards and the commit won't show up since we squash and merge.

akimbler:
Yes, my attempt at making a single W504 worked

tsalo:
Do you know what the BLK100 at 127 corresponds to?

tsalo:
Regardless, feel free to revert that change and I'll merge after. Thanks for doing this!

akimbler:
> Do you know what the BLK100 at 127 corresponds to?

It's specifically referring to the W504 error, it does the PEP8 supported break before the operator

PR #304, opened 2020-08-09
[ENH] Replace TTest with PermutedOLS

### LINKED ISSUE #285: Replace t_test with call to nilearn.mass_univariate.permuted_ols
Stems from https://github.com/neurostuff/NiMARE/pull/282#issuecomment-659540518. We should see if we can replace our own implementation of the "contrast permutation" method from [Maumet & Nichols (2016)](https://doi.org/10.1101/048249) with [`nilearn.mass_univariate.permuted_ols`](https://nilearn.github.io/modules/generated/nilearn.mass_univariate.permuted_ols.html). This would support covariates and would be more theoretically defensible than my implementation of the sign-flipping-based null distribution we currently have. Plus, NiMARE would just need a thin wrapper around the function for its IBMA estimator.

tsalo:
Closes #285. Requires the following:

- [x] https://github.com/nilearn/nilearn/issues/2555
- [x] https://github.com/nilearn/nilearn/pull/2559
- [ ] Nilearn release (0.6.3)

- Add a new IBMA method (with its own FWE correction!) called PermutedOLS, which uses `nilearn.mass_univariate.permuted_ols`.


tsalo:
I've pinned to a nilearn commit and I'll switch to a minimum version when they make their next release.

PR #315, opened 2020-08-23
[FIX] Remove base class get/set params methods

tsalo:
Closes None. These methods are not compatible with **kwargs and *args, which are now used in several major classes.

- Remove `NiMAREBase.get_params()`
- Remove `NiMAREBase.set_params()`
- Remove `NiMAREBase._get_param_names()`


tsalo:
Apparently we need `get_params` (at least for now). I'll close this and may revisit at a later date.

Issue #195, opened 2019-12-05
Clean up masking/preprocessing in CBMAEstimator hierarchy


tyarkoni:
There's some redundancy in the masking/preprocessing done inside the `CBMAEstimator` subclasses that makes maintenance more difficult than it could be. We should probably clean this up and push as much of this stuff as possible into the base `CBMAEstimator` class. I realize there's some heterogeneity in `fit()` signature that makes things a bit more complex, but probably we can handle that with either mixins or by adding something like a `CBMAPairwiseEstimator` child that expects two inputs.

Another thing I'm not super clear on is why we have a lot of calls of the form `self.kernel_estimator.transform(self.dataset, mask=self.mask, masked=True)`. IMO the masking parameters should be set on the `KernelTransformer` at initialization time, as with other classes. I think we should apply the same principle as elsewhere: if the `KernelTransformer` isn't initialized with a masker, then it will always take it from the `Dataset`. Since the dataset is always passed through from the `CBMAEstimator`, I think the above call can then be reduced to `self.kernel_estimator.transform(self.dataset)`, which is a much cleaner API and will cut down on a lot or redundancy. (As a minor aside, shouldn't it be `CBMAEstimator.kernel_transformer` rather than `CBMAEstimator.kernel_estimator`, for consistency with the `KernelTransformer` class?)

This probably shouldn't be touched until we merge #163 (which I'm working on right now), but just bringing it up here for comment.

tsalo:
> I realize there's some heterogeneity in fit() signature that makes things a bit more complex, but probably we can handle that with either mixins or by adding something like a CBMAPairwiseEstimator child that expects two inputs.

Sounds good to me. 

> Another thing I'm not super clear on is why we have a lot of calls of the form self.kernel_estimator.transform(self.dataset, mask=self.mask, masked=True). IMO the masking parameters should be set on the KernelTransformer at initialization time, as with other classes.

KernelEstimators can return either lists of images (`masked=False`) or numpy arrays (`masked=True`) mostly for speed. That speed is important for the permutation-based FWE correction methods. The kernel estimator is initialized and stored as an Estimator parameter in the Estimator's initialization. We could move the `masked` argument from `KernelTransformer.transform()` to `KernelTransformer.__init__()`, but then we'd need to commit to having it return arrays throughout the Estimator rather than just at the FWE-correction stage.

[...]

PR #320, opened 2020-08-27
[ENH] Support Dataset transformations in kernel transformers

### LINKED ISSUE #41: Have kernel classes return `Dataset` instances
The API might be more intuitive if the `Kernel` classes returned a `Dataset` instance, with the resulting images appended to the `.images` list of every `Contrast`. Per discussion with @tsalo, the internal logic of the `.fit` calls could move to module level for efficiency (e.g., if one needs to build up a null distribution of 10,000 sets off images without copying the `Dataset` that many times in memory).

### LINKED ISSUE #307: Add path attribute to Dataset
This stems from https://github.com/neurostuff/NiMARE/issues/41#issuecomment-674142973. To make it easier to create and save images from purely coordinate-based Datasets, we can add a `path` attribute to Datasets.

On the bright side, we can drop the "absolute path" columns from `Dataset.images`. The assumption would be that every absolute path is just `Dataset.path` + the image's relative path.

tsalo:
Closes #41 and closes #307 and references #195.

- Add copy method to Dataset so that kernel transformers, when returning Datasets, don't update the original object.
- Track image basepath in Dataset.
- Support Dataset transformation in kernel transformers. This transformation creates an unmasked version of the MA map, saves it to a file based on `Dataset.basepath` and the parameters of the transformer, and stores the path in the `Dataset.images` attribute.
- Fix bug in `nimare.stats.null_to_p()`. Values outside the range of the null distribution (i.e., p = 0) are now cropped down to epsilon for float.

To do:
- [x] Do not mask MA maps when saving to files. This way they're template-, but not mask-, specific.
- [x] Load images from files when possible. Use masker to apply the mask.
- [ ] Move as much as possible to the base class.
- [ ] Drop absolute paths from images DataFrame. The new basepath attribute should be sufficient.
- [x] Update tests.

Something to perhaps keep an eye on is that there was a new overflow error in one of the MKDA tests. Since these changes shouldn't impact behavior, I think the new error (which reflects a bug in `null_to_p`), probably stems from a random seed thing.

### DIFF
--- a/examples/02_meta-analyses/plot_generate_ma_maps.py
+++ b/examples/02_meta-analyses/plot_generate_ma_maps.py
@@ -39,13 +39,13 @@ dset = nimare.dataset.Dataset(dset_file)
 # argument to control the radius of the kernel.
 
 kernel = nimare.meta.kernel.MKDAKernel(r=2)
-mkda_r02 = kernel.transform(dset)
+mkda_r02 = kernel.transform(dset, return_type="image")
 kernel = nimare.meta.kernel.MKDAKernel(r=6)
-mkda_r06 = kernel.transform(dset)
+mkda_r06 = kernel.transform(dset, return_type="image")
 kernel = nimare.meta.kernel.MKDAKernel(r=10)
-mkda_r10 = kernel.transform(dset)
+mkda_r10 = kernel.transform(dset, return_type="image")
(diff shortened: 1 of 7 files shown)


tsalo:
I'm going to hold off on abstracting things to the base class until later.

PR #343, opened 2020-09-22
[ENH] Add generate module

### LINKED ISSUE #308: New module for simulating data
This stems from https://github.com/neurostuff/NiMARE/issues/241#issuecomment-638967169 (sorry @tyarkoni, I only just caught this part of your comment).

> On a related note, I'm thinking it might not be a bad idea to add a module for fake data generation (and maybe some associated functions to run Monte Carlo simulations). We should probably be systematically evaluating our estimators before/as we add them. (Also, this could be easily turned into a paper at some point.)

jdkent:
closes #308 

- add `generate.py` to create coordinate based datasets for meta-analysis 
- add `test_generate.py`
- add `test_estimator_performance.py` for example on how to test different datasets with different estimators

## Open Questions/Thoughts

- Does the general approach for generating coordinate based datasets look sound (in `generate.py`)?
  - one weird thing I'm leveraging is `ma_compute` to create a probability function to select foci for specific studies
  - other weird things?
- What kinds of outputs should be measured/observed to see if the estimator is "working" and performing adequately?
  - not all multiple correction transformers have a p-value output.
  - some measure of a false positive rate
- How should I generalize to image-based meta-analysis?
- How should I profile time/memory usage?
  - potentially [`pytest-profiling`](https://pypi.org/project/pytest-profiling/)
- create a method to generate random foci between studies (false positives)?

I'll be editing this text to provide more context.  
This pull request is a WIP, I'm opening early so @tsalo and @tyarkoni (and anyone else interested) can provide feedback so I do not stray too far away from what this pull request _should_ do.

use cases being supported:
- [x] generating arbitrary numbers of foci for an arbitrary number of studies
- [x] generating a focus only present in some of the studies.
- [x] having 2 (or more) foci with different probabilities of being selected in a study
- [x] have 1 study with a large number of reported foci and other studies without
- [x] have 1 study with a large sample size and other studies without
- [x] varying the number of noise foci to be reported in each study
- [x] creating only noise foci (kind of)
- [x] varying the spread of study foci around the "ground truth focus"

use cases not supported:
- [ ] reading/permuting input datasets to generate simulated coordinates from observed values
- [ ] user-generated masks to select a portion of the brain
- [ ] specifying coordinates for noise_foci
- [ ] creating coordinates in non-MNI space
- [ ] image-based simulations


tyarkoni:
@jdkent This looks great! Comments below.

> Does the general approach for generating coordinate based datasets look sound (in generate.py)?

LGTM, with the caveat that it's hard to get a qualitative sense of what the data generated by `create_coordinate_dataset()` actually look like. Maybe you could create a demo notebook (@tsalo, any thoughts on where you want it to go?) that plots (a) the generated data (or at least a subset of it) and then (b) the resulting statistical maps when run through a CBMA estimator? I think this would help us get a better sense of whether everything is kosher, though from the looks of the code (and the test), it seems sound.

> one weird thing I'm leveraging is ma_compute to create a probability function to select foci for specific studies
other weird things?

Seems fine to me.

> What kinds of outputs should be measured/observed to see if the estimator is "working" and performing adequately?

In addition to the existing sensitivity test at the ground truth location, we could add some specificity checks to the test—i.e., to verify that there *isn't* a significant result almost anywhere else. An expedient way to do this would be to just select a few random coordinates that are not very close to the ground truth sources, and verify that p > .05. A more robust approach would be to look at the p-values for all voxels in the brain that are not spatially close to the ground truth sources. I.e., you could generate a large-ish ROI mask around the true sources, and then verify that the proportion of voxels outside the mask that show a significant effect is roughly .05 (i.e., the method controls the false positive rate approximately as expected). For this test, you might want to repeatedly generate datasets and calculate the false positive rate over all datasets. Plotting a histogram in the notebook suggested above might also help give us a sense of how an estimator is performing (though the critical thing is that the mean be close to 0.05).

> not all multiple correction transformers have a p-value output.

That should be fine as long as they still output an image of voxels that survive correction; then you can just mask that image with the set of voxels that are far from true sources, and the proportion of significant voxels in the result is your false positive rate.

> some measure of a false positive rate

Yup, see above.

> How should I generalize to image-based meta-analysis?

Good question. Eventually we'll probably want to test these with more realistic ground truth maps where each voxel has some non-zero population value and SD, but for now I think it's fine to stick with the sparse sources. A reasonable first pass would be to add a `create_maps()` function to complement `create_foci()`, where the former would return entire maps rather than study-specific peaks. You could probably use the same general approach—i.e., convolve the study-specific peak with a gaussian, and then return the resulting image (where most of the voxels will have a value of zero). Because the voxel values will now have magnitude information, you may want to add a parameter or two that control the variance in the peak value (i.e., some study maps should contain stronger effects than others).

I wouldn't spend a lot of time at this point trying to come up with a realistic generation procedure; let's first get a general workflow for evaluation and testing (particularly, quantification of error rates) figured out, and then we can iterate and figure out how to produce IBMA meta-analysis datasets that look more realistic.

> How should I profile time/memory usage?

Let's start simple. I think just recording the total time taken to run each combination of estimator/corrector, and maybe max memory consumption, would be great. You could lean heavily on packages like `line_profiler` and `memory-profiler`—which would also make it easier to build fancier reports later on.

> potentially pytest-profiling

Haven't used or looked at it before, but if it does what you need it to, sure!

> create a method to generate random foci between studies (false positives)?

Great idea. But I'd probably defer that until we're satisfied with the basic framework for quantifying and reporting error rates. At that point we can iterate on the data-generating utilities.

Nice work!

tsalo:
> Maybe you could create a demo notebook (@tsalo, any thoughts on where you want it to go?) that plots (a) the generated data (or at least a subset of it) and then (b) the resulting statistical maps when run through a CBMA estimator?

I would definitely like to see this used in testing. Not just unit testing of the module itself, but also integration testing with the CBMA estimators. If memory usage is not too high, an example would be nice. RTD is limited though, so if memory usage is too high then I think a notebook is a good place for it.

jdkent:
use cases being supported:
- [x] generating arbitrary numbers of foci for an arbitrary number of studies
- [x] generating a focus only present in some of the studies.
- [x] having 2 (or more) foci with different probabilities of being selected in a study
- [x] have 1 study with a large number of reported foci and other studies without
- [x] have 1 study with a large sample size and other studies without
- [x] varying the number of noise foci to be reported in each study
- [x] creating only noise foci (kind of)
- [x] varying the spread of study foci around the "ground truth focus"

use cases not supported:
- [ ] user-generated masks to select a portion of the brain
- [ ] specifying coordinates for noise_foci
- [ ] creating coordinates in non-MNI space

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # assert that more than 95% of voxels farther away
+        # from foci are nonsignificant at alpha = 0.05
+        mni_nonzero = np.nonzero(mni_mask.get_fdata())
+        observed_nonsig = p_values_data[mni_nonzero][nonsig_regions[mni_nonzero]] > 0.05
+        assert observed_nonsig.sum() / len(observed_nonsig) >= 0.95
+
+
+def _create_null_mask(mni_mask, ground_truth_foci_ijks, fwhm):
+    threshold_percentage = 90
```
jdkent, line 84: I cannot think of a principled threshold that _should_ be used to distinguish between voxels we expect to be significant versus non-significant. Currently using 90% as an arbitrary threshold.

nimare/generate.py
```diff
@@ -0,0 +1,269 @@
...
+                        "sample_sizes": [sample_size],
+                    },
+                }
+            }
+        }
+
+    return source
+
+
+def create_foci(
```
jdkent, line 127: I do a lot of type checking here, I'm curious if there is any preferred way to validate inputs to a function.

examples/02_meta-analyses/generate_data.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+###############################################################################
+# Create Dataset
+# --------------------------------------------------
+# In this example, each of the 30 fake studies being generated
+# select 4 coordinates from a probability map representing the probability
+# that particular coordinate will be chosen.
+# There are 4 "hot" spots centered on 3D gaussian distributions.
+# Each study has a mean sample_size of 30
```
jdkent, line 52: I'm not confident in the language I'm using here describing how the simulated data are created, suggestions for rephrase or what seems unclear?

nimare/generate.py
tsalo: If there's one valid type for each input we could use `typing`, since NiMARE supports Python 3.6+.

nimare/tests/test_estimator_performance.py
tsalo: I think in Eickhoff et al. (2015) the threshold for a true result was 20% of experiments with a peak at the target, with random spatial noise added. Is that similar to the threshold you're referring to here?

nimare/generate.py
jdkent: sadly I've made it more complicated than that, there are multiple valid types for inputs.

tsalo: Yes, that is also my general strategy throughout the rest of NiMARE.

nimare/tests/test_estimator_performance.py
jdkent: probably related, I'll try to be more clear (please excuse my partial understandings)

`compute_ma` returns a voxelwise probability map that represents the probability of convergence in each voxel.
I'm thresholding that map at a percentage (e.g., 90%), so I have the top 10% of voxels that should show convergence in the meta-analysis. 

The top 10% of voxels are then compared to the actual p-values in the meta-analysis results to see how many voxels within the top 10% are significant at an alpha of 0.05.

jdkent:
This now tests many combinations of meta-analysis estimators, kernels, and correctors for CBMA (48 combinations total) on two simulated datasets resulting in 96 additional tests.

There are several combinations of these objects that fail, and they are noted with explicit if/then clauses within the test.
related to #344

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # peaks2MapsKernel does not work with any meta-analysis estimator
+    if isinstance(kern, kernel.Peaks2MapsKernel):
+        # AttributeError: 'DataFrame' object has no attribute 'masker'
+        meta_expectation = pytest.raises(AttributeError)
+    elif isinstance(kern, kernel.KDAKernel) and isinstance(meta, ale.ALE):
+        # IndexError: index 20000 is out of bounds for axis 0 with size 10010
+        meta_expectation = pytest.raises(IndexError)
+    else:
+        meta_expectation = does_not_raise()
```
tsalo, line 56: This is extremely helpful. Thank you!

nimare/generate.py
```diff
@@ -0,0 +1,269 @@
...
+        (must be >=0).
+
+    Returns
+    -------
+    ground_truth_foci : :obj:`list`
+        generated foci in xyz (mm) coordinates
+    dataset : :class:`nimare.Dataset`
+    """
+    if rng is None:
+        rng = np.random.RandomState(seed=1939)
```
jdkent, line 66: This could be controversial if people want to have more psuedo-random behavior, as this currently stands, if you call this function with the same inputs, you get the same output deterministically, so if you wanted to simulate a bunch of datasets with the same properties, you would need to pass in a different random state.

Another note, I could just have the user pass in a seed instead of an entire randomState object, that appears more reasonable.

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+        or a list specifying the sample size for each
+        study.
+    studies : :obj:`int`
+        Number of studies to generate.
+    foci_coords : :obj:`list` of three item array_like objects (Optional)
+        Ground truth coordinates of foci.
+        If not specified, coordinates will be randomly selected.
+        If not the same length as the number of foci (i.e., `foci_num`),
+        foci will either be over-sampled or under-sampled.
+    foci_noise : :obj:`int` or :obj:`list`
```
tsalo, line 46: I think `n_noise_foci` would be clearer.

tsalo: Also, this is missing `, optional` at the end and the description needs the default.

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+    fwhm : :obj:`float`
+        Full width at half maximum to define the probability
+        spread of the foci.
+    sample_size : :obj:`int` or :obj:`list`
+        Either mean number of participants in each study
+        or a list specifying the sample size for each
+        study.
+    studies : :obj:`int`
+        Number of studies to generate.
+    foci_coords : :obj:`list` of three item array_like objects (Optional)
```
tsalo, line 41: Can you switch from `... (Optional)` to `..., optional` to match `numpydoc`?

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+    foci_num : :obj:`int` or :obj:`list`
+        The number of foci to be generated per study
+    fwhm : :obj:`float`
+        Full width at half maximum to define the probability
+        spread of the foci.
+    sample_size : :obj:`int` or :obj:`list`
+        Either mean number of participants in each study
+        or a list specifying the sample size for each
+        study.
+    studies : :obj:`int`
```
tsalo, line 39: I think `n_studies` would be clearer.

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+    foci_noise=0,
+    foci_weights=None,
+    seed=42,
+    sample_size_variance=None,
+):
+    """Generate coordinate based dataset for meta analysis.
+
+    Parameters
+    ----------
+    foci_num : :obj:`int` or :obj:`list`
```
tsalo, line 30: I'd prefer `n_foci`.

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+    dataset : :class:`nimare.Dataset`
+    """
+    rng = np.random.RandomState(seed=seed)
+    if isinstance(sample_size, list):
+        sample_sizes = sample_size
+        if len(sample_sizes) != studies:
+            raise ValueError("sample_size must be the same length as studies")
+    else:
+        if sample_size_variance is None:
+            raise ValueError("sample_size_variance must be specified")
```
tsalo, line 72:
```suggestion
            raise ValueError("sample_size_variance must be specified if sample_size is not provided for each study")
```

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+        sample_size_lower_limit = int(sample_size - sample_size_variance)
+        # add 1 since randint upper limit is a closed interval
+        sample_size_upper_limit = int(sample_size + sample_size_variance + 1)
+        sample_sizes = rng.randint(sample_size_lower_limit, sample_size_upper_limit, size=studies)
```
tsalo, line 76: This is a little confusing. Based on the variable name and description, I assumed that `sample_sizes` would be drawn from a normal distribution (or at least one with some definition of "variance").

tsalo: Just to be clear- I don't mean you should switch to a Gaussian distribution. Maybe just change the variable name or description to make it clearer that it's half the range of a uniform distribution.

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+        Ground truth coordinates of foci.
+        If not specified, coordinates will be randomly selected.
+        If not the same length as the number of foci (i.e., `foci_num`),
+        foci will either be over-sampled or under-sampled.
+    foci_noise : :obj:`int` or :obj:`list`
+        Number of foci considered to be noise in each study
+        or a list of integers representing how many noise focis
+        there should be in each study.
+    foci_weights : :obj:`list` (Optional)
+        Weighing of each foci representing the probability of
```
tsalo, line 51:
```suggestion
        Weighting of each focus representing the probability of
```
"focus" is pretty uncommon in the literature, but "peak" might be a good substitute for the singular of "foci".

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+        Full width at half maximum to define the probability
+        spread of the foci.
+    sample_size : :obj:`int` or :obj:`list`
+        Either mean number of participants in each study
+        or a list specifying the sample size for each
+        study.
+    studies : :obj:`int`
+        Number of studies to generate.
+    foci_coords : :obj:`list` of three item array_like objects (Optional)
+        Ground truth coordinates of foci.
```
tsalo, line 42: Can you specify that they're XYZ coordinates? Wasn't sure if they should be IJK array indices.

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+
+    Parameters
+    ----------
+    foci_num : :obj:`int` or :obj:`list`
+        The number of foci to be generated per study.
+    fwhm : :obj:`float` or :obj:`list`
+        Full width at half maximum (fwhm) to define the probability
+        spread of the foci or list of fwhms identifying the fwhm
+        for each foci.
+    studies : :obj:`int`
```
tsalo, line 146: I think `n_studies` is clearer.

nimare/generate.py
```diff
@@ -0,0 +1,267 @@
...
+
+    if space == "MNI":
+        template_img = nib.load(
+            os.path.join(get_resource_path(), "templates", "MNI152_2x2x2_brainmask.nii.gz")
+        )
+    else:
+        raise NotImplementedError("Only coordinates for the MNI atlas has been defined")
+
+    # use a template to find all "valid" coordinates
+    template_data = template_img.get_data()
```
tsalo, line 184:
```suggestion
    template_data = template_img.get_fdata()
```
`get_data` is deprecated.

tsalo (commented):
I have some initial thoughts on variable names and formatting, but I haven't dug into the actual simulation procedure. I plan to do that tomorrow or early next week.

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,149 @@
...
+        corr_expectation = pytest.raises(ValueError)
+    elif isinstance(meta, mkda.KDA) and isinstance(corr, FDRCorrector):
+        corr_expectation = pytest.raises(ValueError)
+    elif isinstance(meta, mkda.MKDADensity) and corr.method == "bonferroni":
+        corr_expectation = pytest.raises(ValueError)
+    elif isinstance(meta, mkda.KDA) and corr.method == "montecarlo":
+        # TypeError: correct_fwe_montecarlo() got an unexpected keyword argument 'voxel_thresh'
+        corr_expectation = pytest.raises(TypeError)
+    elif isinstance(meta, mkda.KDA) and corr.method == "bonferroni":
+        corr_expectation = pytest.raises(ValueError)
```
tsalo, line 76: @jdkent I think these should all be considered bugs. Would you be willing to open a separate issue for each of them? Preferably with the full tracebacks.

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,149 @@
...
+    # set up testing dataset
+    ground_truth_foci, dataset = simulatedata_cbma
+
+    # create meta-analysis
+    meta = meta_alg(kern)
+
+    # peaks2MapsKernel does not work with any meta-analysis estimator
+    if isinstance(kern, kernel.Peaks2MapsKernel):
+        # AttributeError: 'DataFrame' object has no attribute 'masker'
+        meta_expectation = pytest.raises(AttributeError)
```
tsalo, line 51: I've opened #347 about this.

nimare/tests/test_estimator_performance.py
jdkent: didn't see the full traceback part, but the majority of errors come back [to here](https://github.com/neurostuff/NiMARE/blob/58e6e82e3ca15ae89390bbf4c294b44f689bd5a0/nimare/correct.py#L64-L69).

nimare/generate.py
jdkent: thanks! incorporated this change.

nimare/generate.py
jdkent: I agree, changed it.

nimare/generate.py
jdkent: good catch, I was on the fence for a while so I didn't commit what the coordinates should represent, specified xyz explicitly now.

nimare/generate.py
jdkent: Thank you, is foci commonly used in the literature, or would peaks be better for the plural?

nimare/generate.py
jdkent: thanks for catching this, I switched the name to `sample_size_interval` and tried to give a more detailed description, how does that sound?

nimare/generate.py
jdkent: done!

nimare/generate.py
jdkent: yeah, I do too, changed it.

nimare/generate.py
jdkent: me too, changed it.

nimare/generate.py
jdkent: changed to match numpydoc.

nimare/generate.py
jdkent: to add it to the description, I would say: `(default=0)` in the description or something else?

tsalo: That sounds good. :+1:

nimare/generate.py
tsalo: Foci is used enough that _I_ consider it interpretable (and indeed use it a lot in the NiMARE manuscript), but TBH peaks might be a little clearer for other folks.

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,157 @@
...
+    return {
+        "meta": meta,
+        "kernel": kern,
+        "corr": corr,
+        "sensitivity": observed_sig_perc,
+        "specificity": observed_nonsig_perc,
+    }
+
+
+def _create_null_mask(mni_mask, ground_truth_foci_ijks, fwhm):
```
jdkent, line 150: call this create_signal_mask

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,157 @@
...
+    observed_nonsig = p_values_data[mni_nonzero][nonsig_regions[mni_nonzero]] > 0.05
+    observed_nonsig_perc = observed_nonsig.sum() / len(observed_nonsig)
+    assert observed_nonsig_perc >= 0.95
+
+    # TODO: use output in reports
+    return {
+        "meta": meta,
+        "kernel": kern,
+        "corr": corr,
+        "sensitivity": observed_sig_perc,
```
jdkent, line 145: call this `estimated_power`

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,157 @@
...
+        pytest.param(kernel.KDAKernel(), id="kda_kernel"),
+        pytest.param(kernel.Peaks2MapsKernel(), id="p2m_kernel"),
+    ],
+)
+@pytest.mark.parametrize(
+    "corr",
+    [
+        pytest.param(FWECorrector(method="bonferroni"), id="fwe_bonferroni"),
+        pytest.param(
+            FWECorrector(method="montecarlo", voxel_thresh=0.001, n_iters=10, n_cores=1),
```
jdkent, line 33: change to `0.05`

nimare/generate.py
tsalo: I don't see any problem with defaulting to a specific seed, although I think it would be better to have the seed be the default value in the actual function definition.

nimare/generate.py
tsalo: That sounds great!

jdkent:
Alright, in this latest sequence of updates I:
- renamed variables from review
- made it more straightforward to generate a dataset with no converging foci
  - and added a section in the [example notebook demonstrating this functionality](https://2975-117724523-gh.circle-artifacts.com/0/tmp/src/NiMARE/docs/_build/html/auto_examples/02_meta-analyses/plot_simulated_data.html#sphx-glr-auto-examples-02-meta-analyses-plot-simulated-data-py)
- use `pytest.xfail` in `test_estimator_performance.py` to make it more obvious which meta/kernel/corrector combinations have potential issues
- minor bugfixes/reformatting

There are other items on the wishlist for this module, but I think it has reached a point where it is useful as is.
@tyarkoni and @tsalo, if you believe I've responded to your reviews adequately, this is ready to go.

nimare/meta/utils.py
```diff
@@ -266,6 +266,25 @@ def generate_input_fn():
     return niis
 
 
+def compute_kda_ma(shape, vox_dims, ijks, r, value, allow_overlap=False):
```
tsalo, line 269: Needs a docstring.

examples/02_meta-analyses/plot_simulated_data.py
```diff
@@ -0,0 +1,228 @@
...
+# This way you can test what happens if one study has way more reported
+# foci than another study.
+# You can also:
+#   - manually specify the sample sizes for each study (i.e., ``sample_size``).
+#   - set the number of false positive foci for each study (i.e., ``n_noise_foci``).
+#   - set the weighed probability of each foci being chosen (i.e., ``foci_weights``).
+#   - set the dispersion of the gaussian spread for each foci (i.e., ``fwhm``).
+#   - set the specific ground truth foci (i.e., ``foci_coords``).
+
+num_studies = 30
```
tsalo, line 97: I think this was just missed in the recent set of changes.

examples/02_meta-analyses/plot_simulated_data.py
```diff
@@ -0,0 +1,228 @@
...
+# --------------------------------------------------------------------
+# Another method to control what percentage of studies contain
+# a focus is to have one focus specified in ``foci_coords``
+# and have a list of zeros and ones in ``n_foci`` to specify
+# which studies contain the focus of interest.
+# The caveat of this strategy is that each study must have at least
+# one coordinate to report necessitating ``n_noise_foci`` be set for
+# studies without a ground truth foci.
+
+num_studies = 30
```
tsalo, line 165: I think this was just missed in the recent set of changes.

examples/02_meta-analyses/generate_data.py
tsalo: I think it's fairly clear. We can improve later, but I don't think it should be a blocker for this PR.

tsalo (changes requested):
There are a few minor things that need to be handled, but I agree that the overall method is solid and mergeable at this point. We can always tune the simulation procedure later.

examples/02_meta-analyses/plot_simulated_data.py
jdkent: fixed!

examples/02_meta-analyses/plot_simulated_data.py
jdkent: fixed!

nimare/meta/utils.py
jdkent: added a docstring

nimare/meta/kernel.py
```diff
@@ -436,18 +428,15 @@ def transform(self, dataset, masker=None, return_type="dataset"):
...
-                sphere = sphere[idx, :].astype(int)
-                kernel_data[tuple(sphere.T)] += self.value
+            ijks = np.vstack((data.i.values, data.j.values, data.k.values)).T
+            kernel_data = compute_kda_ma(
+                dims,
+                vox_dims,
+                ijks,
+                r=self.r,
+                value=self.value,
+                allow_overlap=True,
```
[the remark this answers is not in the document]
jdkent: yeah, I was thinking about this too, `sum_overlap` does make more sense, thanks!

tsalo (approved):
Everything looks good to me. Thanks for your patience in working through all of my nitpicky requests.

jdkent:
@tyarkoni, I think after your review (and potential fixes) we can merge this.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+    weighted_prob_map = sum(
+        [
+            compute_ma(template_data.shape, np.atleast_2d(ground_truth_focus), kernel) * weight
+            for kernel, ground_truth_focus, weight in zip(
+                kernels, ground_truth_foci_ijk, foci_weights
+            )
+        ]
+    )
+    # handle case when sum is 0 by adding a small amount
+    weighted_prob_map = weighted_prob_map / np.sum(weighted_prob_map) + DEF_TINY
```
tyarkoni, line 262: I find this a bit confusing... Could there actually be a situation where the sum is `0`? Not obvious to me what that would be, so I'm not sure this is needed. Also, I think the way this is written right now, the results will always be infinite. You could wrap the addition operation in parentheses to ensure correct precedence, but then I think you'd still get an absurdly large number.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+    weighted_prob_map_ijks = np.argwhere(weighted_prob_map)
+    weighted_prob_vector = weighted_prob_map[np.nonzero(weighted_prob_map)]
+
+    if isinstance(n_noise_foci, int):
+        n_noise_foci_list = [n_noise_foci] * n_studies
```
tyarkoni, line 267: I think you can just overwrite `n_noise_foci`, it doesn't seem like it's re-used.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+        n_noise_foci_list = n_noise_foci
+
+    for f_num, f_noise, study in zip(n_focis, n_noise_foci_list, range(n_studies)):
+        ijk_idxs = rng.choice(
+            weighted_prob_map_ijks.shape[0], f_num, p=weighted_prob_vector, replace=False
+        )
+        foci_ijks = weighted_prob_map_ijks[ijk_idxs]
+
+        if f_noise > 0:
+            weighted_noise_map_ijks = np.delete(weighted_prob_map_ijks, ijk_idxs, axis=0)
```
tyarkoni, line 278: If I understand this correctly, I'm not sure we need/want this... Noise could in theory arise in the same coordinates as signal, so I don't think we should forcibly exclude signal voxels from contention.

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,268 @@
...
+import nibabel as nib
+
+from ..correct import FDRCorrector, FWECorrector
+from ..meta import ale, kernel, mkda
+from ..meta.utils import compute_kda_ma
+from ..transforms import mm2vox
+
+# set significance levels used for testing.
+ALPHA = 0.05
+BETA = 1 - ALPHA
```
tyarkoni, line 14: Not sure if it matters in this context, but beta is usually defined as the complement of power, not alpha. If the quantity you actually want is `1 - ALPHA`, let's call that the `TRUE_NEG` rate or something like that.

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,268 @@
...
+    # are significant at alpha = .05
+    observed_sig = p_values_data[sig_idx] < ALPHA
+    observed_sig_perc = observed_sig.sum() / len(observed_sig)
+    assert (observed_sig_perc >= 0.5) == good_performance
+
+    # assert that more than 95% of voxels farther away
+    # from foci are nonsignificant at alpha = 0.05
+    observed_nonsig = p_values_data[nonsig_idx] > ALPHA
+    observed_nonsig_perc = observed_nonsig.sum() / len(observed_nonsig)
+    assert observed_nonsig_perc >= BETA
```
tyarkoni, line 219: Have not read very carefully, but unless `ALPHA` is set way lower than 0.05, I don't think this is a safe assertion... in the long run I think `observed_nonsig_perc` should be equal to what you're calling `BETA` (the true negative rate), but in the short run it could be lower than that. So probably we want to check that it's reasonably close to the true negative rate, where "reasonably close" can probably be defined via trial and error.

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,268 @@
...
+    else:
+        p_values_data = cres.get_map("p", return_type="array")
+
+    # ensure all values are between 0 and 1 (inclusively)
+    assert (p_values_data >= 0).all() and (p_values_data <= 1).all()
+
+    ################################################################
+    # CHECK IF META/KERNEL/CORRECTOR COMBINATION PERFORMS ADEQUATELY
+    ################################################################
+    # The below combinations cannot even detect
```
tyarkoni, line 152: This worries me a bit... I don't see an obvious reason why these combinations should produce markedly different results. @tsalo, am I missing something?

If the issue is that we just happen to have titrated the sampling procedure so that some estimators detect effects but others don't, we should probably change it so that the signal is unambiguous. But it seems unlikely we would just randomly stumble into such a regime, so it's worth looking into what's going on (assuming I'm not misunderstanding something or missing something obvious). If it's just that the monte carlo correction procedure is more conservative, we should probably boost the signal to make sure that we still detect it. But I would still worry if some estimator/kernel combinations "work" with monte carlo correction and others don't.

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,268 @@
...
+
+    # double the radius of what is expected to be significant
+    r = fwhm
+    # create an expectation of significant/non-significant regions
+    sig_map, nonsig_map = _create_signal_mask(mask, ground_truth_foci_ijks, r=r)
+    sig_idx = cres.masker.transform(sig_map).astype(bool).squeeze()
+    nonsig_idx = cres.masker.transform(nonsig_map).astype(bool).squeeze()
+    # while this combination passes the basic
+    # test of significance at the ground truth foci voxels,
+    # it does not meet the criteria for 50% power
```
tyarkoni, line 196: We should probably set two thresholds then: a narrower radius where we're confident we can detect effects, and a looser one outside of which we're confident we won't detect effects (we can ignore the middle zone).

(Alternatively, we can boost the signal or reduce the variability until this is a non-issue.)

nimare/tests/test_estimator_performance.py
tyarkoni: Oh, unless this is still testing with 0 noise coordinates. If that's the case, then I think this is probably fine—because unless I'm mistaken, we should get tiny p-values everywhere (because there's no noise, so the p-values should be either 0 or the smallest possible permutation-based value). But we should definitely include at least one test that injects noise into the simulated data, so we can make sure that the false positive rate is maintained at a sane level outside areas where there's signal.

tyarkoni (changes requested):
Overall, looks great! I did leave a bunch of comments, most of which you're free to ignore for now. There are also some issues I raised that may reflect my misunderstanding of what's going on. We can talk it over tomorrow, @jdkent.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+
+DEF_TINY = np.finfo(float).eps
+
+
+def create_coordinate_dataset(
+    n_foci,
+    fwhm,
+    sample_size,
+    n_studies,
+    foci_coords=None,
```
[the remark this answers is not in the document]
jdkent: My thought for oversampling came from [Wager, Lindquist, & Kaplan (2007)](https://academic.oup.com/scan/article/2/2/150/2736775), specifically to measure if one study has an outsized impact on the meta-analysis because of the number of peaks it contributes:
> Study 1 (top) contributes 26 peaks to the metaanalysis, many of them very close together, whereas Study 2
(middle) contributes only two.

nimare/tests/test_estimator_performance.py
tsalo: I'd imagine that ALE+MKDAKernel is pretty close to invalid, even if it successfully runs, so I'm not really surprised by any particular results with that combination. I'm also not surprised by the Bonferroni results of the MKDA+KDAKernel or MKDA+MKDAKernel, but I _am_ surprised that those two don't find anything with FDR correction...

tsalo: BTW I think the Monte Carlo procedure is _less_ conservative (at least with 100 iterations), if I'm reading this section correctly.

nimare/tests/test_estimator_performance.py
```diff
@@ -0,0 +1,268 @@
...
+    # The below combinations cannot even detect
+    # significance at the ground truth foci voxel
+    # locations
+    if (
+        (
+            isinstance(meta, mkda.MKDADensity)
+            and isinstance(kern, kernel.KDAKernel)
+            and corr.method != "montecarlo"
+        )
+        or (
```
jdkent, line 161: Double check this code, since I don't think this ever evaluates as true.

nimare/generate.py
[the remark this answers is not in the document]
jdkent: removed `foci_coords` and handle foci generation through the parameter `foci`.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+from .meta.utils import compute_ma, get_ale_kernel
+from .transforms import vox2mm, mm2vox
+
+DEF_TINY = np.finfo(float).eps
+
+
+def create_coordinate_dataset(
+    n_foci,
+    fwhm,
+    sample_size,
```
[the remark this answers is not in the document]
jdkent: Thanks! followed your top three bullet points for this pull request.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+        or a list specifying the sample size for each
+        study.
+    n_studies : :obj:`int`
+        Number of n_studies to generate.
+    foci_coords : :obj:`list` of three item array_like objects, optional
+        Ground truth xyz coordinates of foci.
+        If not specified, coordinates will be randomly selected.
+        If not the same length as the number of foci (i.e., `n_foci`),
+        foci will either be over-sampled or under-sampled.
+    n_noise_foci : :obj:`int` or :obj:`list`, optional
```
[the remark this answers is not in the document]
jdkent: went with `foci`.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+        fwhm,
+        n_studies,
+        foci_coords,
+        n_noise_foci,
+        foci_weights,
+        seed=seed,
+        space="MNI",
+    )
+
+    source_dict = create_source(foci_dict, sample_sizes)
```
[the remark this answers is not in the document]
jdkent: good catch, thanks, should be fixed.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+            os.path.join(get_resource_path(), "templates", "MNI152_2x2x2_brainmask.nii.gz")
+        )
+    else:
+        raise NotImplementedError("Only coordinates for the MNI atlas has been defined")
+
+    # use a template to find all "valid" coordinates
+    template_data = template_img.get_fdata()
+    possible_ijks = np.argwhere(template_data)
+    # number of "convergent" foci each study should report
+    if isinstance(n_foci, int):
```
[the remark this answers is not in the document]
jdkent: reorganized code so that type checking occurs first in the user facing function `create_coordinate_dataset`.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+        n_focis = [n_foci] * n_studies
```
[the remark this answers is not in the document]
jdkent: simplified the variable usage here/refactored to use `foci`, so this particular comment should have been addressed.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+    # foci_coords were specified by the caller
+    elif isinstance(foci_coords, list):
+        ground_truth_foci_ijk = [mm2vox(coord, template_img.affine) for coord in foci_coords]
+        foci_coords_num = len(ground_truth_foci_ijk)
+    else:
+        raise ValueError("foci_coords must be a list or None")
+
+    if not foci_weights:
+        foci_weights = [1] * foci_coords_num
+    if not isinstance(foci_weights, list):
```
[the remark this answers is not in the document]
jdkent: Thanks! I deleted this code, but I will keep that in mind for the future.

nimare/generate.py
```diff
@@ -0,0 +1,290 @@
...
+    seed=42,
+    space="MNI",
+):
+    """Generate study specific foci.
+
+    Parameters
+    ----------
+    n_foci : :obj:`int` or :obj:`list`
+        The number of foci to be generated per study.
+    fwhm : :obj:`float` or :obj:`list`
```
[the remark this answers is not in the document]
jdkent: stuck with an `int`.

nimare/generate.py
jdkent: this is refactored, and I do not think the denominator `np.sum(weighted_prob_map)` would ever be zero (it was for the case when the user specified `foci=0`), but I handle `foci=0` differently now.

nimare/generate.py
jdkent: from our conversation today (10/28/2020), I have this changed.

nimare/tests/test_estimator_performance.py
jdkent: I just wrote `1 - ALPHA` instead to side step interpretation, but I can rename to TRUE_NEG.

nimare/generate.py
jdkent: good idea, done!

nimare/tests/test_estimator_performance.py
jdkent: I've injected noise foci into the test, and it looks like reasonable output with the ALE/ALEKernel, but the ALE/MKDAKernel performs poorly (one interesting note about the poor performance is that ALE/MKDAKernel appears to perform worse with the addition of more noise foci).

nimare/tests/test_estimator_performance.py
jdkent: done! 2 mm threshold for signal and 14 mm threshold for non-signal (noise)

jdkent:
summary of findings:
- errors
  - meta.fit()
    - any Meta + Peaks2MapsKernel does not work
    - ALE + KDAKernel sometimes does not work
  - corr.transform(res)
    - MKDA + any kernel + any corrector except montecarlo simulations
    - KDA + any kernel + any corrector except montecarlo simulations
    - ALE + KDAKernel + montecarlo sometimes does not work
- poor performance
  - ALE + MKDAKernel + any corrector demonstrates inadequate performance

jdkent:
once #362 and #365 are merged, I believe all tests will pass, and we can have another round of reviews since I changed the code somewhat significantly after Tal's review.

jdkent:
Now that tests are passing, can I get another look through @tsalo and @tyarkoni?
If it looks good to you both then I think we can merge.

tyarkoni:
Sorry, @tsalo—I only saw you self-requested a review after I hit merge. Feel free to revert if you like, or we can just apply any new changes as a new PR.

PR #358, opened 2020-10-24
[MAINT] Add Dorota Jarecka to Zenodo file

djarecka:
@tsalo - not sure if you want to keep any specific order, so just added my name before Angie


tsalo:
That's perfect, thank you! I will probably reorganize the Zenodo file a bit, but I'll handle it.

Issue #350, opened 2020-10-05
KDA does not work with FWECorrector(method='montecarlo')


jdkent:
error message:
```
TypeError: correct_fwe_montecarlo() got an unexpected keyword argument 'voxel_thresh'
```
similar to issues #349 and #348

tsalo:
I think the problem is just that the method doesn't have a `voxel_thresh` argument. There's a passing test that uses the Monte Carlo FWE correction for `KDA`:

https://github.com/neurostuff/NiMARE/blob/58e6e82e3ca15ae89390bbf4c294b44f689bd5a0/nimare/tests/test_meta_mkda.py#L94-L95

That said, it's probably worth it to double-check the original KDA paper(s) and see if we can just pass the KDA kernel to an `MKDADensity` Estimator, since the core method might be equivalent. If it _is_ different, I'd like to document how and why that is the case, so we can definitively say that the interface (and parameters) should be different as well.

tsalo:
While I still need to check the original papers, in the mean time, do you want to just update the tests and examples to drop the `voxel_thresh` argument so they pass?

I also accidentally added a `bug` label. I don't think this is as much a bug as an idiosyncratic interface that we _may_ want to improve, so I'm going to switch from `bug` to `enhancement`.

jdkent:
got it, should be removing the `voxel_thresh` argument [when necessary](https://github.com/neurostuff/NiMARE/pull/343/files#diff-c15fffeac20344dca58c5a683a4fa8a663ff740d501067247d1329df9dcccb2fR403-R423)

PR #362, opened 2020-10-29
[FIX] Treat vfwe as an array of floats for KDA

### LINKED ISSUE #361: KDA + KDAKernel + FWE(method=montecarlo) logp_level-voxel map is of type int
The KDA estimator outputs logp_level-voxel maps as integer type, but the map should be float, I think 
changing this line:
https://github.com/neurostuff/NiMARE/blob/58e6e82e3ca15ae89390bbf4c294b44f689bd5a0/nimare/meta/mkda.py#L664
to 
`vfwe_map = of_values.copy().astype(float)`
fixes the issue.

jdkent:
Closes #361
ref #350, #343

- change vfwe array to float

I'll link to the relevant failed test in #343 when it's available.

### DIFF
--- a/nimare/meta/mkda.py
+++ b/nimare/meta/mkda.py
@@ -661,7 +661,7 @@ class KDA(CBMAEstimator):
         perm_max_values = perm_results
 
         # Voxel-level FWE
-        vfwe_map = of_values.copy()
+        vfwe_map = of_values.copy().astype(float)
         for i_vox, val in enumerate(of_values):
             vfwe_map[i_vox] = -np.log10(null_to_p(val, perm_max_values, "upper"))
         vfwe_map[np.isinf(vfwe_map)] = -np.log10(np.finfo(float).eps)
--- a/nimare/tests/test_meta_mkda.py
+++ b/nimare/tests/test_meta_mkda.py
@@ -1,6 +1,8 @@
 """
 Test nimare.meta.mkda (KDA-based meta-analytic algorithms).
 """
+import numpy as np
+
 import nimare
 from nimare.correct import FDRCorrector, FWECorrector
 from nimare.meta import kernel, mkda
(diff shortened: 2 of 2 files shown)


nimare/meta/mkda.py
```diff
@@ -661,7 +661,7 @@ def correct_fwe_montecarlo(self, result, n_iters=10000, n_cores=-1):
         perm_max_values = perm_results
 
         # Voxel-level FWE
-        vfwe_map = of_values.copy()
+        vfwe_map = of_values.copy().astype(float)
```
tsalo, line 664: `numpy.ndarray.astype()` returns a copy by default, so you can drop the old `copy()`.

jdkent, line 664:
```suggestion
        vfwe_map = of_values.astype(float)
```

nimare/meta/mkda.py
```diff
@@ -661,7 +661,7 @@ def correct_fwe_montecarlo(self, result, n_iters=10000, n_cores=-1):
         perm_max_values = perm_results
 
         # Voxel-level FWE
-        vfwe_map = of_values.copy()
+        vfwe_map = of_values.astype(float)
```
tsalo, line 664: Ugh, I'm sorry about this. I should have paid closer attention to the actual context here. I think it's a bad idea to use `of_values` here in the first place, since it gives the impression that some OF values will be retained in `vfwe_map` after the for loop below (or even that the `of_values` array is what's being operated on).

Instead, I think we should use something like `np.empty(of_values.shape, dtype=float)`. `float` is the default datatype for `numpy.empty()`, but in this case it might help to be explicit.

jdkent, line 664:
```suggestion
        vfwe_map = np.empty(of_values.shape, dtype=float)
```

tsalo:
The change looks good to me, but do you want to add a test or something to cover this situation? I think a simple data type check for the relevant maps would be good. Probably for all of the CBMA estimators though.

jdkent:
added one test for the KDA cbma estimator, and #343 (and extensions from that pull request) will catch these errors, at least, that pull request is how I found this issue.
(p.s., can this pull request also count for hacktoberfest? 🥺)

tsalo:
Merge commits mess up the codecov step in CI, but this PR looks good so I'm going to merge.

PR #365, opened 2020-10-29
[FIX] enforce correct lowest p-value

### LINKED ISSUE #364: Sometimes p-values are not clipped correctly with montecarlo simulations
In the function `null_to_p`, sometimes this line:
https://github.com/neurostuff/NiMARE/blob/db6347569309d580fcd51e044da764bb9b1286ea/nimare/stats.py#L144

can evaluate to a percentile that is less than the lower bound of p-values, (e.g., if the null array is 100 items making the minimum p-value 0.01 and if there are repeat values in the null_array like (1, 1, 1, 2, ..., 5, 5, 6), the percentile can be under 0.01 without being zero, which becomes a problem in this line:
https://github.com/neurostuff/NiMARE/blob/db6347569309d580fcd51e044da764bb9b1286ea/nimare/stats.py#L151-L152

resulting in some instances where the returned p-value is outside the expected range (0.01 - 0.99).

I'll submit a fix.

jdkent:
Closes #364 

- p-values are clipped in null_to_p such that they have to be in the range [smallest_value, (1 - smallest_value)]
- added test to check behavior.

[the solution came from stackoverflow](https://stackoverflow.com/questions/4092528/how-to-clamp-an-integer-to-some-range)

### DIFF
--- a/nimare/stats.py
+++ b/nimare/stats.py
@@ -148,7 +148,7 @@ def null_to_p(test_value, null_array, tail="two"):
         raise ValueError('Argument "tail" must be one of ["two", "upper", ' '"lower"]')
 
     smallest_value = np.maximum(np.finfo(float).eps, 1.0 / len(null_array))
-    if p_value == 0:
+    if p_value < smallest_value:
         p_value = smallest_value
     elif p_value == 1:
         p_value = 1.0 - smallest_value
--- a/nimare/stats.py
+++ b/nimare/stats.py
@@ -149,6 +149,8 @@ def null_to_p(test_value, null_array, tail="two"):
 
     smallest_value = np.maximum(np.finfo(float).eps, 1.0 / len(null_array))
 
+    # ensure p_value in the following range:
+    # smallest_value <= p_value <= (1.0 - smallest_value)
     p_value = max(smallest_value, min(p_value, 1.0 - smallest_value))
 
     return p_value
--- a/nimare/stats.py
+++ b/nimare/stats.py
@@ -148,10 +148,8 @@ def null_to_p(test_value, null_array, tail="two"):
         raise ValueError('Argument "tail" must be one of ["two", "upper", ' '"lower"]')
 
     smallest_value = np.maximum(np.finfo(float).eps, 1.0 / len(null_array))
-    if p_value < smallest_value:
-        p_value = smallest_value
-    elif p_value == 1:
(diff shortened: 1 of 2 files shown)


nimare/tests/test_stats.py
```diff
@@ -25,3 +25,7 @@ def test_null_to_p():
     assert math.isclose(stats.null_to_p(101, data, "lower"), 0.99)
     assert math.isclose(stats.null_to_p(101, data, "upper"), 0.01)
     assert math.isclose(stats.null_to_p(101, data, "two"), 0.01)
+
+    # modify data to handle edge case
+    data[98] = 100
+    assert math.isclose(stats.null_to_p(100, data, "upper"), 0.01)
```
tsalo, line 31: Could you duplicate all (or at least most) of the above tests with the modified array?

nimare/stats.py
```diff
@@ -148,10 +148,8 @@ def null_to_p(test_value, null_array, tail="two"):
         raise ValueError('Argument "tail" must be one of ["two", "upper", ' '"lower"]')
 
     smallest_value = np.maximum(np.finfo(float).eps, 1.0 / len(null_array))
-    if p_value == 0:
-        p_value = smallest_value
-    elif p_value == 1:
-        p_value = 1.0 - smallest_value
+
+    p_value = max(smallest_value, min(p_value, 1.0 - smallest_value))
```
tsalo, line 152: This is a very dense line. I don't want you to write less efficient code to improve readability, so would you mind adding a comment above it?

tsalo (changes requested):
Minor requests.

tsalo (approved):
LGTM. Feel free to merge when you're ready.

nimare/tests/test_stats.py
jdkent: duplicated relevant edge cases.

nimare/stats.py
jdkent: done!

jdkent:
Thanks! I don't believe I have write access to this repository so I cannot merge this pull request.

tsalo:
Oh well let's rectify _that_ right now. I'm inviting you to the neurostuff organization and once you're on that I'll give you write access to this repo.

jdkent:
Just accepted membership!

PR #368, opened 2020-11-04
[TST] Do not download test peaks2maps to tmpdir

### LINKED ISSUE #367: Peaks2Maps downloads model every time tests are run
Haven't looked into the cause, but despite the "this is a one-off operation" message, the full (large) Peaks2Maps model seems to download every time I run the tests. Not sure if this is by design, but I think it's reasonable for the tests to keep the model around in this case; it adds a minute or so to the test run time.

tsalo:
Closes #367.

- Allow Peaks2Maps test to download the model to the NiMARE resources folder instead of a temporary directory.
    - This will keep the test from downloading the model every time it's run locally.


tsalo:
My change has unearthed a bug in our data directory determination function, so I will need to fix that as well.

tsalo:
Closed in favor of #419.

PR #369, opened 2020-11-04
[REF] Refactor KernelTransformer hierarchy

tyarkoni:
The `KernelTransformer` hierarchy had a lot of redundancy, so I refactored it. Mostly I've pushed a lot of code into the base class's `transform()` method, which now internally calls an `_apply_kernel()` class method that needs to be implemented in subclasses. I stopped short of making this an abstract method, as the `Peaks2MapKernel` looks like it does something a little different, and I haven't taken the time yet to shoehorn it into the structure of the other kernel classes. But I don't think there's really a good use case for that kernel at the moment, so I'm not very concerned about it (TBH I wonder if we should just remove it entirely—it would be kind of disastrous if people started to use it in meta-analyses).

Tests seem to pass, but I haven't added any new ones, as there's no new functionality.

Note that I moved the base `KernelTransformer` into the `meta.kernel` module. It's now quite long, and I think if we keep all of the parent classes in a single `base` module, it's going to get unreadable very quickly. Plus I find it much easier to work with subclasses when the parent is in the same module. A reasonable compromise if you have an aversion to the current structure would be to move `kernel.py` into its own folder, with separate `base` and `kernel` modules. But to me this seems fine as-is.

### DIFF
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -330,6 +330,7 @@ class CBMAEstimator(MetaEstimator):
         }
 
         # Allow both instances and classes for the kernel transformer input.
+        from .meta.kernel import KernelTransformer
         if not issubclass(type(kernel_transformer), KernelTransformer) and not issubclass(
             kernel_transformer, KernelTransformer
         ):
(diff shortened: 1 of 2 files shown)


tsalo (commented):
I think the proposed changes are great. I only wonder if `_apply_kernel()` should be named `_transform()` (to match our `fit()`/`_fit()` convention in the Estimators)?

tsalo:
> A reasonable compromise if you have an aversion to the current structure would be to move kernel.py into its own folder, with separate base and kernel modules. But to me this seems fine as-is.

I'm happy to have base classes spread out. The only issue is that spreading things out can introduce some circular imports that will break everything for _some_ of the classes. I don't remember the specific cases where that happens, but I do recall struggling with it.

tyarkoni:
> I think the proposed changes are great. I only wonder if _apply_kernel() should be named _transform() (to match our fit()/_fit() convention in the Estimators)?

Makes sense to me.

> I'm happy to have base classes spread out. The only issue is that spreading things out can introduce some circular imports that will break everything for some of the classes. I don't remember the specific cases where that happens, but I do recall struggling with it.

Let's try it out and see how it goes. If we end up having to import something in a local scope here and there to avoid circularity, that seems fine to me. I'll give it a whirl later if I have time.

tyarkoni:
I think this is ready to merge, unless you feel strongly about black's 88-character line length, in which case I can fix the style issues.

tyarkoni:
Actually, I'm going to go ahead and merge this so I can work on some other things without having to resolve merge conflicts later. We can always fix styling separately.

PR #372, opened 2020-11-04
[ENH] Add empirical null method to density-based CBMA Estimators

### LINKED ISSUE #348: MKDADensity does not work with FDRCorrector or FWECorrector(method='bonferroni')
When trying to correct for multiple comparisons using FDR or FWE (using bonferroni, the montecarlo method works fine), the following error arises.

```
ValueError: <class 'nimare.correct.FDRCorrector'> requires "p" maps to be present in the MetaResult, but none were found.
```

psuedo-code
```
from nimare.correct import FDRCorrector
from nimare.meta import mkda

meta = mkda.MKDADensity()
corr = FDRCorrector()
res = meta.fit(dset)
cres = corr.transform(res)  # this is where the error occurs
```

### LINKED ISSUE #349: KDA does not work with FDRCorrector or FWECorrector(method='bonferroni')
same error as #348

### LINKED ISSUE #371: Support resampling-based null in all density-based CBMA estimators
Here is the basic approach for the resampling-based uncorrected null:

1. Generate MA arrays for all studies in dataset.
1. Randomly select one voxel from each study.
1. Calculate summary value (e.g., ALE or OF value) from MA values at random voxels.
1. Retain summary value in null distribution across permutations.
1. Use null distribution to convert actual summary value array (e.g., ALE map) to p-values.

tsalo:
Closes #371, closes #348, closes #349.

TODO:
- Validate method. I'm thinking a close code review and an example should be sufficient.

- Add an empirical null-based approach to the ALE, MKDADensity, and KDA Estimators.
- Add tests for empirical null method.
- Simplify ALE calculation (I'm kicking myself over this).

### DIFF
--- a/nimare/meta/ale.py
+++ b/nimare/meta/ale.py
@@ -46,6 +46,10 @@ class ALE(CBMAEstimator):
         ALEKernel.
     null : {"analytic", "empirical"}, optional
         Method by which to determine uncorrected p-values.
+    n_iters : int, optional
+        Number of iterations to use to define the null distribution.
+        This is only used if ``null=="empirical"``.
+        Default is 10000.
     **kwargs
         Keyword arguments. Arguments for the kernel_transformer can be assigned
         here, with the prefix '\kernel__' in the variable name.
--- a/nimare/meta/ale.py
+++ b/nimare/meta/ale.py
@@ -127,9 +127,10 @@ class ALE(CBMAEstimator):
 
         Parameters
         ----------
-        data : array, pandas.DataFrame, or list of img_like
+        data : (S [x V]) array, pandas.DataFrame, or list of img_like
             Data from which to estimate ALE scores.
-            The data can be (1) a contrast-by-voxel array of MA values,
+            The data can be:
+            (1) a 1d contrast-len or 2d contrast-by-voxel array of MA values,
             (2) a DataFrame containing coordinates to produce MA values,
             or (3) a list of imgs containing MA values.
 
@@ -145,6 +146,7 @@ class ALE(CBMAEstimator):
         elif isinstance(data, list):
             ma_values = self.masker.transform(data)
         elif isinstance(data, np.ndarray):
+            assert data.ndim in (1, 2)
             ma_values = data.copy()
         else:
             raise ValueError('Unsupported data type "{}"'.format(type(data)))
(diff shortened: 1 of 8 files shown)


tsalo:
@jdkent, I think I've dealt with the expected errors related to #348, #349, and #350 in `test_estimator_performance.py`, but there are still four failed tests at the stage where it compares p-values. I'm not really sure why the p-values aren't within the expected range. Would you mind taking a look?

jdkent:
I'll take a look this afternoon!

nimare/meta/mkda.py
```diff
@@ -90,24 +93,144 @@ def _fit(self, dataset):
...
+        n_studies, n_voxels = ma_maps.shape
+        weight_vec = np.squeeze(self.weight_vec_)  # cannot have singleton
+        assert weight_vec.ndim == 1
+        null_distribution = np.zeros(n_iters)
+        for i_iter in range(n_iters):
+            # One random MA value per study
+            null_ijk = np.random.choice(np.arange(n_voxels), n_studies)
+            iter_ma_values = ma_maps[np.arange(n_studies), null_ijk]
+            # Calculate summary statistic
+            iter_ma_values *= weight_vec
```
jdkent, line 169: Is this re-weighing of the `iter_ma_values` necessary? I see the ma_values that are passed into this function are already weighted, correct? https://github.com/neurostuff/NiMARE/pull/372/files#diff-09e94548290cad221e601257340cd9acfc7d1e851de54756374671589b3a6d74R98.

If that's the case I think this is (at least part of) the reason for the poor performance of MKDADensity with ALEKernel.

nimare/meta/ale.py
```diff
@@ -44,6 +44,12 @@ class ALE(CBMAEstimator):
     kernel_transformer : :obj:`nimare.base.KernelTransformer`, optional
         Kernel with which to convolve coordinates from dataset. Default is
         ALEKernel.
+    null : {"analytic", "empirical"}, optional
```
tyarkoni, line 47: Maybe something more explicit, like `null_method`.

nimare/meta/mkda.py
tsalo: I think you're right. The `ma_values` are already weighted. I'll fix that. Thanks!

tyarkoni (commented):
Looks good. Only one completely trivial suggestion. I had other thoughts, but I'll defer them for now, because they're guaranteed to come up when we think about abstracting out the common logic in the ALE and MKDA classes.

nimare/meta/mkda.py
jdkent: That fixed the failed tests on my end, so hopefully those tests will now pass on CI.

nimare/meta/ale.py
tsalo: +1

PR #374, opened 2020-11-05
[WIP] Minor ALE performance improvements

tyarkoni:
Some very minor performance tweaks to `ALE`. These have negligible impact on performance with existing data so far as I can tell (though I didn't profile systematically), but could in theory make a difference in certain cases (e.g., for high-res imaging data with many more voxels).

Summary of changes:
* Updating of the histograms within study is now vectorized, rather than looping over voxels.
* A few non-performance changes for clarity—mostly just assigning key variables on their own line.

Am still looking over a few other things, so I've marked this WIP for now.

### DIFF
--- a/nimare/meta/ale.py
+++ b/nimare/meta/ale.py
@@ -169,36 +169,34 @@ class ALE(CBMAEstimator):
                 reduced_ma_values, bins=hist_bins, density=False
             )[0]
 
-        # Inverse of step size in histBins
-        step = 1 / step_size
+        inv_step_size = int(np.round(1 / step_size))
 
-        # Null distribution to convert ALE to p-values.
-        ale_hist = ma_hists[0, :]
+        # Normalize MA histograms to get probabilities
+        ma_hists /= ma_hists.sum(1)[:, None]
+
+        # Histogram of integrated values can have larger MA values than any
+        # individual contrast, so use full [0, 1] range here.
+        ale_hist = np.zeros((inv_step_size,))
+        ale_hist[np.floor(hist_bins * inv_step_size).astype(int)] = ma_hists[0, :]
+        all_bins = np.round(np.arange(0, 1, step_size), 4)
+
         for i_exp in range(1, ma_hists.shape[0]):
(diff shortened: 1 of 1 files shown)


nimare/meta/ale.py
```diff
@@ -153,8 +151,8 @@ def _compute_summarystat(self, data):
         else:
             raise ValueError('Unsupported data type "{}"'.format(type(data)))
 
-        stat_values = 1.0 - np.prod(1.0 - ma_values, axis=0)
-        return stat_values
+        stat_values = 1. - np.prod(1. - ma_values, axis=0)
+        return ale_values
```
tsalo, line 155:
```suggestion
        return stat_values
```

nimare/meta/ale.py
```diff
@@ -204,59 +202,48 @@ def _compute_null_analytic(self, ma_maps):
...
+
+            # Compute output MA values, ale_hist indices, and probabilities
+            ale_scores = 1 - np.outer(1 - hist_bins[exp_idx], 1 - hist_bins[ale_idx]).ravel()
+            score_idx = np.floor(ale_scores * inv_step_size).astype(int)
+            probabilities = np.outer(exp_hist[exp_idx], ale_hist[ale_idx]).ravel()
+
+            # Reset histogram and set probabilities. Use at() because there can
+            # be redundant values in score_idx.
+            ale_hist = np.zeros((inv_step_size,))
+            np.add.at(ale_hist, score_idx, probabilities)
```
tsalo, line 241: Everything else looks good to me, so I assume that this is the section that is leading to the failed tests. I will try to dig into the new code, but for now I don't actually understand what's going on.

tyarkoni: I think the issue has to do with inconsistent rounding/casting—for the MKDA kernel, we get `score_idx` values in the top bin (i.e., 1), and they're being rounded above that. I don't think it's worth chasing this down right now, as it may go away by itself when I refactor this.

tyarkoni:
Closing in favor of #386.

PR #375, opened 2020-11-05
[ENH] Add analytic null method to MKDADensity

### LINKED ISSUE #370: Implement analytical null in KDA and MKDA estimators
ALE's analytical null _should_ be applicable to other density-based methods. The basic approach is this:

1. Define histogram ranging from zero to maximum possible summary value (e.g., ALE or OF) for dataset.
    - This should be easy for MKDA, but could be very difficult for KDA. Also, we'll need to incorporate the weights that KDA and MKDA can support.
1. Calculate MA array for dataset.
1. Calculate summary value array for dataset.
1. Bin observed MA values in study-specific histograms.
    - (Zeros are counted separately due to Python's histogram code being terrible).
1. Looping through the studies, build a summary value histogram by convolving each study's MA histogram with the summary value histogram.
    - If that's not clear, just remember that the summary value histogram is set at first to the first study's MA histogram.
1. Convert summary value histogram into null distribution by normalizing the histogram and then calculating the cumulative sum.

tsalo:
Closes #370. This must be merged _after_ #372. This *does not* implement the null method for KDA, since it's very difficult to determine the maximum possible OF value for KDA meta-analyses when preallocating the histogram bins.

- Add `_compute_null_analytic()` method to MKDADensity.
- Add test for MKDADensity with the analytic null.
- Add `_compute_summarystat()` method to MKDADensity. This is duplicated in #372.
- Add `_compute_null_empirical()` method to MKDADensity. This was a commit mistake, and this method should be reviewed in #372.
- Add `_summarystat_to_p()` method to MKDADensity. Also a duplicate of #372.

### DIFF
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -331,6 +331,7 @@ class CBMAEstimator(MetaEstimator):
 
         # Allow both instances and classes for the kernel transformer input.
         from .meta.kernel import KernelTransformer
+
         if not issubclass(type(kernel_transformer), KernelTransformer) and not issubclass(
             kernel_transformer, KernelTransformer
         ):
--- a/nimare/meta/kernel.py
+++ b/nimare/meta/kernel.py
@@ -161,7 +161,7 @@ class KernelTransformer(Transformer):
                     "Output directory does not exist. Set the path to an "
                     "existing folder with Dataset.update_path()."
                 )
-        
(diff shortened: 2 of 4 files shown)


tyarkoni (approved):
LGTM!

Issue #373, opened 2020-11-04
Move CBMA summary statistic calculation from Estimator to KernelTransformer


tsalo:
Per today's call, building a generic CBMADensityEstimator will require access to the summary statistic calculation step based on the kernel, rather than the estimator. As such, we should move that function (let's call it `_compute_summarystat()` for now) from the Estimator classes to the KernelTransformer classes.

tsalo:
This will have to happen _after_ #377 is handled.

tyarkoni:
Given that we're keeping the separate `ALE` and `MKDADensity` classes, and some of the class methods will have very different implementations (e.g., `_compute_null_analytic`), there's probably no longer a good reason to move this to the kernel, so I vote we close it.

tyarkoni:
@tsalo just tagging you here to make sure you see my edit to the previous comment. The residual point being that maybe we should abandon the "freely mix and match kernels and estimators" idea and restrict to only certain legal combinations inside each estimator.

PR #378, opened 2020-11-06
[ENH] Add analytical null approach to KDA estimator

### LINKED ISSUE #377: Implement analytical null in KDA estimator
See #370 for the basic procedure. The KDA estimator doesn't currently have the analytical null implemented, due to complications that are unique to the KDA kernel.

Specifically, the analytical null method predefines a histogram based on the _maximum possible value_ for the summary statistic (e.g., ALE or OF score). In the case of the other kernels (ALE and MKDA), the maximum possible value for each study is easily inferred because it's the maximum observed value (i.e., MA values for these kernels have a predefined maximum value regardless of convergence across foci). However, KDA sums MA values across foci within each study, so the maximum possible value ends up being the number of foci in the study multiplied by the weighting factor used in the model. The weighting factor can be (fairly) easily inferred from the modeled activation maps, but the number of foci in the study cannot.

To infer the study-wise vector of weighting factors, just take the minimum non-zero value in each study's MA map.

EDIT: To clarify why this maximum value is important: I believe that the histogram bins above the maximum observed value really come into play in the Monte Carlo FWE correction, where summary statistic values from random permutations are evaluated against the null distribution. If random summary statistics appear with higher values than were preallocated for in the initial null distribution, then there will be an error.

tsalo:
Closes #377.

- Add an analytical null method to the KDA estimator.
- Determine number of foci per study in `KDA._fit()` and then pass that to `KDA._compute_null_analytic()`. This is quite different from other analytic null methods (i.e., ALE and MKDA).


tsalo:
I think two possible workarounds for determining the number of foci per study within the analytic null methods are:

1. Directly extract the info from `Estimator.inputs_["coordinates"]` within `Estimator._compute_null_analytic()`. The problem with doing this is that `KDA. _compute_null_analytic()` is different from the `MKDA` and `ALE` versions, which means that we'll probably want to move at least part of it to the `KDAKernel` (see #373).
2. Extract the info from `Estimator.inputs_["coordinates"]` in `Estimator._fit()` and feed it into `Estimator._compute_null_analytic()` as an argument. This has the same drawbacks as the above, but doesn't require direct access to `Estimator` attributes within the `_compute_null_analytic()` method.

tyarkoni:
Isn't the solution here identical to the one for ALE? It seems like we can just set up discrete bins b_0 ... b_k, where k is the maximum number of overlapping foci anywhere in the dataset. Then we compute the normalized histogram of b's for every study (effectively giving a multinomial distribution), and then convolve them, just like with ALE. Is there something about KDA that makes that unworkable, or is the idea to try to come up with something more efficient?

tsalo:
It's not unworkable. We just need access to additional information beyond what's in the MA maps in order to define the bins (namely the number of foci per study). The workaround in this PR gets that information from `self.inputs_["coordinates"]`.

tsalo:
It seems I have to think more carefully about the histogram bins being used, but once I figure that out this should be good.

nimare/meta/mkda.py
```diff
@@ -832,6 +834,97 @@ def _compute_null_empirical(self, ma_maps, n_iters=10000):
...
+        for i_study in range(n_studies):
+            temp_ma_values = ma_values[i_study, :]
+            min_ma_values[i_study] = np.min(temp_ma_values[temp_ma_values != 0])
+
+        ma_scalar = np.mean(min_ma_values)
+        max_ma_values = min_ma_values * n_foci_per_study
+        max_poss_value = self._compute_summarystat(max_ma_values)
+        # Weighting is not supported yet, so I'm going to build my bins around the min MA value.
+        self.null_distributions_["histogram_bins"] = np.arange(
+            -ma_scalar / 2, max_poss_value + (ma_scalar * 1.5) + 0.001, ma_scalar
```
tsalo, line 872: Assuming a weight of 1 across the board, the histogram would look something like `[-0.5, 0.5, 1.5]`, with zeros going into the -0.5 to 0.5 bin, ones going into the 0.5 to 1.5 bin, etc.

tyarkoni: I don't think you need to use histograms here, since the values are always ints. You could use `np.unique()` with `return_counts=True`, which is probably also faster.

tsalo: The values may not be ints, depending on the `KDAKernel`'s `value` parameter and any weighting we might apply in the future (as in `MKDADensity`). Still, I don't think that impacts your point.

I'm not sure how to do what you're talking about, though. Would it be compatible with FWE if it's limited to _observed_ values, though? I know you mentioned dumping summary statistic values above the observed values into a final bin, but that would impact the way we implement the Monte Carlo procedure, right?

tyarkoni: Let me first say that I still think this is a rabbit hole that's not worth going down—I doubt anyone's going to use KDA, and if they do, they'll probably be doing themselves and/or the scientific community a disservice. There are almost no meta-analysis research questions I can think of where a fixed-effect estimate would be a good idea, so providing this as an easy-to-use tool seems likely to cause more harm than good.

With that said, what's the problem with having a larger number of bins than you might think you need? Even if the vast majority of values are ints, most of the probabilities will just get a probability of 0 and be ignored. it doesn't actually matter what bin the values end up in, AFAICS; remapping the observed values to p-values should still proceed just fine, no?

nimare/meta/mkda.py
```diff
@@ -725,6 +724,12 @@ class KDA(CBMAEstimator):
 
     Available correction methods: :func:`KDA.correct_fwe_montecarlo`
 
+    Warnings
+    --------
+    The KDA algorithm has been replaced in the literature with the MKDA algorithm.
+    As such, this estimator should almost never be used, outside of systematic
+    comparisons between algorithms.
```
tsalo, line 731: This may help dissuade people from using the KDA estimator.

tyarkoni: Well, we can hope. :)

nimare/meta/mkda.py
tsalo: > I doubt anyone's going to use KDA, and if they do, they'll probably be doing themselves and/or the scientific community a disservice.

I've added a warning to the docstring about this, and can add a `LGR.warning` as well.

> With that said, what's the problem with having a larger number of bins than you might think you need? Even if the vast majority of values are ints, most of the probabilities will just get a probability of 0 and be ignored. it doesn't actually matter what bin the values end up in, AFAICS; remapping the observed values to p-values should still proceed just fine, no?

I'm not sure, but shifting the bins seems to have solved the problem, so I'm happy to leave it as is.

nimare/meta/mkda.py
tsalo: To be honest, I doubt anyone will use anything except ALE (and maybe MKDAChi2, out of familiarity with Neurosynth), given that Salimi-Khorshidi paper. Unless we get SDM working, of course.

tyarkoni: I'm not sure about that; there are clear principled trade-offs between ALE and MKDA. The former is too liberal, the latter is too conservative; the ALE kernel is more biologically plausible, but requires stronger assumptions and produces less interpretable values. With KDA, it seems harder to argue that any gains justify the fundamentally bad fixed effect assumption.

If you mean that people are just going to use whichever method they think gives "better" results (in the "more likely to produce significance" sense), there's probably something to that, but that's potentially also something we could mitigate to some degree with good best practices documentation.

It does occur to me that the flaws in KDA could in principle be addressed without going full MKDA—e.g., by allowing non-binary values at each voxel, but weighting the overall map in proportion to the MKDA CIM (i.e., overlapping peaks would produce values > 1, but the overall mean across all values with non-zero activation would be 1). This would in a sense be something intermediate between MKDA and ALE, both in terms of sensitivity/specificity profile and in terms of strength of assumptions. Once we have a good performance test suite, it would be pretty trivial to implement that and test it out if we were so inclined.

tsalo: I meant more that people will just use the algorithm for which they have a clean citation. Salimi-Khorshidi found, in a specific situation, that ALE outperformed MKDA, and that eight year old conclusion is going to be the end of the investigation for most folks. At least until we publish a comparison study of our own.

tsalo:
I ran the analytic null through @jdkent's notebook and the p-values are equivalent to the empirical null ones! I'm just planning to build off of #393 once that's merged instead of trying to get this through first.

tsalo:
Closing in favor of #397.

PR #380, opened 2020-11-07
[TST] Test both analytic and empirical methods in ALE and MKDA

### LINKED ISSUE #379: Test both types of null distribution in performance tests
Now that we have both kinds of null distribution set up in (most of) our density-based CBMA algorithms, I think we should incorporate them into the performance test. @jdkent, would you be willing to take this on, since you wrote the performance tests?

jdkent:
Closes #379 

- Add tests for both null methods in ALE and MKDA estimators.
- Split performance tests into a separate job to speed up continuous integration.
- Add Makefile and pytest markers to select and run specific groups of tests.
- Fix a minor bug in the ALE estimator's analytic null distribution.
- Change random state initialization to make tests more reproducible.


tsalo:
I'm thinking that we should split `test_estimator_performance.py` off into a separate CircleCI job, since it takes so long. Just one job (i.e., not tested across Python versions). We do the same thing in `tedana` for each of our integration tests.

jdkent:
Thanks! That's a good idea, I've split off the tests that check p-values (since it's the montecarlo iterations that take up the most time), but I kept the smoke tests to be run across python versions since those tests do not take as much time (still probably about 5 minutes though), let me know if I should move the smoke tests to the separate job too.

.circleci/config.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           name: Run tests
           command:  |
             source activate py37_env
-            py.test nimare
+            py.test -m "not performance" nimare
```
tsalo, line 94: We may want a proper Makefile to manage our tests. Something like [this one](https://github.com/Brainhack-Donostia/aroma/blob/bcca026a9a270ecf11a38d32b3967ac72bf49a81/Makefile). WDYT? Do you have a preference on how to split tests?

tsalo: Perhaps `pytest.mark` is a better choice... TBH I'm not familiar with it.

jdkent: We can use both a makefile and `pytest mark`. `pytest.mark` replaces the `--skipintegration`, `-k` syntax when calling pytest, and the `if skip_integration:` logic in the integration tests, I honestly haven't used marks a lot so I do not have a strong preference, but from my little experience, marks are relatively easy to use, and I can add multiple marks to a single test function to give finer control on what tests should be run.
I'll get a makefile made.

tsalo:
I noticed a couple of odd things in the py37 test. First, it seems like the performance tests are still being run. I also saw that 100 tests were deselected, but I didn't see the normal `s` next to any tests to label them as skipped. Finally, I'm a little confused by the coverage warnings saying that the coverage files don't appear to be valid. I don't recall seeing those before, but I could be wrong.

jdkent:
I don't know about the invalid coverage files either, if you are talking about this:
`Coverage.py warning: Data file '/tmp/src/NiMARE/.coverage.0928c01398a0.236.693848' doesn't seem to be a coverage data file: Couldn't use data file '/tmp/src/NiMARE/.coverage.0928c01398a0.236.693848': no such table: coverage_schema`

[it appears that warning occurs in other builds too](https://app.circleci.com/pipelines/github/neurostuff/NiMARE/858/workflows/86b79cc2-5213-49ce-a22d-59692ece651a/jobs/3522)

tsalo:
Yes, that's the warning I'm referring to. I noticed that it isn't appearing in the CI for #383, so I think it's introduced here.

jdkent:
I see that warning in at least the final build for #383, The first time I do not see any of those warnings is in #313, but #314 has those warnings. No idea what changed.

tsalo:
Ack, sorry about that! I was looking in the wrong job for #383. I might open an issue about it at some point, but it's clearly not something you're introducing, so please just ignore my original comment.

Once the conflict is resolved and tests pass, I'll approve this PR.

nimare/meta/ale.py
```diff
@@ -227,7 +227,7 @@ def _compute_null_analytic(self, ma_maps):
 
             # Reset histogram and set probabilities. Use at() because there can
             # be redundant values in score_idx.
-            ale_hist = np.zeros((inv_step_size,))
+            ale_hist = np.zeros(ale_hist.shape)
```
tsalo, line 230: Is this a fix or a refactor? I.e., is `inv_step_size` equal to the length of `ale_hist`?

In any case, it's a clearer version, so I'd like to use it if it's correct. I just want to make sure that the summary is accurate.

jdkent: `inv_step_size` was 10,000 by default whereas ale_hist.shape was 10,0010. This only made a difference when I was using the MKDA kernel with ALE where the `ale_score` could be 10,001, which is is out of range for `inv_step_size`. But this was just so there wouldn't be an error in the code, not necessarily because this is correct, @tyarkoni, is it better to use original size of `ale_hist` or should we use `inv_step_size` instead?

tyarkoni: If this is a reflection of my recent changes, I don't think I thought very deeply about this, so if they break things, I see no problem in reverting.

tsalo:
Now that tests are passing, the only other thing I think we need in this PR is an updated summary. From what I recall, here is what you've accomplished:

1. Add tests for both null methods in ALE and MKDA estimators.
1. Split performance tests into a separate job to speed up continuous integration.
1. Add Makefile and pytest markers to select and run specific groups of tests.
1. Fix a minor bug in the ALE estimator's analytic null distribution.

Is that everything?

jdkent:
yeah, almost! one more thing I changed was how the random state was initialized for the tests for better reproducibility.
(I'll edit your above message to include that).

tsalo:
Awesome. I'm going to add those point to the Changes proposed section of the first comment, and then I think it's ready to merge. @tyarkoni any problems with merging this?

tyarkoni:
LGTM!

PR #382, opened 2020-11-07
[FIX] respect n_noise_foci value

### LINKED ISSUE #381: n_noise_foci not respected when foci is 0
`create_coordinate_dataset(foci=0, n_noise_foci=30)` only gives one coordinate per study since the [`n_noise_foci` variable is overwritten when there are no signal foci](https://github.com/neurostuff/NiMARE/blob/1e257180b57ddcc1559814340f1f666777202d4b/nimare/generate.py#L243).

jdkent:
Closes #381 

- take max of 1 or whatever n_noise_foci is.
- add test to ensure length of dataset is correct.

### DIFF
--- a/nimare/generate.py
+++ b/nimare/generate.py
@@ -240,7 +240,7 @@ def _create_foci(
     for study_signal_ijks, study in zip_longest(paired_signal_ijks, range(n_studies)):
         if study_signal_ijks is None:
             study_signal_ijks = np.array([[]])
-            n_noise_foci = 1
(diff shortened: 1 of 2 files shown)


tsalo (changes requested):
This fix seems pretty straightforward, and thanks for making sure to cover it in a test. I just have one small request for the variable names in the test.

tsalo (approved):
Thanks! The code looks good to me, so feel free to merge once tests pass.

PR #383, opened 2020-11-07
[ENH] Notebook testing performance

jdkent:
Closes #None

- tests the false positive rate for ale, kda, and mkda
- correlate p-values from ale, kda, and mkda to see which methods are more correlated with each other


tsalo:
I think the current differences are probably due to the null method used. The default for ALE and MKDA should be analytical, while the default for KDA is still the empirical null.

Regarding the example itself, can you make the notebook render in the example gallery?

jdkent:
I messed up some commits, I'll get it fixed up:
[some updates on the notebook](https://github.com/jdkent/NiMARE/blob/enh/comprehensive_test/examples/test_performance.ipynb)

observations:
- empirical and analytic null methods correlate strongly with each other (~0.99)
- the ale analytic null method results in slightly more false positives than expected.
- both mkda null methods result in a conservative false positive rate
- the ale analytic null method is not evenly distributed in p-values from [0-1] (there appears to be bias towards certain p-values).

tsalo:
It seems like the MKDA analytic and empirical results are exactly the same... which is rather impressive if there are no bugs.

jdkent:
they appear really close, ALE does better when:
```
step_size = 0.00001
hist_bins = np.round(np.arange(0, max_poss_ale + 0.001, step_size), 5)
```
the ale analytic & empirical correlation is: 0.9999446144830378

Here's the false positive rates:
image


the p-value distribution:
image

p-value correlation:
image

So increasing the granularity of the hist_bins by 10x for ALE appears to bring the analytic and empirical methods really close to each other (but this may bite into the performance gains that Tal has been working on)

tsalo:
We can profile the difference, but I don't think it'll be that big a deal. I think your newest results are a pretty compelling argument for 100000 bins.

tsalo:
Regarding the example gallery, I think that we will want to use [nbexamples](https://sphinx-nbexamples.readthedocs.io/en/latest/index.html).

Regarding merging, I think we should get #393 merged first, then an adapted version of #378, and _then_ re-run your notebook with all three estimators and the new structure.

tsalo:
Can you also measure and plot elapsed time in this notebook?

tsalo:
I spent some time trying to figure out how to use `nbexamples` and if there was a way to just render pre-run notebooks with `sphinx-gallery`, but I struck out on both counts. I don't think it's worth holding this up for that, so I think we should move forward with the KDA kernel and updating w.r.t. `master`.

tsalo:
I wonder if we should place full-on analyses, like this notebook, in a separate repository. It's useful for NiMARE users, so we should definitely link to it in the documentation, but it's not exactly an "example" and it's also very computationally intensive to run.

jdkent:
I was recently thinking the same thing, I'll create a separate repo

tsalo:
Let me know when you're ready to work on the new repo. I'm happy to contribute if you want any help.

Issue #384, opened 2020-11-07
ALE estimator with null_method=empirical and FWE=montecarlo is noticeably slower


jdkent:
Haven't taken a look at the code yet, but the combination of the ALE estimator with the empirical null_method with the fwe montecarlo corrector performs slowly, (the first couple FWE montecarlo iterations take somewhere between 10-20 seconds and the following iterations take 2-5 seconds each).

This is impacting #380 to finish tests within the default timeout period of 10 minutes with no output.

tyarkoni:
I've vectorized the empirical null generation methods for both ALE and MKDA, so that should help a bit. Am working on optimizing the monte carlo routines as well, so hopefully they'll be much faster soon.

tyarkoni:
Update: #386 oesn't seem to have helped; tests are still stalling on Circle.

tyarkoni:
Possibly related: things seem to stall (at least on my machine) with FWE=montecarlo and n_cores > 1. Seems like there might be some underlying multiprocessing issue.

tsalo:
We've done a lot to the CBMA estimators and FWE correction. I wonder if this is still applicable.

jdkent:
I don't think this is still applicable, I'll close.

PR #386, opened 2020-11-10
[PERF] Various performance improvements

tyarkoni:
Various performance improvements and tweaks. Summary of changes:
* The empirical null generation methods for ALE and MKDA are now vectorized and should be much faster.
* The analytical null generation method for MKDA is now much simpler and extremely fast. Under the assumption of voxel independence (almost certainly false, but impossible to avoid in an analytical solution), computing the sum of a bunch of binomial distributions (which is basically what each experiment map amounts to) can be done efficiently via serial convolution. This is not possible for ALE, unfortunately. There may be some analogous way to convolve multinomial distributions efficiently, but if so, it's not obvious to me what it is.
* The analytical null generation method for ALE is now vectorized and the code is cleaner, though performance seems to have improved only marginally.
* The `compute_kda_ma` function can now process an entire dataset at once, rather than only a single experiment. This allows for a considerable speed-up (on the order of 30 - 50%), as we minimize the overhead of `transform()`. (I started to implement something similar in `compute_ma`, but then realized that this isn't really possible, because the kernel in ALE depends on each experiment's sample size, so the gain in efficiency would be minimal compared to the gain for MKDA, and the logic would be more convoluted.)
* `null_to_p()` can now take an array of test values, rather than just a single value. This is only slightly more efficient, as the function that does the heavy lifting (`percentileofscore()`) isn't vectorized. But it's still something of an improvement, and will eliminate a bunch of ugly loops throughout the codebase. (I haven't updated all of the invocations of `null_to_p()` yet to pass arrays.)

Other minor fixes/improvements that aren't strictly performance-related, but I included anyway:
* Abstracted the weight computation in MKDA and cleaned up some of the logic.
* Refactored `MKDAKernel` so it's basically just a duplicate of `KDAKernel` with only `sum_overlap` changed.
* A few other small things I'm forgetting or am too lazy to summarize.


nimare/stats.py
```diff
@@ -136,24 +136,22 @@ def null_to_p(test_value, null_array, tail="two"):
     P-values are clipped based on the number of elements in the null array.
     Therefore no p-values of 0 or 1 should be produced.
     """
+    test_value = np.atleast_1d(test_value)
+    # TODO: this runs in N^2 time; is there a more efficient alternative?
+    p = np.array([stats.percentileofscore(null_array, v, "rank") for v in test_value])
```
jdkent, line 141: If you would rather it take `N^2 space`, I think I translated the rank portion of the [percentileofscore function](https://github.com/scipy/scipy/blob/v1.5.4/scipy/stats/stats.py#L1942-L2029)

```
def percentileofscore(a, scores):
    n = len(a)
    a_repeat = np.repeat(a[:, np.newaxis], scores.shape[0], axis=1)
    left = np.count_nonzero(a_repeat < scores, axis=0)
    right = np.count_nonzero(a_repeat <= scores, axis=0)
    # percentage
    return (right + left + (right > left)) * 50.0/n
```

Don't know if that would be feasible for the size of the arrays we need.

jdkent: Here is a test for this function to make sure it does the same thing as scipy:

```
from nimare import stats
from scipy.stats import percentileofscore as scipy_percentileofscore

def test_percentileofscore():
    a = np.arange(1, 101)
    scores = np.arange(0, 50)
    nimare_p = stats.percentileofscore(a, scores)
    scipy_p = np.array([scipy_percentileofscore(a, score) for score in scores])
    assert np.all(nimare_p == scipy_p)
```

tyarkoni: Good thought, but not really feasible—we're looking at ~200k voxels for typical analyses, and a 200k x 200k matrix isn't going to fit in memory on most machines (and even if it did, it would be very slow). I think a more efficient solution will probably use something like [numpy.searchsorted](https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html) to do the heavy lifting.

Feel free to open a PR (or add commits to this one) if you like—though it's not a very high priority, as `null_to_p` is usually only invoked once per analysis. So right now it seems to add a constant 2 - 3 seconds to any `fit()` call that uses MNI 2 mm space, but it doesn't get any worse for bigger analyses (including more permutations).

nimare/meta/kernel.py
```diff
@@ -262,19 +269,18 @@ def _transform(self, mask, coordinates):
...
+        exp_idx = coordinates["id"].values
+        transformed = compute_kda_ma(
+            dims, vox_dims, ijks, self.r, self.value, exp_idx, sum_overlap=self._sum_overlap
+        )
+        exp_ids = np.unique(exp_idx)
+        return transformed, exp_ids
 
 
-class KDAKernel(KernelTransformer):
+class MKDAKernel(KDAKernel):
```
jdkent, line 281: some of the tests fail becuase they check if the kernel is an instance of KDAKernel, which MKDAKernel is now, [I pushed a commit to my fork fixing this](https://github.com/neurostuff/NiMARE/compare/performance-improvements...jdkent:performance-improvements)

nimare/meta/utils.py
```diff
@@ -284,30 +284,47 @@ def compute_kda_ma(shape, vox_dims, ijks, r, value, sum_overlap=False):
         Sphere radius, in mm.
     value : :obj:`int`
         Value for sphere.
+    exp_idx : array_like
+        Optional indices of experiments. If passed, must be of same length as
+        ijks. Each unique value identiies all coordinates in ijk that come from
```
jdkent, line 289:
```suggestion
        ijks. Each unique value identifies all coordinates in ijk that come from
```
typo

nimare/meta/utils.py
```diff
@@ -284,30 +284,47 @@ def compute_kda_ma(shape, vox_dims, ijks, r, value, sum_overlap=False):
...
-        sphere = np.round(sphere.T + peak)
+    squeeze = exp_idx is None
+    if exp_idx is None:
+        exp_idx = np.ones(len(ijks))
+
+    uniq, exp_idx = np.unique(exp_idx, return_inverse=True)
+    n_studies = len(uniq)
+
+    kernel_data = np.zeros([n_studies] + list(shape), dtype=type(value))
+    n_dim = ijks.shape[1]
```
jdkent, line 311: I passed ijks as a list (not a numpy array) to this function causing this line to throw an error, [I pushed a commit to my fork fixing the tests](https://github.com/neurostuff/NiMARE/compare/performance-improvements...jdkent:performance-improvements)

nimare/meta/kernel.py
tyarkoni: Thanks!

tyarkoni:
@tsalo maybe hold off on review; the analytical null computation for MKDA seems to be broken now. It was working fine last night, but I must have changed something inadvertently. This also highlights the need to add assertions against specific values.

tsalo:
@tyarkoni do you think #378 will be useful here? Just as an additional base for the generalized analytical version?

tyarkoni:
@tsalo I fixed the bug mentioned above, so this should be ready for review.

Re: #378, not sure it will help... the beauty of MKDA is that the binary kernel makes a bunch of problems easy to solve. For KDA, the problem seems to me to be directly analogous to ALE, in that each experiment can be thought of as a multinomial distribution (though it's a somewhat easier problem than ALE, as the number of possible bins is much smaller). There's probably some more efficient way to convolve those than what I do in #374 (which isn't integrated into this PR), but I'm not sure what it is.

nimare/meta/mkda.py
```diff
@@ -305,19 +250,17 @@ def _run_fwe_permutation(self, params):
         between vFWE and cFWE.
         """
         iter_ijk, iter_df, conn, voxel_thresh = params
-        iter_ijk = np.squeeze(iter_ijk)
-        iter_df[["i", "j", "k"]] = iter_ijk
```
tsalo, line 309: I can't find where you fill in the permutation's default DataFrame with its IJK values.

nimare/meta/mkda.py
```diff
@@ -192,63 +188,15 @@ def _compute_null_analytic(self, ma_maps):
...
-        # histBins) of that value or lower.
-        null_distribution = stat_hist / np.sum(stat_hist)
-        null_distribution = np.cumsum(null_distribution[::-1])[::-1]
+        # MKDA maps are binary, so we only have k + 1 bins in the final
+        # histogram, where k is the number of studies. We can analytically
+        # compute the null distribution by convolution.
+        prop_active = ma_values.mean(1)
+        ss_hist = 1.
+        for exp_prop in prop_active:
+            ss_hist = np.convolve(ss_hist, [1 - exp_prop, exp_prop])
```
tsalo, line 197: Whoa... that's seriously awesome.

tsalo: It gives different results from the old version but there is at least one major bug in the old version ([here](https://github.com/neurostuff/NiMARE/blob/44c493a0b71a85b8aeae297a7f566bad532035c7/nimare/meta/mkda.py#L241-L243)), so I can't really consider backwards incompatibility a problem.

tsalo (changes requested):
The changes look good to me- I see one possible bug in the FWE correction code.

This also alerted me to bugs in the analytical null method for MKDA, wherein I was calculating ALE values from the MA values. I noticed that, when I tried switching to the OF value calculation (i.e., summing), I got indexing errors. I wish we had the time to figure out how to fix the current version in order to compare against the new one, but I understand that it's probably not worth the effort.

nimare/meta/mkda.py
tsalo: If they aren't inserted somewhere, then the "random" permutations would just be done on the original dataset.

tsalo: Compare the old version (first image) to the new version (second image):

image

image

nimare/meta/mkda.py
tyarkoni: It converges almost perfectly with the Monte Carlo version (see [this new test](https://github.com/neurostuff/NiMARE/blob/69f1aada8fc24d9baea3e50ea1330bdb31305d56/nimare/tests/test_meta_mkda.py#L119), which has a pretty low tolerance for difference), so I'm pretty confident that they're both right—or at least, any bug would have to be in the common code, likely in `compute_summarystat` (which is possible—I did modify it). :)

nimare/meta/mkda.py
tyarkoni: Yikes, good catch! This is pretty bizarre though, because the results look basically the same coming out of `ALE` and `MKDADensity`. Will look into it.

tyarkoni: Oh, nevermind—I was thinking of the uncorrected results. I haven't actually looked at what comes out of the corrector. Could well be totally broken.

tyarkoni:
@tsalo @jdkent the performance tests are timing out (probably related to #384), but everything else is passing. Since the plan is to split the tests up anyway, I vote we merge this as-is and deal with it separately. Any objections?

tsalo:
It looks like the linter caught a few things that I'd like to clean up before merging. Would you mind running `black` on the package?

nimare/meta/mkda.py
```diff
@@ -18,6 +18,7 @@
 from ..transforms import p_to_z
 from ..utils import round2
 from .kernel import KDAKernel, MKDAKernel
+from .utils import compute_kda_ma
```
tsalo, line 21: It looks like this is unused.

tyarkoni:
> It looks like the linter caught a few things that I'd like to clean up before merging. Would you mind running `black` on the package?

Done!

tyarkoni:
@tsalo okay to merge now? I ran black, so the remaining linting errors are presumably due to differing configurations/versions, which I don't see much point in trying to standardize on.

tsalo (approved):
Sorry for the delay, but everything looks good to me!

PR #392, opened 2020-11-11
[FIX] Change default seed to None

jdkent:
Closes None

- changes default seed in `create_coordinate_dataset` to None, so that each time it is called it will generate a different random dataset.

### DIFF
--- a/nimare/generate.py
+++ b/nimare/generate.py
@@ -16,7 +16,7 @@ def create_coordinate_dataset(
     sample_size=30,
     n_studies=30,
     n_noise_foci=0,
-    seed=42,
+    seed=None,
     space="MNI",
 ):
     """Generate coordinate based dataset for meta analysis.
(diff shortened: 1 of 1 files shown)


nimare/generate.py
```diff
@@ -44,7 +44,7 @@ def create_coordinate_dataset(
         Number of foci considered to be noise in each study. (Default=0)
     seed : :obj:`int`
```
tsalo, line 45: Can you add None as an option?

nimare/generate.py
```diff
@@ -44,7 +44,7 @@ def create_coordinate_dataset(
...
         Random state to reproducibly initialize random numbers.
```
tsalo, line 46: Perhaps add something like "If None is provided, then the random state will be randomly selected."

tsalo (approved):
LGTM

PR #393, opened 2020-11-12
[REF] CBMA re-organization and improvement

tyarkoni:
This PR refactors and standardizes the CBMA hierarchy. Changes include:
* A new `cbma` folder containing `base`, `ale`, and `mkda` modules.
* Individual estimator and kernel transformer classes are now imported into the `meta` namespace (though importing through the individual submodules will also still work, so code shouldn't break).
* The `ALE` and `MKDADensity` classes have been dramatically stripped down, with most of the common functionality moving into `CBMAEstimator`.
* The unified FWE correction procedure is no longer broken for `MKDADensity`, and should be considerably faster, as it doesn't need to compute p-values in each permutation.
* Cleaner, vectorized cluster correction routine (should also be slightly faster).
* More efficient `null_to_p`; now only processes unique values in array, and then reconstructs the original.
* New `p_to_summarystat()` classmethod in `CBMAEstimator` that looks up a critical summary stat value based on the passed p-value.

With this done, we're in a good position to refactor some of the paired test classes, which could also in theory use a lot of the same machinery. But let's keep that for a separate PR.


nimare/meta/cbma/base.py
```diff
@@ -0,0 +1,480 @@
...
+
+    def _run_fwe_permutation(self, params):
+        """
+        Run a single Monte Carlo permutation of a dataset. Does the shared work
+        between vFWE and cFWE.
+        """
+        iter_ijk, iter_df, conn, voxel_thresh = params
+
+        iter_ijk = np.squeeze(iter_ijk)
+        iter_df[["i", "j", "k"]] = iter_ijk
```
tsalo, line 293: Would it require less memory to copy `self.inputs_["coordinates"]` here instead of copying it in `correct_fwe_montecarlo` and duplicating it `n_iters` times?

nimare/meta/cbma/base.py
```diff
@@ -0,0 +1,480 @@
...
+
+        iter_ma_maps = self.kernel_transformer.transform(
+            iter_df, masker=self.masker, return_type="array"
+        )
+        iter_of_map = self.compute_summarystat(iter_ma_maps)
```
tsalo, line 298: Using MKDA-specific terminology here. Perhaps change `of` to `ss`?

nimare/meta/cbma/base.py
```diff
@@ -0,0 +1,480 @@
...
+
+        Examples
+        --------
+        >>> meta = MKDADensity()
+        >>> result = meta.fit(dset)
+        >>> corrector = FWECorrector(method='montecarlo', voxel_thresh=0.01,
+                                     n_iters=5, n_cores=1)
+        >>> cresult = corrector.transform(result)
+        """
+        z_values = result.get_map("z", return_type="array")
```
tsalo, line 349: It looks like this array is no longer used in this method.

nimare/meta/cbma/base.py
tyarkoni: Hmm... I was going to say that it wouldn't, because there's only one pandas `DataFrame`, and the `n_iters` repetitions are all just references to the same object. But then I realized that when `n_cores > 1`, `imap` *might* end up copying the arguments in each iteration. Though I'm actually not sure—I think it might be copy-on-write, in which case the copying would only happen at the point where we reassign the ijk columns, which is going to happen anyway.

Either way, I think this is likely to be trivial compared to all the other stuff that's going on, so I'm probably not going to muck with it for now. But if we ever do a really serious optimization sweep, it will be worth taking a closer look at.

tsalo (commented):
I think the code itself looks good, although there are a few docstrings that I'd like to fill out with an eye toward explaining the inheritance structure.

One thing that I think needs to be done before this is merged is updating the API section of the documentation and checking the docs build. If you'd rather not do that, I can handle it in a follow-up PR.

tsalo:
I've opened #395 targeting the `refactor-cbma` branch to handle the API update.

nimare/meta/cbma/base.py
```diff
@@ -0,0 +1,476 @@
...
+            self._compute_null_analytic(ma_values)
+        else:
+            self._compute_null_empirical(ma_values, n_iters=self.n_iters)
+        p_values, z_values = self._summarystat_to_p(stat_values, null_method=self.null_method)
+
+        images = {"stat": stat_values, "p": p_values, "z": z_values}
+        return images
+
+    def _compute_weights(self, ma_values):
+        return None
```
tsalo, line 100: Do we need this if it isn't an abstract method?

nimare/meta/cbma/base.py
```diff
@@ -0,0 +1,476 @@
...
+        elif isinstance(data, np.ndarray):
+            ma_values = data
+        elif not isinstance(data, np.ndarray):
+            raise ValueError('Unsupported data type "{}"'.format(type(data)))
+
+        # Apply weights before returning
+        return self._compute_summarystat(ma_values)
+
+    def _compute_summarystat(self, ma_values):
+        pass
```
tsalo, line 173: Should this be an abstract method?

nimare/meta/cbma/base.py
```diff
@@ -0,0 +1,476 @@
...
+from ...base import MetaEstimator
+from ...results import MetaResult
+from ...stats import null_to_p
+from ...transforms import p_to_z
+from ...utils import round2
+
+LGR = logging.getLogger(__name__)
+
+
+class CBMAEstimator(MetaEstimator):
```
tsalo, line 23: This probably should happen now, but I'm thinking that we may want to further divide this into a `DensityBasedEstimator` and something else, so separate density-based approaches like ALE and MKDA from dramatically different ones like SDM-PSI.

nimare/meta/cbma/base.py
tyarkoni: Yes, because it serves as a dummy method for subclasses that don't take weights (e.g., `ALE`). It just allows us to avoid having different `_fit()` methods.

nimare/meta/cbma/base.py
tyarkoni: It will be, but only once we've refactored `MKDAChi2` to have the same general architecture, which I'm planning to do separately. Otherwise it will break that class. (I suppose we could define a placeholder method in `MKDAChi2` if you like.)

nimare/meta/cbma/base.py
tyarkoni: Agreed. I initially wanted to do it that way (i.e., to put all the abstracted methods into a `DensityCBMAEstimator` class), but figured we should see how far we can get by consolidating everything, instead of multiplying classes in advance. If we can't re-use everything in there right now for the pairwise estimators, we may also need to define separate `OneSample` and `Pairwise` subclasses.

tyarkoni:
@tsalo I added some docstrings. Assuming this looks good, I think we're okay to merge, unless you have other comments.

tsalo (approved):
LGTM. I'll fix the conflicts on the other PR and change the target branch once this is merged.

PR #397, opened 2020-11-14
[ENH] Add analytic null method to KDA estimator

### LINKED ISSUE #350: KDA does not work with FWECorrector(method='montecarlo')

### LINKED ISSUE #377: Implement analytical null in KDA estimator

tsalo:
Replaces #378 and closes #377. I'm also going to use this to close #350, even though the core problem was already dealt with elsewhere, I think.

- Add an analytical null method to the KDA estimator.
- Refactor KDA estimator to match new structure.


nimare/meta/cbma/mkda.py
```diff
@@ -518,168 +488,92 @@ def _compute_summarystat(self, data):
...
-        iter_max_value = np.max(iter_of_map)
-        return iter_max_value
 
-    def correct_fwe_montecarlo(self, result, n_iters=10000, n_cores=-1):
-        """
-        Perform FWE correction using the max-value permutation method.
-        Only call this method from within a Corrector.
+        def just_histogram(*args, **kwargs):
+            """Collect the first output (weights) from numpy histogram."""
+            return np.histogram(*args, **kwargs)[0].astype(float)
```
tsalo, line 542: We could use this with ALE too.

nimare/tests/test_meta_mkda.py
```diff
@@ -131,3 +148,20 @@ def test_analytic_empirical_convergence(testdata_cbma_full):
     # Correlation must be near unity and mean difference should be tiny
     assert np.corrcoef(p_analytical, p_empirical)[0, 1] > 0.98
     assert (p_analytical - p_empirical).mean() < 1e-3
+
+
+def test_kkda_analytic_empirical_convergence(testdata_cbma_full):
+    est_a = KDA(null_method="analytic")
+    n_iter = 10000
+    est_e = MKDADensity(null_method="empirical", n_iter=n_iter)
```
tsalo, line 156: I missed the second estimator, but the test is passing, which worries me a little. Especially since it's using `testdata_cbma_full`, which should have multiple foci per study.

nimare/meta/cbma/mkda.py
tsalo: I know that `apply_along_axis` is just a for loop, but I figured this made for cleaner code, even if it's not any faster.

nimare/tests/test_meta_mkda.py
tsalo: Perhaps none of the foci are near enough to one another for the KDA/MKDA distinction to matter?

tsalo: I'm sure it's not a big deal, so I'm going to resolve this.

nimare/tests/test_estimator_performance.py
```diff
@@ -404,11 +404,7 @@ def _transform_res(meta, meta_res, corr):
     #######################################
     # all combinations of meta-analysis estimators and multiple comparison correctors
     # that do not work together
-    if isinstance(meta, mkda.KDA) and corr.method == "montecarlo":
-        # TypeError: correct_fwe_montecarlo() got an unexpected keyword argument 'voxel_thresh'
-        voxel_thresh = corr.parameters.pop("voxel_thresh")
-        corr_expectation = does_not_raise()
-    elif (
+    if (
```
tsalo, line 407: @jdkent is this the correct way to update this test? Especially with regard to how `voxel_thresh` appears later on.

nimare/meta/cbma/mkda.py
tsalo: Implemented in ALE in #411.

tsalo:
I'm going to merge this so we can move forward with #383.

Issue #389, opened 2020-11-10
Replace existing "empirical" generation of uncorrected p-values with full-blown coordinate shuffling


tyarkoni:
For ALE and MKDA, we now have both analytical and permutation-based procedures for assigning uncorrected p-values to voxels. However, both methods are conservative, in that they assume voxels are independent, which is not true in practice. I.e., the current empirical permutation approach (which asymptotically converges with the analytical approach) randomly samples a single voxel from each study repeatedly, and builds up a null distribution by repeating the procedure many times. This overestimates the number of truly independent observations, and hence produces conservative correction thresholds (or uncorrected p-values).

We already support p-value generation based on actual shuffling of coordinates in the `FWECorrector`, so the proposal here is really just to provide the option to extract uncorrected p-values that way too. And since the current "empirical" approach is basically inferior to the analytical approach in all respects, we can probably just drop it entirely in favor of the full-blown coordinate shuffling. (I would probably keep these methods around but undocumented; for one thing, they're very helpful in ensuring that the analytical procedures are working as expected, as we can easily test for convergence.)

Probably we should rename the options too, as "empirical" is a bit confusing, IMO. I would suggest calling them "approximate" and "montecarlo", which would intuitively convey to users both that (a) the "approximate" approach doesn't really give the 'correct' answer (because it's biased by the independence assumption), and (b) the "montecarlo" approach is based on resampling and is likely to take a long time. Then the docs could more explicitly lay out the tradeoff, taking pain to emphasize that the "approximate" approach is actually conservative, so if you're willing to accept reduced sensitivity, it makes sense to just always use that.

[EDIT: One other minor issue: our current coordinate shuffling procedures don't take into account spatial relationships *within* a study. I.e., if two peaks overlap in the observed data, the probability of their overlap in any permutation is no greater than two peaks selected at random. I gather the reference implementation of MKDA accounts for the spatial structure in the resampling process, though I'm not sure exactly how that's done. We may want to explore this and do the same thing in NiMARE at some point.]

tsalo:
So this issue has two proposals, right? The first is a simple terminology change. `empirical` becomes `montecarlo` and `analytic` becomes `approximate`.

> We already support p-value generation based on actual shuffling of coordinates in the FWECorrector, so the proposal here is really just to provide the option to extract uncorrected p-values that way too.

[...]

PR #398, opened 2020-11-16
[ENH] Add uncorrected Monte Carlo method

tsalo:
References #389. This should address everything in #389 except for the inter-peak distance management.

- Rename "empirical" null method to "montecarlo."
- Rename "analytic" null method to "approximate."
- Implement full coordinate shuffling in montecarlo null method.
- Use shared "histogram_bins" attribute for both "approximate" and "montecarlo" null methods.
- Infer vFWE correction from montecarlo null method, and add argument to FWE corrector to _skip_ cFWE (and thus additional permutations) if only vFWE is desired.
- Add new `_determine_histogram_bins()` method that must be implemented in each Estimator (rather than in the CBMAEstimator base class). This defines the histograms for both the "montecarlo" and "approximate" null methods.

### DIFF
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -89,7 +89,7 @@ class CBMAEstimator(MetaEstimator):
         if self.null_method.startswith("approximate"):
             self._compute_null_approximate(ma_values)
         else:
-            self._compute_null_montecarlo(ma_values, n_iters=self.n_iters)
+            self._compute_null_montecarlo(n_iters=self.n_iters, n_cores=1)
         p_values, z_values = self._summarystat_to_p(stat_values, null_method=self.null_method)
 
         images = {"stat": stat_values, "p": p_values, "z": z_values}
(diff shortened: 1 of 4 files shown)


nimare/meta/cbma/base.py
```diff
@@ -260,30 +247,63 @@ def _p_to_summarystat(self, p, null_method=None):
...
             ss_idx = np.floor(p * n_vals).astype(int)
             ss = null_dist[-ss_idx]
         else:
-            raise ValueError("Argument 'null_method' must be one of: 'analytic', 'empirical'.")
+            raise ValueError("Argument 'null_method' must be one of: 'approximate', 'montecarlo'.")
 
         return ss
 
-    def _run_fwe_permutation(self, params):
+    def _run_montecarlo_permutation(self, params):
```
tsalo, line 274: This is pretty darn slow, but the histogram is probably necessary, as noted by @tyarkoni in #389.

tsalo:
The core approach is ready for review, but there's a lot more to be done to integrate it into everything (as well as the name change), so I'm marking this as a Draft. Please feel free to comment though!

tsalo:
I'm going to close this and re-implement the same approach, minus name changes, in a new PR once #397 is merged.

PR #402, opened 2020-11-19
[FIX] Sort all arrays and DataFrames in Dataset by ID

tsalo:
Closes None. I am working through the listings for the NiMARE manuscript, and I noticed that the MA maps for the first study in a Sleuth dataset didn't match up with the actual coordinates for that study.

I think there will still be an odd mismatch when there aren't coordinates for all studies in a Dataset, but I don't know what to do about that at the moment.

- Sort `Dataset.ids` and rows of all DataFrame-based Dataset attributes based on the `id` column whenever they're set.

### DIFF
--- a/nimare/dataset.py
+++ b/nimare/dataset.py
@@ -118,7 +118,7 @@ class Dataset(NiMAREBase):
 
     @ids.setter
     def _ids(self, ids):
-        ids = np.asarray(ids)
+        ids = np.sort(np.asarray(ids))
         assert isinstance(ids, np.ndarray) and ids.ndim == 1
         self.__ids = ids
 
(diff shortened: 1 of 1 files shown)


tsalo:
This is a minor fix that shouldn't adversely impact anything, so I'd like to merge today. @tyarkoni any objections/concerns?

tyarkoni:
In principle this looks fine, but if mismatches are a possibility, we should probably transition to using getters/accessors that don't depend on order at all—e.g., using `.groupby()` to check/filter that everything lines up at access time, or sticking with custom containers for `Study` and such that defer consolidation of inputs until the very last moment. We can potentially discuss this on Monday if we have time, as it bears some relation to the question of how to represent `Dataset` objects in a client library for NeuroStore.

PR #407, opened 2020-11-25
[FIX] Allow no coordinates in a dataset

### LINKED ISSUE #405: Support for image only dataset?
In creating simulated image datasets, it appears there is an expectation that [at least one study/contrast contains coordinates](https://github.com/neurostuff/NiMARE/blob/8dc6c55680c4eece74917b7d8802501e1cae0b4c/nimare/utils.py#L66-L104), since one cannot [concatenate](https://github.com/neurostuff/NiMARE/blob/8dc6c55680c4eece74917b7d8802501e1cae0b4c/nimare/utils.py#L106) 0 dataframes ([an empty list](https://github.com/neurostuff/NiMARE/blob/8dc6c55680c4eece74917b7d8802501e1cae0b4c/nimare/utils.py#L66)). If I'm only using the image-based meta-analysis estimators, is it necessary to have coordinates reported?

jdkent:
Closes #405

- return empty dataframe for coordinates to pass data validation during dataset creation.
- add test for empty dataset

I'm not sure if we want to allow an empty dataset object or explicitly test that at least one of either coordinates or images be set for this to be a valid "dataset" object.

### DIFF
--- a/nimare/utils.py
+++ b/nimare/utils.py
@@ -104,7 +104,17 @@ def dict_to_coordinates(data, masker, space):
             all_dfs.append(con_df)
 
     if not all_dfs:
-        return pd.DataFrame({"id": []})
+        return pd.DataFrame(
+            {
+                "id": [],
+                "study_id": [],
+                "contrast_id": [],
+                "x": [],
+                "y": [],
+                "z": [],
+                "space": [],
+            },
+        )
 
     df = pd.concat(all_dfs, axis=0, join="outer", sort=False)
     df = df[columns].reset_index(drop=True)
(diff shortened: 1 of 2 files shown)


nimare/utils.py
```diff
@@ -103,6 +103,9 @@ def dict_to_coordinates(data, masker, space):
             con_df = pd.DataFrame(temp_data.T, columns=exp_columns)
             all_dfs.append(con_df)
 
+    if not all_dfs:
+        return pd.DataFrame({"id": []})
```
tsalo, line 107: Could you include these columns as well: "study_id", "contrast_id", "x", "y", "z", and "space".

tsalo (changes requested):
Minor request. Other than that these changes make sense. Thanks!

tsalo (approved):
The changes look good to me! I think we could expand the testing at some point, but I'm fine with a smoke test for now.

jdkent:
merging to allow my continuing work on ibma generation

PR #409, opened 2020-11-25
[FIX] Add nullhist_to_p and crop invalid p-values

### LINKED ISSUE #403: Statistic-to-p conversion can produce infinite values for analytic null
The summary statistic-to-p conversion from histograms can result in infinite values (well, p-values of 0 and infinite negative log-10 p-values). I assume this is an issue that can be dealt with in the actual conversion, rather than the histograms.

Somewhat similar to #364.

tsalo:
Closes #403. 

- Add `nimare.stats.nullhist_to_p()` function for converting test statistics to p-values using a null distribution stored in a histogram.
- Use `nimare.stats.nullhist_to_p()` in CBMA estimators.
- Crop out-of-bounds p-values to ~~epsilon~~ smallest observed non-zero p-value.

TODO:

- [X] Replace epsilon with better value when cropping.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ orbs:
   codecov: codecov/codecov@1.0.5
 jobs:
 
-  build_py36:
+  make_py36_env:
     working_directory: /tmp/src/NiMARE
     docker:
       - image: continuumio/miniconda3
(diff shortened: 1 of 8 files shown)


tsalo:
Merging this now so I can push forward with the other changes to the empirical and analytic nulls.

PR #410, opened 2020-11-26
[FIX] Restructure Peaks2MapsKernel to operate like other kernels

### LINKED ISSUE #347: Peaks2MapsKernel does not work with any CBMA algorithms
While the actual utility of the kernel is unknown, we still want the kernel transformer to successfully interact with the estimators. The error is `AttributeError: 'DataFrame' object has no attribute 'masker'`. Based on that error message, I think this should be reasonable easy to solve. We just need to make sure that the `Peaks2MapsKernel` can accept alternate input types, the same way the other kernel transformers can.

Stems from #343.

tsalo:
Closes #347.

- Restructure `Peaks2MapsKernel` to follow the `KernelTransformer` convention.
- Rename `compute_ma` to `compute_ale_ma` to match convention.
- Rename `peaks2maps` to `compute_p2m_ma` to match convention.
- Do not test `Peaks2MapsKernel` in `test_estimator_performance.py` (see #417).

### DIFF
--- a/nimare/extract/extract.py
+++ b/nimare/extract/extract.py
@@ -384,6 +384,7 @@ def download_peaks2maps_model(data_dir=None, overwrite=False, verbose=1):
     url = "https://zenodo.org/record/1257721/files/ohbm2018_model.tar.xz?download=1"
 
     temp_dataset_name = "peaks2maps_model_ohbm2018__temp"
+    data_dir = _get_dataset_dir("", data_dir=data_dir, verbose=verbose)
     temp_data_dir = _get_dataset_dir(temp_dataset_name, data_dir=data_dir, verbose=verbose)
 
     dataset_name = "peaks2maps_model_ohbm2018"
--- a/nimare/extract/utils.py
+++ b/nimare/extract/utils.py
@@ -138,7 +138,7 @@ def _get_dataset_dir(dataset_name, data_dir=None, default_paths=None, verbose=1)
                 errors.append("\n -{0} ({1})".format(path, short_error_message))
 
     raise OSError(
-        "NiMARE tried to store the dataset in the following " "directories, but:" + "".join(errors)
+        "NiMARE tried to store the dataset in the following directories, but: " + "".join(errors)
     )
 
 
--- a/nimare/generate.py
+++ b/nimare/generate.py
@@ -5,7 +5,7 @@ import numpy as np
 import nilearn
 
 from .dataset import Dataset
-from .meta.utils import compute_ma, get_ale_kernel
+from .meta.utils import compute_ale_ma, get_ale_kernel
 from .transforms import vox2mm, mm2vox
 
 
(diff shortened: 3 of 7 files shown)


tsalo:
The jobs are being killed, which means that I need to take a look at them and see if I can reduce the computational load.

jdkent:
Looks like [_compute_null_analytic](https://github.com/neurostuff/NiMARE/blob/59276130bb373389b5aa7d36e2ba351533056a8a/nimare/meta/cbma/ale.py#L99) for the ALE estimator is a source of the increased memory consumption. 
Specifically here: 
https://github.com/neurostuff/NiMARE/blob/59276130bb373389b5aa7d36e2ba351533056a8a/nimare/meta/cbma/ale.py#L147-L151

ALE has 100s of non-zero histogram bins, whereas p2m has 1000s of non-zero bins, creating the variables `ale_scores`, `score_idx`, and `probabilities` with millions of entries.

A potential solution could be to save those variables with single precision (i.e., float32) as opposed to double precision (float64) if there are more than say 2000 `exp_idx` or `ale_idx`. This appears to be meaningful precision loss when computing `score_idx` using `ale_scores` as a float32 or float64 in my example:
```
ale_scores32 = 1 - np.outer(1 - hist_bins[exp_idx], 1 - hist_bins[ale_idx]).ravel().astype(np.float32)
ale_scores64 = ale_scores = 1 - np.outer(1 - hist_bins[exp_idx], 1 - hist_bins[ale_idx]).ravel()

ale_scores32.size # 38529432

score_idx32 = np.floor(ale_scores32 * inv_step_size).astype(int)
score_idx64 = np.floor(ale_scores64 * inv_step_size).astype(int)

(score_idx64 == score_idx32).sum() # 38506719
(score_idx64 != score_idx32).sum() # 22713
```
so this may not be a good solution.

Another potential solution is to turn off tests for ale(analytic)+p2m_kernel since they take too much RAM.

tsalo:
Our `ale_scores` precision should be determined by the precision of our histogram bins, I think. The smallest difference we _should_ have to be able to track in `ale_scores` should correspond to our histogram bin resolution (in #411 it's 1e-5) squared (i.e., 1e-10). We should be able to determine the lowest-resolution float we can use based on our histogram bins using [`numpy.min_scalar_type`](https://numpy.org/doc/stable/reference/generated/numpy.min_scalar_type.html#numpy.min_scalar_type), and then convert our values ahead of time!

tsalo:
I'm just going to excise all precision-related changes and drop Peaks2MapsKernel from the performance tests.

We should open a new issue about the problem, though.

tsalo:
The docs build is failing, but I don't think that's tied to changes in this PR (see #418), so I am going to merge if all other tests pass.

nimare/extract/extract.py
```diff
@@ -384,6 +384,7 @@ def download_peaks2maps_model(data_dir=None, overwrite=False, verbose=1):
     url = "https://zenodo.org/record/1257721/files/ohbm2018_model.tar.xz?download=1"
 
     temp_dataset_name = "peaks2maps_model_ohbm2018__temp"
+    data_dir = _get_dataset_dir("", data_dir=data_dir, verbose=verbose)
```
tsalo, line 387: This will probably help with #367 and #368..

PR #411, opened 2020-11-28
[ENH] Improve convergence between ALE null methods

### LINKED ISSUE #396: Improve convergence between ALE analytic and empirical null methods
ALE analytic method converges with empirical  more when:
```
step_size = 0.00001
hist_bins = np.round(np.arange(0, max_poss_ale + 0.001, step_size), 5)
```
the ale analytic & empirical correlation is: 0.9999446144830378

Here's the false positive rates:
image

the p-value distribution:
image

p-value correlation:
image

So increasing the granularity of the hist_bins by 10x for ALE appears to bring the analytic and empirical methods really close to each other (but this may bite into the performance gains that Tal has been working on)

_Originally posted by @jdkent in https://github.com/neurostuff/NiMARE/issues/383#issuecomment-727057249_

tsalo:
Closes #396.

- Increase resolution of ALE analytic null histogram from bins of 0.0001 to bins of 0.00001.
- Round up maximum MA values based on target resolution in ALE histogram generation so that we don't need an arbitrary buffer.
    - This involves including one extra bin in the histogram that has an impossible value (e.g., 1.000001). See https://github.com/neurostuff/NiMARE/pull/411#issuecomment-736996415 for rationale.
- Simplify histogram generation in ALE.
    - This involves cropping histogram bins when shifting from edges to centers.
- ~~Also increase resolution of SCALE analytic null histogram by same factor.~~ Reverted since it was too memory-intensive for the CI.

### DIFF
--- a/nimare/meta/cbma/ale.py
+++ b/nimare/meta/cbma/ale.py
@@ -117,22 +117,28 @@ class ALE(CBMAEstimator):
             raise ValueError('Unsupported data type "{}"'.format(type(ma_maps)))
 
         # Determine bins for null distribution histogram
+        # Remember that numpy histogram bins are bin edges, not centers
+        # Assuming values of 0, .001, .002, etc., bins are -.0005-.0005, .0005-.001, etc.
+        inv_step_size = 100000
+        step_size = 1 / inv_step_size
         max_ma_values = np.max(ma_values, axis=1)
+        # round up based on resolution
+        max_ma_values = np.ceil(max_ma_values * inv_step_size) / inv_step_size
         max_poss_ale = self.compute_summarystat(max_ma_values)
-        step_size = 0.00001
-        hist_bins = np.round(np.arange(0, max_poss_ale + 0.001, step_size), 5)
+        # create bin centers, then shift them into bin edges
+        hist_bins = np.round(
+            np.arange(0, max_poss_ale + step_size, step_size), 5
+        ) - (step_size / 2)
+
+        def just_histogram(*args, **kwargs):
+            """Collect the first output (weights) from numpy histogram."""
+            return np.histogram(*args, **kwargs)[0].astype(float)
+
+        ma_hists = np.apply_along_axis(just_histogram, 1, ma_values, bins=hist_bins, density=False)
+
+        # Shift the bins to correspond to centers instead of edges
+        hist_bins += (step_size / 2)
         self.null_distributions_["histogram_bins"] = hist_bins
-
-        ma_hists = np.zeros((ma_values.shape[0], hist_bins.shape[0]))
-
-        n_zeros = np.sum(ma_values == 0, 1)
-        ma_hists[:, 0] = n_zeros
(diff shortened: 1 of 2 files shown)


nimare/meta/cbma/ale.py
```diff
@@ -117,22 +117,28 @@ def _compute_null_analytic(self, ma_maps):
...
+        # round up based on resolution
+        max_ma_values = np.ceil(max_ma_values * inv_step_size) / inv_step_size
         max_poss_ale = self.compute_summarystat(max_ma_values)
-        step_size = 0.0001
-        hist_bins = np.round(np.arange(0, max_poss_ale + 0.001, step_size), 4)
-        self.null_distributions_["histogram_bins"] = hist_bins
-
-        ma_hists = np.zeros((ma_values.shape[0], hist_bins.shape[0]))
+        # create bin centers, then shift them into bin edges
+        hist_bins = np.round(np.arange(0, max_poss_ale + (1.5 * step_size), step_size), 5) - (
```
tsalo, line 129: The 1.5 _should_ result in the maximum possible statistic value falling within the last bin, with no trailing bins after that.

tsalo:
The only failing test is `test_corr_transform_smoke[fwe_montecarlo-normal_data-ale+analytic-mkda_kernel]`. I'm not sure what the problem is.

jdkent:
the error is coming from [this line](https://github.com/neurostuff/NiMARE/pull/411/files#diff-4cafb2c711a65e5b19450202efbd9250dc086c814fb97588c1fd88afcc1add0cR270)
```
ss_idx = np.maximum(0, np.where(hist_weights <= p)[0][0] - 1)
```
the error occurs when `p` is less than any of the `hist_weights`. If `p = 0.05` and `hist_weights = array([1.0, 0.5, 0.2])`, then `np.where(hist_weights <= p)` returns `(array([], dtype=int64),)`. Since the array is empty, it does not have an index of `[0]`.

This code fixes the immediate problem, I'll take a look at the code to see if there is something else strange going on:
```
ss_idx = np.maximum(
                0,
                np.where(
                    hist_weights <= p,
                    np.arange(hist_weights.size),
                    hist_weights.size,
                ).min() - 1,
            )
```

tsalo:
Thank you for digging into this! It seems like the issue, then, is that the maximum possible value for an MKDA kernel + the ALE summary statistic formula (i.e., 1) is fairly highly probable, while any summary statistic value above that is impossible, and thus p-values associated with those higher values drop down to zero. By not including impossibly high values in the histogram, I was removing these p=0 locations.

One solution that seems to work is to simply include one extra bin when I create the histogram bins. Thus, the last bin's p-value will always be zero (since it exceeds the maximum possible summary statistic value), and when we grab the bin _before_ the identified one, we will grab the maximum possible value's bin.

tsalo:
I think the extra bin solution works and makes sense, so I'm going to merge this.

PR #420, opened 2020-12-15
[DOC] Warn users about Peaks2MapsKernel and KDA Estimator

### LINKED ISSUE #417: Peaks2MapsKernel+ALE is too memory-intensive for tests
Looks like [_compute_null_analytic](https://github.com/neurostuff/NiMARE/blob/59276130bb373389b5aa7d36e2ba351533056a8a/nimare/meta/cbma/ale.py#L99) for the ALE estimator is a source of the increased memory consumption. 
Specifically here: 
https://github.com/neurostuff/NiMARE/blob/59276130bb373389b5aa7d36e2ba351533056a8a/nimare/meta/cbma/ale.py#L147-L151

ALE has 100s of non-zero histogram bins, whereas p2m has 1000s of non-zero bins, creating the variables `ale_scores`, `score_idx`, and `probabilities` with millions of entries.

A potential solution could be to save those variables with single precision (i.e., float32) as opposed to double precision (float64) if there are more than say 2000 `exp_idx` or `ale_idx`. This appears to be meaningful precision loss when computing `score_idx` using `ale_scores` as a float32 or float64 in my example:
```
ale_scores32 = 1 - np.outer(1 - hist_bins[exp_idx], 1 - hist_bins[ale_idx]).ravel().astype(np.float32)
ale_scores64 = ale_scores = 1 - np.outer(1 - hist_bins[exp_idx], 1 - hist_bins[ale_idx]).ravel()

ale_scores32.size # 38529432

score_idx32 = np.floor(ale_scores32 * inv_step_size).astype(int)
score_idx64 = np.floor(ale_scores64 * inv_step_size).astype(int)

(score_idx64 == score_idx32).sum() # 38506719
(score_idx64 != score_idx32).sum() # 22713
```
so this may not be a good solution.

Another potential solution is to turn off/skip tests for ale(analytic)+p2m_kernel since they take too much RAM.

_Originally posted by @jdkent in https://github.com/neurostuff/NiMARE/issues/410#issuecomment-736855179_

tsalo:
Closes #417, in that we will not test Peaks2MapsKernel in conjunction with any meta-analytic estimators, so any issues with memory usage can be ignored.

- Add warning to docstring and log warning at initialization of Peaks2MapsKernel, to clarify to users that they should **not** use the kernel for any serious research.
- Add warning to docstring and log warning at initialization of KDA estimator, to make it clear that the KDA estimator is inferior to MKDA and should not be used for real meta-analyses.

### DIFF
--- a/nimare/meta/cbma/mkda.py
+++ b/nimare/meta/cbma/mkda.py
@@ -456,10 +456,12 @@ class KDA(CBMAEstimator):
     Kernel density analysis was first introduced in [1]_ and [2]_.
 
     Available correction methods: :func:`KDA.correct_fwe_montecarlo`
+
     Warning
     -------
-    The KDA meta-analytic algorithm is objectively inferior to the MKDA algorithm.
-    Please do not use this Estimator for serious meta-analyses.
+    The KDA algorithm has been replaced in the literature with the MKDA algorithm.
+    As such, this estimator should almost never be used, outside of systematic
+    comparisons between algorithms.
 
     References
     ----------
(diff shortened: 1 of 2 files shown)


nimare/meta/cbma/mkda.py
```diff
@@ -456,6 +456,11 @@ class KDA(CBMAEstimator):
     Kernel density analysis was first introduced in [1]_ and [2]_.
 
     Available correction methods: :func:`KDA.correct_fwe_montecarlo`
+
```
tsalo, line 459:
```suggestion

```

PR #424, opened 2020-12-16
[ENH] Implement full coordinate-set empirical null method 

### LINKED ISSUE #389: Replace existing "empirical" generation of uncorrected p-values with full-blown coordinate shuffling

tsalo:
Closes #389.

- Add `_determine_histogram_bins()` method to CBMA Estimators. This method determines the bins for the analytic null and the full empirical null, but not the basic (single-coordinate) empirical null, which doesn't use a histogram.
- Add `_run_montecarlo_permutation()` and `_compute_null_reduced_empirical()` methods to base CBMAEstimator.
- Edit `_compute_null_empirical()` to run full coordinate-set method.
- Add `vfwe_only` parameter to the standard Monte Carlo permutation method for FWE correction.
- Directly infer the weighting factor for the kernel in the KDA estimator instead of trying to guess it from MA values.
    - This is currently incompatible with continuous kernels, but I may change that before merging this PR.

### DIFF
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -89,9 +89,7 @@ class CBMAEstimator(MetaEstimator):
             self._compute_null_analytic(ma_values)
 
         elif self.null_method == "empirical":
-            self._compute_null_empirical(
-                ma_values, n_iters=self.n_iters, n_cores=self.n_cores
-            )
+            self._compute_null_empirical(ma_values, n_iters=self.n_iters, n_cores=self.n_cores)
 
         else:
             self._compute_null_reduced_empirical(ma_values, n_iters=self.n_iters)
@@ -228,7 +226,7 @@ class CBMAEstimator(MetaEstimator):
             p_values = null_to_p(
                 stat_values,
                 self.null_distributions_["values_corr-none_method-reducedEmpirical"],
-                tail="upper"
+                tail="upper",
             )
 
         else:
@@ -537,8 +535,7 @@ class CBMAEstimator(MetaEstimator):
                 with mp.Pool(n_cores) as p:
                     perm_results = list(
                         tqdm(
-                            p.imap(self._correct_fwe_montecarlo_permutation, params),
-                            total=n_iters
+                            p.imap(self._correct_fwe_montecarlo_permutation, params), total=n_iters
                         )
                     )
 
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -85,8 +85,11 @@ class CBMAEstimator(MetaEstimator):
         stat_values = self.compute_summarystat(ma_values)
 
         # Determine null distributions for summary stat (OF) to p conversion
+        self._determine_histogram_bins(ma_values)
         if self.null_method.startswith("analytic"):
             self._compute_null_analytic(ma_values)
+        elif self.null_method == "empirical_full":
+            self._compute_null_empirical_full(ma_values, n_iters=self.n_iters)
         else:
             self._compute_null_empirical(ma_values, n_iters=self.n_iters)
         p_values, z_values = self._summarystat_to_p(stat_values, null_method=self.null_method)
@@ -219,7 +222,7 @@ class CBMAEstimator(MetaEstimator):
         """
         if null_method.startswith("analytic"):
             assert "histogram_bins" in self.null_distributions_.keys()
-            assert "histogram_weights" in self.null_distributions_.keys()
+            assert "histweights_corr-none_method-analytic" in self.null_distributions_.keys()
 
             inv_step_size = round2(1 / np.diff(self.null_distributions_["histogram_bins"])[0])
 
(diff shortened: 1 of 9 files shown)


tsalo:
The basic gist here seems to be that we'll need to be careful with our tests since this new null method seems to get killed very easily.

nimare/meta/cbma/mkda.py
```diff
@@ -45,12 +45,18 @@ class MKDADensity(CBMAEstimator):
     """
 
     def __init__(
-        self, kernel_transformer=MKDAKernel, null_method="empirical", n_iters=10000, **kwargs
+        self,
+        kernel_transformer=MKDAKernel,
+        null_method="empirical",
```
tsalo, line 50:
```suggestion
        null_method="analytic",
```

nimare/meta/cbma/mkda.py
```diff
@@ -467,7 +494,12 @@ class KDA(CBMAEstimator):
     """
 
     def __init__(
-        self, kernel_transformer=KDAKernel, null_method="empirical", n_iters=10000, **kwargs
+        self,
+        kernel_transformer=KDAKernel,
+        null_method="empirical",
```
tsalo, line 499:
```suggestion
        null_method="analytic",
```

tsalo:
`build_docs` is failing on the combination of ALEKernel and KDA estimator because of the way that the KDA analytic null method is implemented. The KDA method uses the smallest observed MA value to define its histogram bins, which is great for the KDA kernel where we don't know the kernel's value, but which is very bad for the ALE kernel, which is continuous, since it results in an oversized histogram.

I still need to dig into the other workflows, which seem to be killed on different tests.

tsalo:
Current failures:

- test_meta_fit_performance[normal_data-ale+analytic-mkda_kernel]
- test_meta_fit_performance[normal_data-ale+analytic-kda_kernel]
- test_meta_fit_performance[normal_data-mkda+empirical-ale_kernel]
- test_corr_transform_performance[fwe_bonferroni-normal_data-ale+analytic-kda_kernel]
- test_corr_transform_performance[fwe_bonferroni-normal_data-ale+analytic-mkda_kernel]
- test_corr_transform_performance[fwe_bonferroni-normal_data-mkda+empirical-ale_kernel]
- test_corr_transform_performance[fdr_indep-normal_data-ale+analytic-kda_kernel]
- test_corr_transform_performance[fdr_indep-normal_data-ale+analytic-mkda_kernel]
- test_corr_transform_performance[fdr_indep-normal_data-mkda+empirical-ale_kernel]
- test_corr_transform_performance[fdr_negcorr-normal_data-ale+analytic-kda_kernel]
- test_corr_transform_performance[fdr_negcorr-normal_data-ale+analytic-mkda_kernel]
- test_corr_transform_performance[fdr_negcorr-normal_data-mkda+empirical-ale_kernel]

jdkent:
It looks like the tests are failing because when I was trying to tease apart whether the estimator/kernel/null_method/corrector combinations "worked", I bundled together my idea of "sensitivity" and "specificity", namely, whether the combination found significant results in the regions where I expected significant results (sensitivity) and non-significant results everywhere else (specificity). Until now, if a combination had poor sensitivity it also had poor specificity(?), but with this pull request, the combinations have either poor sensitivity or specificity. There is a peculiar divergence between the analytic and "empirical" null methods with the ALE estimator with the KDA or MKDA kernels. The analytic null method leads to many voxels being declared significant, but the empirical nulls have the opposite problem. I could not find a problem with how the null were calculated, so I am inclined to think the diverging issues are an idiosyncrasy of using the ALE/(M)KDA Kernel combinations with different null methods.

I'll submit a pull request to this branch fixing the tests once they pass locally.

  | ALEKernel | KDAKernel | MKDAKernel |  
-- | -- | -- | -- | --
ALE | Good | Poor Specificity | Poor Specificity | analytic
  | Good | Poor Sensitivity | Poor Sensitivity | empirical
  | Good | Poor Sensitivity | Poor Sensitivity | reduced empirical
KDA | Good | Good | Good | analytic
  | Good | Good | Good | empirical
  | Good | Good | Good | reduced empirical
MKDA | Poor Sensitivity | Good | Good | analytic
  | Poor Sensitivity | Good | Good | empirical
  | Good | Good | Good | reduced empirical

nimare/meta/cbma/base.py
```diff
@@ -218,18 +204,31 @@ def _summarystat_to_p(self, stat_values, null_method="analytic"):
...
+            assert "values_corr-none_method-reducedEmpirical" in self.null_distributions_.keys()
             p_values = null_to_p(
-                stat_values, self.null_distributions_["empirical_null"], tail="upper"
+                stat_values,
+                self.null_distributions_["values_corr-none_method-reducedEmpirical"],
+                tail="upper",
             )
+
         else:
             raise ValueError("Argument 'null_method' must be one of: 'analytic', 'empirical'.")
```
jdkent, line 233: should we add `reduced_empirical`?
```suggestion
            raise ValueError("Argument 'null_method' must be one of: 'analytic', 'empirical', 'reduced_empirical'.")
```

tsalo: Not sure. I think @tyarkoni wanted to leave it undocumented, since the reduced empirical method should only be used for testing once we have the full empirical method merged in.

jdkent:
Oh, I have a pull request that _should_ fix the tests tsalo/NiMARE#7, if failed tests are holding this pull request back

PR #425, opened 2020-12-17
[FIX] Fix the warnings about mismatched kernels and estimators

tsalo:
Closes None, but fixes a documentation bug I introduced in #416.

- Check kernel transformer as both an instance and a class. Before it just checked for instances.


tsalo:
I checked the `build_docs` workflow and there are no longer any warnings for matched kernel+estimator combos- just mismatched ones. Going to merge now.

Issue #341, opened 2020-09-15
Converter for nilearn fetched NeuroVault images


tsalo:
Add a function to convert NeuroVault image bunches fetched by [`nilearn.datasets.fetch_neurovault()`](http://nilearn.github.io/modules/generated/nilearn.datasets.fetch_neurovault.html#nilearn.datasets.fetch_neurovault) to NiMARE Datasets. To be moved to neurostuff at a later date.

tsalo:
We skipped past nilearn directly to NeuroVault with #432, so I'm going to close this.

PR #432, opened 2021-01-05
[ENH] add function to convert neurovault collections to a NiMARE dataset

jdkent:
references #341

- creates a downloader for neurovault and converts to a NiMARE data, but I did not see the nilearn function referenced in #341, will refactor.

### DIFF
--- a/examples/01_datasets/plot_neurovault_io.py
+++ b/examples/01_datasets/plot_neurovault_io.py
@@ -22,7 +22,7 @@ Using Neurovault Statistical Maps in NiMARE
 # * `3264 <https://neurovault.org/collections/3264/>`_,
 # * `3192 <https://neurovault.org/collections/3192/>`_,
 # * `457 <https://neurovault.org/collections/457/>`_.
-# 
+#
 # I can load specific statistical maps from these collections
 # into a NiMARE dataset:
 from nimare.io import convert_neurovault_to_dataset
@@ -35,15 +35,15 @@ collection_ids = (2884, 2621, 3085, 5623, 3264, 3192, 457)
 # named in the dataset and what their respective group
 # statistical maps are named on neurovault
 contrasts = {
-    "working_memory": [
-        "Working memory load of 2 faces versus 1 face - NT2_Tstat",
-        "t-value contrast 2-back minus 0-back",
-        "Searchlight multivariate Decoding 2: visual working memory",
-        "Context-dependent group-specific WM information",
-        "WM working memory zstat1",
-        "WM task over CRT task map",
-        "tfMRI WM 2BK PLACE zstat1",
-    ]
+    "working_memory": (
+        "Working memory load of 2 faces versus 1 face - NT2_Tstat|"
+        "t-value contrast 2-back minus 0-back|"
+        "Searchlight multivariate Decoding 2: visual working memory|"
+        "Context-dependent group-specific WM information|"
+        "WM working memory zstat1|"
+        "WM task over CRT task map|"
+        "tfMRI WM 2BK PLACE zstat1"
+    )
 }
 
 # Convert how the statistical maps on neurovault are represented
--- a/examples/01_datasets/plot_neurovault_io.py
+++ b/examples/01_datasets/plot_neurovault_io.py
@@ -1,0 +1,92 @@
+"""
+
+.. _datasets2:
+
+===========================================
+Using Neurovault Statistical Maps in NiMARE
+===========================================
+
+"""
+###############################################################################
+# Neurovault + NiMARE: Load freely shared statistical maps for Meta-Analysis
+# --------------------------------------------------------------------------
+# `Neurovault <https://neurovault.org/>`_ is an online platform that hosts
+# unthresholded statistical maps, including group statistical maps.
+# NiMARE can read these statistical maps when given a list of collection_ids.
+# I search "working memory" on neurovault, and find these relevant collections:
+# * `2884 https://neurovault.org/collections/2884/`_,
+# * `2621 https://neurovault.org/collections/2621/`_,
+# * `3085 https://neurovault.org/collections/3085/`_,
+# * `5623 https://neurovault.org/collections/5623/`_,
+# * `3264 https://neurovault.org/collections/3264/`_,
+# * `3192 https://neurovault.org/collections/3192/`_,
+# * `457 https://neurovault.org/collections/457/`_.
+# I can load specific statistical maps from these collections
+# into a NiMARE dataset:
+from nimare.io import convert_neurovault_to_dataset
+
+# The specific collections I would like to download group level
+# statistical maps from
+collection_ids = (2884, 2621, 3085, 5623, 3264, 3192, 457)
+
+# A mapping between what I want the contrast(s) to be
+# named in the dataset and what their respective group
+# statistical maps are named on neurovault
+contrasts = {
+    "working_memory": [
+        "Working memory load of 2 faces versus 1 face - NT2_Tstat",
+        "t-value contrast 2-back minus 0-back",
+        "Searchlight multivariate Decoding 2: visual working memory",
+        "Context-dependent group-specific WM information",
+        "WM working memory zstat1",
+        "WM task over CRT task map",
+        "tfMRI WM 2BK PLACE zstat1",
+    ]
+}
+
+# Convert how the statistical maps on neurovault are represented
+# in a NiMARE dataset
+map_type_conversion = {"Z map": "z", "T map": "t"}
+
+dset = convert_neurovault_to_dataset(
+    collection_ids,
+    contrasts,
+    img_dir=None,
+    map_type_conversion=map_type_conversion,
+)
+
+###############################################################################
+# Conversion of Statistical Maps
+# ------------------------------
+# Some of the statistical maps are T statistics and others are Z statistics.
+# To perform a Fisher's meta analysis, we need all Z maps.
+# Luckily, NiMARE has a function named ``transform_images`` that will
+# help us.
+from nimare.transforms import transform_images
+
+# Not all studies have Z maps!
+print(dset.images["z"])
+
+dset.images = transform_images(
+    dset.images, target="z", masker=dset.masker, metadata_df=dset.metadata
+)
+
+# All studies have Z maps!
+print(dset.images["z"])
+
+
+###############################################################################
+# Run a Meta-Analysis
+# -------------------
+# With the missing Z maps filled in, we can run a Meta-Analysis
+# and plot our results
+from nimare.meta.ibma import Fishers
+from nilearn.plotting import plot_stat_map
+
+meta = Fishers()
+
+meta_res = meta.fit(dset)
+
+plot_stat_map(meta_res.get_map("z"), threshold=3.3)
+# The result may look questionable, but this code provides
+# a template on how to use neurovault in your meta analysis.
--- a/examples/01_datasets/plot_neurovault_io.py
+++ b/examples/01_datasets/plot_neurovault_io.py
@@ -14,13 +14,15 @@ Using Neurovault Statistical Maps in NiMARE
 # unthresholded statistical maps, including group statistical maps.
 # NiMARE can read these statistical maps when given a list of collection_ids.
 # I search "working memory" on neurovault, and find these relevant collections:
-# * `2884 https://neurovault.org/collections/2884/`_,
-# * `2621 https://neurovault.org/collections/2621/`_,
-# * `3085 https://neurovault.org/collections/3085/`_,
-# * `5623 https://neurovault.org/collections/5623/`_,
-# * `3264 https://neurovault.org/collections/3264/`_,
-# * `3192 https://neurovault.org/collections/3192/`_,
-# * `457 https://neurovault.org/collections/457/`_.
+#
+# * `2884 <https://neurovault.org/collections/2884/>`_,
+# * `2621 <https://neurovault.org/collections/2621/>`_,
+# * `3085 <https://neurovault.org/collections/3085/>`_,
+# * `5623 <https://neurovault.org/collections/5623/>`_,
+# * `3264 <https://neurovault.org/collections/3264/>`_,
+# * `3192 <https://neurovault.org/collections/3192/>`_,
+# * `457 <https://neurovault.org/collections/457/>`_.
+# 
 # I can load specific statistical maps from these collections
 # into a NiMARE dataset:
 from nimare.io import convert_neurovault_to_dataset
--- a/examples/01_datasets/plot_neurovault_io.py
+++ b/examples/01_datasets/plot_neurovault_io.py
@@ -15,13 +15,13 @@ Using Neurovault Statistical Maps in NiMARE
 # NiMARE can read these statistical maps when given a list of collection_ids.
 # I search "working memory" on neurovault, and find these relevant collections:
 #
-# * `2884 <https://neurovault.org/collections/2884/>`_,
-# * `2621 <https://neurovault.org/collections/2621/>`_,
-# * `3085 <https://neurovault.org/collections/3085/>`_,
-# * `5623 <https://neurovault.org/collections/5623/>`_,
-# * `3264 <https://neurovault.org/collections/3264/>`_,
-# * `3192 <https://neurovault.org/collections/3192/>`_,
-# * `457 <https://neurovault.org/collections/457/>`_.
+# * `2884 <https://neurovault.org/collections/2884/>`_
+# * `2621 <https://neurovault.org/collections/2621/>`_
+# * `3085 <https://neurovault.org/collections/3085/>`_
+# * `5623 <https://neurovault.org/collections/5623/>`_
+# * `3264 <https://neurovault.org/collections/3264/>`_
+# * `3192 <https://neurovault.org/collections/3192/>`_
+# * `457 <https://neurovault.org/collections/457/>`_
 #
 # I can load specific statistical maps from these collections
 # into a NiMARE dataset:
@@ -47,7 +47,7 @@ contrasts = {
 }
 
 # Convert how the statistical maps on neurovault are represented
-# in a NiMARE dataset
+# in a NiMARE dataset.
 map_type_conversion = {"Z map": "z", "T map": "t"}
 
 dset = convert_neurovault_to_dataset(
@@ -62,7 +62,7 @@ dset = convert_neurovault_to_dataset(
 # ------------------------------
 # Some of the statistical maps are T statistics and others are Z statistics.
 # To perform a Fisher's meta analysis, we need all Z maps.
-# Luckily, NiMARE has a function named ``transform_images`` that will
+# Thoughtfully, NiMARE has a function named ``transform_images`` that will
 # help us.
 from nimare.transforms import transform_images
 
--- a/nimare/generate.py
+++ b/nimare/generate.py
@@ -116,13 +116,15 @@ def create_coordinate_dataset(
     return ground_truth_foci, dataset
 
 
-def create_neurovault_dataset(collection_ids=NEUROVAULT_IDS, contrast=CONTRAST_OF_INTEREST, img_dir=None):
+def create_neurovault_dataset(
+    collection_ids=NEUROVAULT_IDS, contrast=CONTRAST_OF_INTEREST, img_dir=None,
+    **dset_kwargs):
     """
     Create a dataset from neurovault collections, assuming the neurovault collections
     have beta, t, and varcope maps, and z-maps need to be created.
 
     """
-    dataset = convert_neurovault_to_dataset(collection_ids, contrast, img_dir)
+    dataset = convert_neurovault_to_dataset(collection_ids, contrast, img_dir, **dset_kwargs)
     dataset.images = transform_images(
         dataset.images, target="z", masker=dataset.masker, metadata_df=dataset.metadata
     )
--- a/nimare/generate.py
+++ b/nimare/generate.py
@@ -12,7 +12,7 @@ from .io import convert_neurovault_to_dataset
 
 # defaults for creating a neurovault dataset
 NEUROVAULT_IDS = (8836, 8838, 8893, 8895, 8892, 8891, 8962, 8894, 8956, 8854, 9000)
-CONTRAST_OF_INTEREST = "as-Animal"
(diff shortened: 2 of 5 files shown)


tsalo:
I don't think a nilearn-specific method is necessary. I just figured that, since nilearn already has functions for fetching from neurovault, we should leverage them. I'm happy to have a version that doesn't use nilearn, as long as it's not too bloated/unwieldy.

tyarkoni:
I don't have any principled objection to using the nilearn utility, though my experience with it a while back was that it's a bit clunky (but I was doing fairly specific things, so it's possible it was just the use case). @jdkent feel free to look into it, and if you decide it's cleaner to use it, that's fine with me.

nimare/io.py
```diff
@@ -275,3 +279,87 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+    """
+    if img_dir is None:
+        img_dir = Path(tempfile.mkdtemp(prefix="neurovault_"))
+    else:
+        img_dir = Path(img_dir)
+        img_dir.mkdir(parents=True, exist_ok=False)
+
+    map_type_conversion = {
+        "T map": "t",
+        "variance": "varcope",
```
tsalo, line 314: I'm not sure if we can assume that NeuroVault "variance" maps correspond to varcopes.

tyarkoni:
> I'm not sure if we can assume that NeuroVault "variance" maps correspond to varcopes.

Agreed that it's definitely not safe to assume that maps named "variance" are varcopes; OTOH, without that assumption, there's not going to be any automated way to identify the variance maps, and users will be left to identify all map types manually. So I think it's reasonable to use that as a heuristic, and just make it clear to users that it's their responsibility to verify that the maps mean what they think they mean (either before or after the fact).

If we want to be careful about it, we could check the image arrays to verify that all values are >= 0 (because variances can't be negative), and then issue a warning (or possibly even exception) if that's not the case. But of course that still provides no guarantees.

tsalo:
>If we want to be careful about it, we could check the image arrays to verify that all values are >= 0 (because variances can't be negative), and then issue a warning (or possibly even exception) if that's not the case. But of course that still provides no guarantees.

I think it would probably be good to (1) try to use other maps to get variance maps when they are available (e.g., T + contrast) instead of the available variance maps and (2) log a warning for _any_ variance maps coming from NeuroStars, since we don't know what units they use. WDYT?

tyarkoni:
> I think it would probably be good to (1) try to use other maps to get variance maps when they are available (e.g., T + contrast) instead of the available variance maps and (2) log a warning for any variance maps coming from NeuroStars, since we don't know what units they use. WDYT?

I'm on board with (2); (1) makes me a bit nervous. It's an empirical question as to whether there are more likely to be mistakes in identification of variance maps, or in other maps we use to generate variance maps... I'm not confident the answer is the former (e.g., I worry that some users will have even more trouble determining what the right maps to use as contrasts are).

I think at minimum the user should probably take some responsibility for this... maybe we need an argument or package-level setting that indicates how maps should be detected, where the default is to assume the user has it exactly correct (unless the map is obtained from a remote source, in which case we issue a warning), and fail with an exception if required maps are missing—but then also allow the user to explicitly indicate they want maps to be automatically computed when we determine that that's possible (though I think we should still prioritize use of existing variance maps over computation from other maps).

tsalo:
>I'm on board with (2); (1) makes me a bit nervous. It's an empirical question as to whether there are more likely to be mistakes in identification of variance maps, or in other maps we use to generate variance maps... I'm not confident the answer is the former (e.g., I worry that some users will have even more trouble determining what the right maps to use as contrasts are).

TBH, my intuition is that contrast image units are clearer than variance image units, since the actual variance unit varies by package (and I believe even by tool within package- at least for AfNI), but it's also probably not worth the effort. If only modeling software was a field in the NeuroVault metadata! Then we could just hardcode in a dictionary mapping the software to the expected variance units...

>I think at minimum the user should probably take some responsibility for this... maybe we need an argument or package-level setting that indicates how maps should be detected

That sounds pretty cool, although it also sounds like it would be difficult to implement.

tyarkoni:
> That sounds pretty cool, although it also sounds like it would be difficult to implement.

Sorry, that was a bit unclear. I didn't mean we would have some kind of DSL the user can use; I was just thinking of a string that indicates which of two or three standard modes to use—i.e., to assume that the passed data are correct and raise no warnings; to always compute the variance map from other maps when possible; etc. So it would involve no additional work beyond what's already under discussion except to give the user some control over what happens.

jdkent:
Updates on this pull request:

- [I provided an example on how I would expect someone to use this function](https://4665-117724523-gh.circle-artifacts.com/0/tmp/src/NiMARE/docs/_build/html/auto_examples/01_datasets/plot_neurovault_io.html)
  - (looks like I need to fix formatting)

- I added a dictionary option for users to specify how the images are specified on neurovault
  > EDIT: we could potentially also let users pass a dictionary specifying mappings onto image types. We could even conceivably allow them to pass regexes that operate over the file names. But let's not take that on right now; I'm just mentioning it here for future reference.

I can add a check for variance maps saying they need to be above zero.

nimare/io.py
```diff
@@ -275,3 +278,123 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+        Dataset object containing experiment information from neurovault.
+    """
+    if img_dir is None:
+        img_dir = Path(tempfile.mkdtemp(prefix="neurovault_"))
+    else:
+        img_dir = Path(img_dir, exist_ok=True)
+        img_dir.mkdir(parents=True, exist_ok=True)
+
+    if map_type_conversion is None:
+        map_type_conversion = {
```
tyarkoni, line 332: Probably worth adding all of the other controlled options in NeuroVault that have an obvious mapping to NiMARE. Minimally, that's "Z map" and "p map"; not sure about others.

nimare/io.py
```diff
@@ -275,3 +278,123 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+        }
+
+    if not isinstance(collection_ids, dict):
+        collection_ids = {nv_coll: nv_coll for nv_coll in collection_ids}
+
+    dataset_dict = {}
+    for coll_name, nv_coll in collection_ids.items():
+        for contrast_name, contrast_ids in contrasts.items():
+            nv_url = f"https://neurovault.org/api/collections/{nv_coll}/images/?format=json"
+            images = requests.get(nv_url).json()
```
tyarkoni, line 345: I think this will unnecessarily retrieve the image data again for each contrast; we can move it to the first loop.

nimare/io.py
```diff
@@ -275,3 +278,123 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+
+            dataset_dict[f"study-{coll_name}"] = {
+                "contrasts": {
```
tyarkoni, line 348: See above comment—I think the default record for the study should be created outside the loop over contrasts, as it only needs to be done once. (Plus I think if there are multiple contrasts, this will overwrite all but the last. Might be worth adding a test that has multiple contrasts.)

nimare/io.py
```diff
@@ -275,3 +278,123 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+                }
+            }
+
+            for img_dict in images["results"]:
+                if not (
+                    (
+                        img_dict["name"] in contrast_ids
+                        or any(
+                            [
+                                img_dict["name"].startswith(f"{contrast}_stat")
```
tyarkoni, line 366: Not sure about the `_stat` part in the string; I think that will only work for BIDS-compliant maps. I'm not saying we should necessarily drop this, as it does seem a little bit liberal to just search for *any* image that starts with the contrast name. But maybe there's some intermediate approach that would make sense (or not—haven't thought about it).

Relatedly, is there any downside to allowing users to pass regexes as values in `contrast_ids`? That would dramatically increase flexibility (and probably obviate having to worry about the `_stat` thing above). To avoid having to add another argument that indicates whether or not to use strings as REs, we could check each value's type and only use `re.search` if it's a compiled regex instance. This is what we do in pybids and it seems to work pretty well.

nimare/io.py
```diff
@@ -275,3 +278,123 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+            "variance": "varcope",
+            "univariate-beta map": "beta",
+        }
+
+    if not isinstance(collection_ids, dict):
+        collection_ids = {nv_coll: nv_coll for nv_coll in collection_ids}
+
+    dataset_dict = {}
+    for coll_name, nv_coll in collection_ids.items():
+        for contrast_name, contrast_ids in contrasts.items():
```
tyarkoni, line 343: Might call this `contrast_matches` instead of `contrast_ids`.

nimare/io.py
tyarkoni: One other minor issue: I think checking `any()` here implies that any image in any collection that matches at least one of the `contrast_ids` values will be treated as a match. Is this what we want? I'm not saying it isn't, just that there's ambiguity right now about whether `contrast_ids` is a list of possible matches applied to every collection separately, or whether there's a one-to-one mapping from collections to entries in this list (as the example seems to suggest). I don't have a strong preference, but we should probably be consistent, and make sure it's clear in the docs.

nimare/io.py
tyarkoni: Also, we might want to move this dictionary out of here and into its own module-level constant (e.g., `DEFAULT_MAP_TYPE_CONVERSION`), as users may want to add to it and pass that rather than having to specify all values from scratch.

nimare/io.py
```diff
@@ -275,3 +278,123 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+                (
+                    dataset_dict[f"study-{coll_name}"]["contrasts"][contrast_name]["images"][
+                        map_type_conversion[img_dict["map_type"]]
+                    ]
+                ) = filename.as_posix()
+
+                # temporary sample size
+                sample_size = img_dict["number_of_subjects"]
+
+            # take the sample size from the final valid image entry (they should all be the same)
```
tyarkoni, line 391: Hmm... It's true that the sample size *should* be the same, but we can't guarantee that. They could be different if, e.g., the contrast matching procedure went awry, or if there was a data entry problem. Probably this is something we should warn the user about. Not sure what the best way to implement that is—and maybe it should be handled inside the `Dataset` initializer rather than here.

tyarkoni (changes requested):
Looks good overall, with some suggestions.

nimare/io.py
jdkent: Created the constant and added "Z map" and "p map" keys.

nimare/io.py
jdkent: d'oh, fixed this

nimare/io.py
jdkent: d'ohx2, fixed this too.

nimare/io.py
jdkent: I believed I simplified the logic by allowing (forcing?) the user to pass a regex.

nimare/io.py
jdkent: I'm calling it `contrast_regex` now since it is interpreted as a regex.

nimare/io.py
jdkent: added some logic here to make sure sample_size is the same across images, and chooses most common sample_size if there are multiple options.

nimare/io.py
```diff
@@ -275,3 +289,121 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+                        map_type_conversion[img_dict["map_type"]]
+                    ]
+                ) = filename.as_posix()
+
+                # aggregate sample sizes (should all be the same)
+                sample_sizes.append(img_dict["number_of_subjects"])
+
+            # take modal sample size (raise warning if there are multiple values)
+            if len(set(sample_sizes)) > 1:
+                sample_size = max(set(sample_sizes), key=sample_sizes.count)
```
jdkent, line 391: Use counter from collections

jdkent:
@tyarkoni, I think I've covered what we talked about for this pull request, do you or @tsalo have time to give this another quick once-over?

nimare/generate.py
```diff
@@ -105,6 +111,61 @@ def create_coordinate_dataset(
...
 
+def create_neurovault_dataset(
+    collection_ids=NEUROVAULT_IDS,
+    contrasts=CONTRAST_OF_INTEREST,
+    img_dir=None,
+    map_type_conversion=None,
+    **dset_kwargs,
+):
+    """
+    Parameters
```
tsalo, line 122: This needs a function description. Perhaps something like the following:

```suggestion
    """Download images from NeuroVault and use them to create a Dataset.

    Parameters
```

nimare/io.py
```diff
@@ -275,3 +290,130 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
         )
     dict_ = convert_sleuth_to_dict(text_file)
     return Dataset(dict_, target=target)
+
+
+def convert_neurovault_to_dataset(
+    collection_ids, contrasts, img_dir=None, map_type_conversion=None, **dset_kwargs
+):
+    """
+    Convert a group neurovault collections into a NiMARE Dataset.
```
tsalo, line 299:
```suggestion
    Convert a group of NeuroVault collections into a NiMARE Dataset.
```

nimare/io.py
```diff
@@ -275,3 +290,130 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+        more informative names in the dataset.
+    contrasts : :obj:`dict`
+        Dictionary whose keys represent the name of the contrast in
+        the dataset and whose values represent a regular expression that would
+        match the names represented in NeuroVault.
+        For example, under the ``Name`` column in this URL
+        https://neurovault.org/collections/8836/,
+        a valid contrast could be "as-Animal", which will be called "animal" in the created
+        dataset if the contrasts argument is ``{'animal': "as-Animal"}``.
+    img_dir : :obj:`str` or None
```
tsalo, line 321:
```suggestion
    img_dir : :obj:`str` or None, optional
```

nimare/io.py
```diff
@@ -275,3 +290,130 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+        Base path to save all the downloaded images, by default the images
+        will be saved to a temporary directory with the prefix "neurovault".
+    map_type_conversion : :obj:`dict` or None
```
tsalo, line 324:
```suggestion
    map_type_conversion : :obj:`dict` or None, optional
```

nimare/generate.py
```diff
@@ -105,6 +111,61 @@ def create_coordinate_dataset(
...
+    ----------
+    collection_ids : :obj:`list` of :obj:`int` or :obj:`dict`
```
tsalo, line 124:
```suggestion
    collection_ids : :obj:`list` of :obj:`int` or :obj:`dict`, optional
```

nimare/generate.py
```diff
@@ -105,6 +111,61 @@ def create_coordinate_dataset(
...
+        A list of collections on neurovault specified by their id.
+        The collection ids can accessed through the neurovault API
+        (i.e., https://neurovault.org/api/collections) or
+        their main website (i.e., https://neurovault.org/collections).
+        For example, in this URL https://neurovault.org/collections/8836/,
+        `8836` is the collection id.
+        collection_ids can also be a dictionary whose keys are the informative
+        study name and the values are collection ids to give the collections
+        human readable names in the dataset.
+    contrasts : :obj:`dict`
```
tsalo, line 134:
```suggestion
    contrasts : :obj:`dict`, optional
```

nimare/generate.py
```diff
@@ -105,6 +111,61 @@ def create_coordinate_dataset(
...
+        Dictionary whose keys represent the name of the contrast in
+        the dataset and whose values represent a regular expression that would
+        match the names represented in NeuroVault.
+        For example, under the ``Name`` column in this URL
+        https://neurovault.org/collections/8836/,
+        a valid contrast could be "as-Animal", which will be called "animal" in the created
+        dataset if the contrasts argument is ``{'animal': "as-Animal"}``.
+    img_dir : :obj:`str` or None
```
tsalo, line 142:
```suggestion
    img_dir : :obj:`str` or None, optional
```

nimare/generate.py
```diff
@@ -105,6 +111,61 @@ def create_coordinate_dataset(
...
+        Base path to save all the downloaded images, by default the images
+        will be saved to a temporary directory with the prefix "neurovault"
+    map_type_conversion : :obj:`dict` or None
```
tsalo, line 145:
```suggestion
    map_type_conversion : :obj:`dict` or None, optional
```

nimare/io.py
```diff
@@ -275,3 +290,130 @@ def convert_sleuth_to_dataset(text_file, target="ale_2mm"):
...
+        Keyword arguments to pass in when creating the Dataset object.
+        see :obj:`nimare.dataset.Dataset` for details.
+
+    Returns
+    -------
+    :obj:`nimare.dataset.Dataset`
+        Dataset object containing experiment information from neurovault.
+    """
+    if img_dir is None:
+        img_dir = Path(tempfile.mkdtemp(prefix="neurovault_"))
```
tsalo, line 338: I've tried to set up a NiMARE data folder that operates in the same way nilearn's data folder does. Should we use that instead?

tsalo (changes requested):
Some minor docstring changes, a question about where to store images, and one final question:

This should fail on NIDM-Results-format NeuroVault collections like the 21 pain study one we use elsewhere, right?

nimare/io.py
jdkent: oh yeah, I can use that as the default.

jdkent:
> This should fail on NIDM-Results-format NeuroVault collections like the 21 pain study one we use elsewhere, right?

In principle, I think this should be supported, but for right now I'm assuming each neurovault collection represents a single study as that covers the current use case for performing meta-analyses on the output of neuroscout.

examples/01_datasets/plot_neurovault_io.py
```diff
@@ -0,0 +1,94 @@
...
+# NiMARE can read these statistical maps when given a list of collection_ids.
+# I search "working memory" on neurovault, and find these relevant collections:
+#
+# * `2884 <https://neurovault.org/collections/2884/>`_,
+# * `2621 <https://neurovault.org/collections/2621/>`_,
+# * `3085 <https://neurovault.org/collections/3085/>`_,
+# * `5623 <https://neurovault.org/collections/5623/>`_,
+# * `3264 <https://neurovault.org/collections/3264/>`_,
+# * `3192 <https://neurovault.org/collections/3192/>`_,
+# * `457 <https://neurovault.org/collections/457/>`_.
```
tsalo, line 24: The example looks great overall, but I don't love how this part renders...

image

nimare/generate.py
```diff
@@ -105,6 +111,61 @@ def create_coordinate_dataset(
...
+    -------
+    :obj:`nimare.dataset.Dataset`
+        Dataset object containing experiment information from neurovault.
+    """
+    dataset = convert_neurovault_to_dataset(
+        collection_ids, contrasts, img_dir, map_type_conversion, **dset_kwargs
+    )
+    dataset.images = transform_images(
+        dataset.images, target="z", masker=dataset.masker, metadata_df=dataset.metadata
+    )
```
tsalo, line 164: Any particular reason only z images are targeted here?

tsalo (commented):
Everything is look great. I have one minor request for the example and one question for the main dataset creation function.

examples/01_datasets/plot_neurovault_io.py
jdkent: to be clear, it's the commas/period that's off right? or is there something else as well?

tsalo: I think dropping the commas and period would be good, or keeping them but putting the items on a single line like a sentence. Either way would look good.

nimare/generate.py
jdkent: z images were targeted because that was the only image type I could reasonably reconstruct from what I downloaded (I only downloaded T images or Z images). Is there another (additional) target that is preferred?

tsalo: That makes sense. I wasn't sure if it would download everything available given the data type dictionary I saw. If the target is generally going to be z images, can you just add a note to the docstring to that effect?

jdkent: Sorry, I don't think I know what you are suggesting to add (to the `convert_neurovault_to_dataset` function docstring?)

The downloader filters based on name of the contrast (e.g., "Working memory load of 2 faces versus 1 face - NT2_Tstat")  and whether that contrast is represented in the `map_type_conversion` dictionary.
If you want to download beta maps and varcopes instead, that is also a valid use case (and closer to what I'm doing for refactoring the meta-analysis for the neuroscout paper.

> If the target is generally going to be z images, can you just add a note to the docstring to that effect?

I'm interpreting this as suggesting `convert_neurovault_to_dataset` will primarily download t or z images, which should not be true since any matching image specified in  `map_type_conversion` should be downloaded, it just so happens with this collection of studies the easiest path forward was to try to download both t and z maps (but each study only has either a z or t map) and convert the studies with t images to z for this analysis.

nimare/generate.py
```diff
@@ -105,6 +111,61 @@ def create_coordinate_dataset(
...
+def create_neurovault_dataset(
+    collection_ids=NEUROVAULT_IDS,
+    contrasts=CONTRAST_OF_INTEREST,
+    img_dir=None,
+    map_type_conversion=None,
+    **dset_kwargs,
+):
+    """Download images from NeuroVault and use them to create a dataset.
+
+    Parameters
```
tsalo, line 123:
```suggestion

    This function will also attempt to generate Z images for any contrasts
    for which this is possible.

    Parameters
```

Maybe something like this?

nimare/generate.py
tsalo: I was just thinking of something like https://github.com/neurostuff/NiMARE/pull/432/files#r588523321. Just some indication about the additional step being taken in this function.

nimare/generate.py
jdkent: oh, oops, sorry, for the function `create_neurovault_dataset` (not `convert_neurovault_to_dataset`) that makes sense, yes.

tsalo (approved):
Thanks! LGTM!

Issue #418, opened 2020-12-10
Example breaking docs build


tsalo:
Here's the traceback from [here](https://app.circleci.com/pipelines/github/neurostuff/NiMARE/956/workflows/065f575e-1037-49e8-9e09-2af981477db5/jobs/4181):

```
Extension error:
Here is a summary of the problems encountered when running the examples

Unexpected failing examples:
/tmp/src/NiMARE/examples/02_meta-analyses/plot_simulated_data.py failed leaving traceback:
Traceback (most recent call last):
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/sphinx_gallery/scrapers.py", line 335, in save_figures
    rst = scraper(block, block_vars, gallery_conf)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/sphinx_gallery/scrapers.py", line 151, in matplotlib_scraper
    fig.savefig(image_path, **these_kwargs)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/matplotlib/figure.py", line 2311, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/matplotlib/backend_bases.py", line 2217, in print_figure
    **kwargs)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/matplotlib/backend_bases.py", line 1639, in wrapper
    return func(*args, **kwargs)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/matplotlib/backends/backend_agg.py", line 509, in print_png
    FigureCanvasAgg.draw(self)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/matplotlib/backends/backend_agg.py", line 407, in draw
    self.figure.draw(self.renderer)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/matplotlib/artist.py", line 41, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/matplotlib/figure.py", line 1837, in draw
    pos = locator(ax, renderer)
  File "/opt/conda/envs/py36_env/lib/python3.6/site-packages/nilearn/plotting/displays.py", line 1662, in _locator
    return transforms.Bbox([[left_dict[axes], y0],
[...]

PR #439, opened 2021-02-04
[FIX,REF] start changing how to handle resampling

### LINKED ISSUE #438: IBMA estimators resampling input (potentially non-optimally)
## Summary

when preprocessing the input for IBMA, [the input images are resampled to match the masker object (through the transform)](https://github.com/neurostuff/NiMARE/blob/106b51634707fd7999617f277378aefc349867f8/nimare/base.py#L293).
nilearn's resampling function default options may not be optimal for meta-analytic maps.

For example, some code from @tyarkoni:
`resample_to_img(nb.load(img), template, clip=True, interpolation='linear')`

deviates from the defaults `interpolation='continuous'` and `clip=False`.
Changing those defaults appears to give the maps being analyzed a more biologically plausible result.

## Additional details

- NiMARE version: dev

add nimare to the neuroscout paper

biologically plausible results conformant to previous analyses with PyMare

### What actually happened?
weird pixelated results

jdkent:
Closes #438

- check if resampling is necessary (and raise error if resampling is not requested)
- add resampling kwargs for user flexibility
- TODO: decide where to pass resampling arguments (should they be passed through the `fit` method or should we make it a part of initializing the estimator object so the user's decision is saved in a more permanent way? and to keep the `fit` method clean)

### DIFF
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -289,9 +289,10 @@ class MetaEstimator(Estimator):
         self.resample = kwargs.get("resample", False)
 
         # defaults for resampling images (nilearn's defaults do not work well)
-        self.resample_kwargs = {
-            k.split("resample__")[1]: v for k, v in kwargs.items() if k.startswith("resample__")
-        } or {"clip": True, "interpolation": "linear"}
+        self._resample_kwargs = {"clip": True, "interpolation": "linear"}
+        self._resample_kwargs.update(
+            {k.split("resample__")[1]: v for k, v in kwargs.items() if k.startswith("resample__")}
+        )
 
     def _preprocess_input(self, dataset):
         """Preprocess inputs to the Estimator from the Dataset as needed."""
@@ -306,17 +307,14 @@ class MetaEstimator(Estimator):
                 # If no resampling is requested, check if resampling is required
                 if not self.resample:
                     check_imgs = {img: nb.load(img) for img in self.inputs_[name]}
-                    check_imgs["reference_masker"] = mask_img
-                    _check_same_fov(**check_imgs, raise_error=True)
-                    # remove the reference image
-                    del check_imgs["reference_masker"]
+                    _check_same_fov(**check_imgs, reference_masker=mask_img, raise_error=True)
                     imgs = list(check_imgs.values())
                 else:
                     # resampling will only occur if shape/affines are different
                     # making this harmless if all img shapes/affines are the same
                     # as the reference
                     imgs = [
-                        resample_to_img(nb.load(img), mask_img, **self.resample_kwargs)
+                        resample_to_img(nb.load(img), mask_img, **self._resample_kwargs)
                         for img in self.inputs_[name]
                     ]
 
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -290,9 +290,9 @@ class MetaEstimator(Estimator):
         self.resample = kwargs.get("resample", False)
 
         # defaults for resampling images (nilearn's defaults do not work well)
-        self.resample_kwargs = kwargs.get(
-            "resample_kwargs", {"clip": True, "interpolation": "linear"}
-        )
+        self.resample_kwargs = {
+            k.split("resample__")[1]: v for k, v in kwargs.items() if k.startswith("resample__")
+        } or {"clip": True, "interpolation": "linear"}
 
     def _preprocess_input(self, dataset):
         """Preprocess inputs to the Estimator from the Dataset as needed."""
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -297,21 +297,29 @@ class MetaEstimator(Estimator):
     def _preprocess_input(self, dataset):
         """Preprocess inputs to the Estimator from the Dataset as needed."""
         masker = self.masker or dataset.masker
+
+        mask_img = masker.mask_img or masker.labels_img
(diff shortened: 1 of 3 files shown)


nimare/base.py
```diff
@@ -283,14 +287,35 @@ def __init__(self, *args, **kwargs):
...
+                # resample_imgs = True
+                # If no resampling is requested, check if resampling is required
+                if not resample:
+                    check_imgs = {img: nb.load(img) for img in self.inputs_[name]}
+                    check_imgs['reference_masker'] = masker.mask_img
+                    _check_same_fov(**check_imgs, raise_error=True)
+
+                # defaults for resampling images (nilearn's defaults do not work well)
+                if not resample_kwargs:
+                    resample_kwargs = {'clip': True, interpolation: 'linear'}
```
tsalo, line 307: `interpolation` just needs to be a string, I think.

nimare/base.py
```diff
@@ -283,14 +287,35 @@ def __init__(self, *args, **kwargs):
...
+                if not self.resample:
+                    check_imgs = {img: nb.load(img) for img in self.inputs_[name]}
+                    check_imgs["reference_masker"] = masker.mask_img
+                    _check_same_fov(**check_imgs, raise_error=True)
+
+                # resampling will only occur if shape/affines are different
+                # making this harmless if all img shapes/affines are the same
+                # as the reference
+                imgs = [
+                    resample_to_img(nb.load(img), masker.mask_img, **resample_kwargs)
```
tsalo, line 312:
```suggestion
                    resample_to_img(nb.load(img), masker.mask_img, **self.resample_kwargs)
```

nimare/base.py
```diff
@@ -283,14 +287,35 @@ def __init__(self, *args, **kwargs):
             mask = get_masker(mask)
         self.masker = mask
 
+        self.resample = kwargs.get("resample", False)
+
+        # defaults for resampling images (nilearn's defaults do not work well)
+        self.resample_kwargs = kwargs.get(
+            "resample_kwargs", {"clip": True, "interpolation": "linear"}
+        )
```
tsalo, line 295: In this approach, we're talking about feeding in a single argument, `resample_kwargs`, as a dictionary. In order, similar situations in NiMARE, we instead accept individual arguments with a prefix instead of bundling them in a dictionary. See how we work with kernel transformer arguments fed into meta-analytic estimators:

https://github.com/neurostuff/NiMARE/blob/106b51634707fd7999617f277378aefc349867f8/nimare/meta/cbma/base.py#L41-L44

If you want to go with the dictionary-based approach, I'd like to propagate it across the package.

tsalo (changes requested):
One minor change that will hopefully fix the tests, and one other thing I think we should discuss before merging.

nimare/base.py
jdkent: I do not have strong feelings about the approach, I'll switch to prefixes for consistency.

nimare/base.py
```diff
@@ -9,6 +9,10 @@
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
 
+from nilearn.image import resample_to_img, concat_imgs
+from nilearn._utils.niimg_conversions import _check_same_fov
+
+import nibabel as nb
```
tsalo, line 15: Can you run `isort`?

tsalo:
The docs failure seems to be the random bug documented in #418. I still don't have a solution, so I think we can just restart the CI and it might go away.

nimare/base.py
```diff
@@ -283,14 +286,46 @@ def __init__(self, *args, **kwargs):
...
+        if isinstance(mask_img, str):
+            mask_img = nb.load(mask_img)
+
         for name, (type_, _) in self._required_inputs.items():
             if type_ == "image":
+                # If no resampling is requested, check if resampling is required
+                if not self.resample:
+                    check_imgs = {img: nb.load(img) for img in self.inputs_[name]}
+                    check_imgs["reference_masker"] = mask_img
+                    _check_same_fov(**check_imgs, raise_error=True)
```
tyarkoni, line 310: It's pretty trivial, and would only slightly improve readability, but to save the `del` call later you could just pass `reference_masker` as an argument directly to `_check_same_fov` instead of adding it to `check_imgs`, since you remove it right away

nimare/base.py
```diff
@@ -283,14 +286,46 @@ def __init__(self, *args, **kwargs):
             mask = get_masker(mask)
         self.masker = mask
 
+        self.resample = kwargs.get("resample", False)
+
+        # defaults for resampling images (nilearn's defaults do not work well)
+        self.resample_kwargs = {
```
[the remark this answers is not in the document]
jdkent: good catch! I do not think I want to expose this to users, and yes you are right that appears to be the convention, changed!

nimare/base.py
```diff
@@ -283,14 +286,46 @@ def __init__(self, *args, **kwargs):
...
+            k.split("resample__")[1]: v for k, v in kwargs.items() if k.startswith("resample__")
```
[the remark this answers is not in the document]
jdkent: great point, changed it so there is a default dict and arguments can update that dict.

nimare/base.py
jdkent: clean code is happy code :), changed.

Issue #423, opened 2020-12-16
threshold z-map images from Dataset objects to create coordinates


jdkent:
It would be useful to create coordinates from an image within a Dataset object in circumstances where:
- for some of the studies you only have coordinates, but others you have images
- want to compare an IBMA approach with a CBMA approach when you have images.
This can be accomplished with nilearns [get_clusters_table](https://nilearn.github.io/modules/generated/nilearn.reporting.get_clusters_table.html)

tsalo:
I think this is a great idea, but I have a couple of thoughts:

1. Would applying the same threshold to all image-only studies introduce a bias? 
2. It seems like this would make handling null findings in CBMAs more relevant since there's no guarantee that the images will have any significant clusters.
3. It should be trivial to convert other images types to z-maps before the coordinate extraction too!

tyarkoni:
> Would applying the same threshold to all image-only studies introduce a bias?

Yes in the sense that the distribution of voxel-wise image statistics will be biased (e.g., if you use a low threshold, you'll get more 1's in the MKDA maps than if you use a high threshold). But that's only a problem in the usual "gives people one more knob with which to p-hack" sense. There's no obvious reason why it should bias meta-analysis statistics, because picking a different threshold shouldn't increase the probability of spatial convergence. (Well, except maybe in the sense that at the limit, the null is obviously false everywhere, so with enough non-zero values, the whole brain becomes significant. But that's a separate and much deeper conceptual issue.)

> It seems like this would make handling null findings in CBMAs more relevant since there's no guarantee that the images will have any significant clusters.

Probably a dumb question, but: why are null findings a problem? Can't we just run empty maps through the existing procedures and then they just add a bit of uncertainty? Or is the issue just in the representation of the data—i.e., that we can't tell whether a study is missing coordinates because of null findings, or because of missing data? If it's the latter, maybe we can just adopt a convention that `[]` is different from `null` (or `None` in Python).

> It should be trivial to convert other images types to z-maps before the coordinate extraction too!

+1

tsalo:
> There's no obvious reason why it should bias meta-analysis statistics, because picking a different threshold shouldn't increase the probability of spatial convergence.

That's a relief. Thanks!

tyarkoni:
[...]

PR #445, opened 2021-02-17
[FIX] transform_images extra dimension

### LINKED ISSUE #444: transform_images adds extra dimension to data object
## Summary

Statistic maps should be 3D, but when using transform_images to generate a missing map, the outputs are 4D

## Additional details

- NiMARE version: 0.0.5

create z maps

3D output

### What actually happened?
4D output

## Reproducing the bug
run this test:
https://github.com/neurostuff/NiMARE/blob/43d20273f859e3d397547ec2e30c8de9f6fde4fb/nimare/tests/test_transforms.py#L17

and check `nib.load(new_z_files[-1]).shape`, it will say `(10, 10, 10, 1)`, but I want `(10, 10, 10)`.

jdkent:
Closes #444  .

- `squeeze()` output to remove extra dimension 
- add check to ensure correct dimensionality of output

### DIFF
--- a/nimare/tests/test_transforms.py
+++ b/nimare/tests/test_transforms.py
@@ -2,6 +2,7 @@
 Test nimare.transforms
 """
 import numpy as np
+import nibabel as nib
 
 from nimare import transforms, utils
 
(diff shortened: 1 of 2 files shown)


jdkent:
I think this is ready for review (if tests pass).

This pull request is necessary for me to finish working on #423, since [nilearn's get_cluster_table](https://nilearn.github.io/modules/generated/nilearn.reporting.get_clusters_table.html) assumes the statistical map is 3D.

tsalo (approved):
The changes look good to me. Is this ready to merge?

jdkent:
I think so, I have nilearn/nilearn#2701 open so this will not be a problem in the future, but it will probably be a while before that gets merged and nilearn cuts a new release, so merging this now is probably best.

PR #447, opened 2021-02-24
[FIX] Permutation p-values

### LINKED ISSUE #442: Unexpected clusters with high z scores in ALE subtraction analysis
## Summary
Whenever I run an ALE subtraction analysis using `meta.cbma.ALESubtraction()` with NiMARE v0.0.5, the resulting contrast map shows some unexpected clusters with very high _z_ scores (see output below). The size and shape of these clusters (with their very rough edges) suggests to me that these _z_ scores have not been estimated correctly.

## Additional details
- NiMARE version: 0.0.5

Running an ALE subtraction analysis between two data sets (i.e., two Sleuth files with non-overlapping sub-groups of experiments from a larger set of experiments).

To get a subtraction map with _z_ scores showing the difference between `dset1` and `dset2`. Performing this analysis in GingerALE or NiMARE v0.0.3 shows some clusters with _z_ scores of up to ~4, in plausible locations and with a smooth shape at the edges.

### What actually happened?

Performing the same analysis in NiMARE v0.0.5 produces very similar results throughout most of the brain, but there are also some weird-looking additional clusters, all with a rough-edged shape and at implausible locations (e.g., in the cerebellum). Note that the same happens with different input data sets, with `low_memory=True`, and with higher a number of iterations (e.g., `n_iters=50000`).

### Input
[older.txt](https://github.com/neurostuff/NiMARE/files/5975454/older.txt)
[younger.txt](https://github.com/neurostuff/NiMARE/files/5975455/younger.txt)

### Code

```python
# Read Sleuth files
from nimare import io, meta
dset1 = io.convert_sleuth_to_dataset(text_file='older.txt')
dset2 = io.convert_sleuth_to_dataset(text_file='younger.txt')

# Perform subtraction analysis
sub = meta.cbma.ALESubtraction(n_iters=10000, low_memory=False)
sres = sub.fit(dset1, dset2)

# Save + plot
from nibabel import save
from nilearn import plotting
img_z = sres.get_map('z_desc-group1MinusGroup2')
save(img_z, filename='older_minus_younger.nii.gz')
p = plotting.plot_glass_brain('older_minus_younger.nii.gz', colorbar=True)
p.savefig('older_minus_younger.png')
```

### Output
[older_minus_younger.nii.gz](https://github.com/neurostuff/NiMARE/files/5976276/older_minus_younger.nii.gz)
older_minus_younger

tyarkoni:
Fixes the issue identified in #442, which turned out to stem from a bug in `stats.null_to_p`. I replaced the previous call to `scipy.stats.percentileofscore` with an alternative approach that relies on `np.searchsorted`; this approach has the benefit of being vectorized, so it should be much faster in cases where the null distribution is IID across voxels (though it won't help, and might even slow things down slightly, in cases where we need to loop in order to generate a different null distribution for each voxel).

I also updated/improved the tests to make sure they cover common use cases.

Closes #442.

### DIFF
--- a/nimare/stats.py
+++ b/nimare/stats.py
@@ -3,7 +3,6 @@ import logging
 import warnings
 
 import numpy as np
-from scipy import stats
 
 from . import utils
 
@@ -149,7 +148,7 @@ def null_to_p(test_value, null_array, tail="two", symmetric=False):
     approximately twice as efficient computationally, and has lower variance.
     """
 
-    if tail not in {'two', 'upper', 'lower'}:
+    if tail not in {"two", "upper", "lower"}:
         raise ValueError('Argument "tail" must be one of ["two", "upper", "lower"]')
 
     return_first = isinstance(test_value, (float, int))
@@ -166,7 +165,7 @@ def null_to_p(test_value, null_array, tail="two", symmetric=False):
 
     def compute_p(t, null):
         null = np.sort(null)
-        idx = np.searchsorted(null, t, side='left').astype(float)
+        idx = np.searchsorted(null, t, side="left").astype(float)
         return 1 - idx / len(null)
 
     if tail == "two":
--- a/nimare/stats.py
+++ b/nimare/stats.py
@@ -110,7 +110,7 @@ def pearson(x, y):
     return rs
 
 
-def null_to_p(test_value, null_array, tail="two"):
+def null_to_p(test_value, null_array, tail="two", symmetric=False):
     """Return p-value for test value(s) against null array.
 
     Parameters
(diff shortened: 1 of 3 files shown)


nimare/stats.py
```diff
@@ -147,26 +153,26 @@ def null_to_p(test_value, null_array, tail="two"):
...
-    p = np.array([stats.percentileofscore(null_array, v, "strict") for v in test_value])
-    p /= 100.0
     if tail == "two":
-        p = (0.5 - np.abs(p - 0.5)) * 2
-    elif tail == "upper":
-        p = 1 - p
-    elif tail != "lower":
-        raise ValueError('Argument "tail" must be one of ["two", "upper", "lower"]')
+        test_value = np.abs(test_value)
+        null_array = np.abs(null_array)
```
tsalo, line 158: Does this mean that `tail = "two"` will be the same as `tail = "upper"` when the null array and test values are all positive?

tyarkoni: Oh, good point—this will break for asymmetric (or non-centered) distributions. I was focusing on the ALE subtraction case, where the null distribution will be symmetric, but this won't be true for a standard ALE analysis.

The downside of reverting to our previous approach (double the minimum of the smaller one-tailed p-value) is that it's conservative in the case of a symmetric distribution, plus it requires doing two one-sided tests. (We weren't actually doing two one-tailed tests before, which was a bug, though in practice the difference would be negligible in nearly all cases.)

My proposal in light of the above is to revert to the previous procedure as the default for `tail='two'`, but also add a `symmetric=False` argument that, when set to `True`, will use the above approach (taking the absolute value), which is a little more powerful. Then we can use the latter whenever we know the null distribution is symmetric (e.g., for subtractions). Sound good?

tsalo: That sounds good to me. One (naive) question though- could we just center the array and adjust the test value accordingly, then just use your approach?

Something like:
```python
if tail == "two":
    # mean center
    m = null_array.mean()
    test_value -= m
    null_array -= m
    # take absolute values
    test_value = np.abs(test_value)
    null_array = np.abs(null_array)
```

tyarkoni: > could we just center the array and adjust the test value accordingly, then just use your approach?

If we knew for a fact that the distribution was symmetric, I think we could do that. But we don't know that, and it seems very unlikely to be true for, e.g., ALE statistics, so I think we should probably keep it this way.

tsalo: Okay sounds good then. Thanks!

tyarkoni:
@tsalo okay from my perspective everything looks good to merge once tests pass (the last commit just fixes the style issues; all the other tests are already passing).

tsalo (approved):
Everything looks good to me. Merging!

PR #455, opened 2021-03-09
[FIX] Add work-around for maskers that do not accept 1D input

### LINKED ISSUE #454: Squeezing results data before unmasking breaks NiftiLabelsMaskers
## Summary

As of #450, we have started squeezing PyMARE outputs before unmasking them to images. Unfortunately, `NiftiLabelsMaskers` require 2D inputs to the `inverse_transform` method, and our tests did not cover the actual unmasking.

## Additional details

- NiMARE version: 0.0.7

Run an IBMA with a NiftiLabelsMasker.

The IBMA should complete without an error.

### What actually happened?
There is an IndexError:

```python 
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

## Reproducing the bug

Add the following line to [`test_ibma_with_custom_masker`](https://github.com/neurostuff/NiMARE/blob/75b5539e21d9e260b5fcb531613e4dc02df2d417/nimare/tests/test_meta_ibma.py#L112):

```python
assert meta.results.get_map("z").shape == (10, 10, 10)
```

jdkent:
Closes #454.

upstream fix nilearn/nilearn#2725

- treat input as 2D for maskers that do not support 1D inputs and squeeze the result
- ensure the dimensions are correct in test

### DIFF
--- a/nimare/results.py
+++ b/nimare/results.py
@@ -3,6 +3,8 @@ import copy
 import logging
 import os
 
+from nibabel.funcs import squeeze_image
+
 from .utils import get_masker
 
 LGR = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


tsalo (approved):
Thanks for handling this and for opening nilearn/nilearn#2724!

I was looking for an image squeeze function yesterday, but I only looked in nilearn. Didn't even occur to me to check nibabel.

Everything looks solid to me- my only request is that you add the `get_map` test to the parameterized IBMA test as well. Just so we can be sure that NiftiMasker stays solid as well.

tsalo (changes requested):
Failed to bundle my comment under a review... Sorry about that!

tsalo (approved):
Everything is good. My requested changes were... not good.

tsalo:
I'm not sure why CodeCov hasn't run, but the real tests are passing, so I think we can merge. @jdkent are you comfortable with me merging this now?

PR #460, opened 2021-03-12
[REF] Write temporary files to the NiMARE data directory

tsalo:
Closes None. In working on the NiMARE paper, I've had trouble with memory issues. At minimum, writing temporary files (created when the `low_memory` option is used) to the NiMARE data directory will make it easier for users to track the files and remove them as necessary.

- Write temporary files created in `ALEKernel`, `MKDAKernel`, `KDAKernel`, and `ALESubtraction` to the NiMARE data directory instead of temporary directories.
- Log the temporary filenames. I'd love to log their sizes as well, but I'm not sure how to estimate the file size before it's created. It should just be a matter of the data type and the size though...

### DIFF
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -1,7 +1,6 @@
 """CBMA methods from the ALE and MKDA families."""
 import logging
 import multiprocessing as mp
-import os
 import inspect
 
 import nibabel as nib
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -79,6 +79,8 @@ class CBMAEstimator(MetaEstimator):
         ma_values = self.kernel_transformer.transform(
             self.inputs_["coordinates"], masker=self.masker, return_type="array"
         )
+        if isinstance(ma_values, np.memmap):
+            raise Exception("Output of transformer is a memmap!")
 
         self.weight_vec_ = self._compute_weights(ma_values)
 
@@ -94,9 +96,6 @@ class CBMAEstimator(MetaEstimator):
 
         else:
             self._compute_null_reduced_empirical(ma_values, n_iters=self.n_iters)
-
-        if self.kernel_transformer.low_memory:
-            os.remove(ma_values.filename)
 
         p_values, z_values = self._summarystat_to_p(stat_values, null_method=self.null_method)
 
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -1,6 +1,7 @@
 """CBMA methods from the ALE and MKDA families."""
 import logging
 import multiprocessing as mp
+import os
 import inspect
 
 import nibabel as nib
@@ -93,6 +94,9 @@ class CBMAEstimator(MetaEstimator):
 
         else:
             self._compute_null_reduced_empirical(ma_values, n_iters=self.n_iters)
+
+        if self.kernel_transformer.low_memory:
+            os.remove(ma_values.filename)
 
         p_values, z_values = self._summarystat_to_p(stat_values, null_method=self.null_method)
 
--- a/nimare/meta/kernel.py
+++ b/nimare/meta/kernel.py
@@ -15,6 +15,7 @@ import numpy as np
 import pandas as pd
 from nilearn import image
 
+from ..utils import use_memmap
 from ..base import Transformer
 from ..transforms import vox2mm
 from .utils import compute_ale_ma, compute_kda_ma, compute_p2m_ma, get_ale_kernel
@@ -70,6 +71,7 @@ class KernelTransformer(Transformer):
         )
         self.image_type = "{ps}_{n}".format(n=self.__class__.__name__, ps=param_str)
 
+    @use_memmap(LGR)
     def transform(self, dataset, masker=None, return_type="image"):
         """Generate modeled activation images for each Contrast in dataset.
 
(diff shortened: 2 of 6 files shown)


tsalo:
@tyarkoni @jdkent do you have any idea what could cause a bus error based on the proposed changes?

jdkent:
Don't totally understand the bus error, but `mkdtemp` must be doing some important stuff behind the scenes.
You can still use `mkdtemp` and specify where the temporary directory should be created (e.g., `mkdtemp(dir='/my/special/directory')`). If you want to ensure no file collisions, you can look into [mkstemp](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp) instead, and those files can go into the nimare dataset directory.

Going back to the overarching problem, do you want to keep track of these files? or do you just want them to be deleted/cleaned up in a more timely manner?

I guess I don't know how useful those files are for debugging.

tsalo:
>You can still use mkdtemp and specify where the temporary directory should be created (e.g., mkdtemp(dir='/my/special/directory')).

I didn't know that! I'll try using tempfile again.

>Going back to the overarching problem, do you want to keep track of these files? or do you just want them to be deleted/cleaned up in a more timely manner?

I don't want to keep them around long-term, but I do want to have them someplace accessible in cases where the script may crash or otherwise fail to delete them.

tsalo:
>You can still use mkdtemp and specify where the temporary directory should be created (e.g., mkdtemp(dir='/my/special/directory')).

I didn't know that! I'll try using tempfile again.

>Going back to the overarching problem, do you want to keep track of these files? or do you just want them to be deleted/cleaned up in a more timely manner?

I don't want to keep them around long-term, but I do want to have them someplace accessible in cases where the script may crash or otherwise fail to delete them.

tsalo:
>You can still use mkdtemp and specify where the temporary directory should be created (e.g., mkdtemp(dir='/my/special/directory')).

I didn't know that! I'll try using tempfile again.

>Going back to the overarching problem, do you want to keep track of these files? or do you just want them to be deleted/cleaned up in a more timely manner?

I don't want to keep them around long-term, but I do want to have them someplace accessible in cases where the script may crash or otherwise fail to delete them.

tsalo:
>You can still use mkdtemp and specify where the temporary directory should be created (e.g., mkdtemp(dir='/my/special/directory')).

I didn't know that! I'll try using tempfile again.

>Going back to the overarching problem, do you want to keep track of these files? or do you just want them to be deleted/cleaned up in a more timely manner?

I don't want to keep them around long-term, but I do want to have them someplace accessible in cases where the script may crash or otherwise fail to delete them.

tsalo:
One thing that's really weird is that I get bus errors from this step on my HPC using the old code, but not using the new code. Hopefully solving the problem for the CI won't break things for the actual analyses I'm running.

nimare/meta/cbma/base.py
```diff
@@ -94,6 +95,9 @@ def _fit(self, dataset):
         else:
             self._compute_null_reduced_empirical(ma_values, n_iters=self.n_iters)
 
+        if self.kernel_transformer.low_memory:
+            os.remove(ma_values.filename)
```
tsalo, line 99: @jdkent WDYT of this for cleanup?

tsalo: It turns out that was a bad idea, and the kernel transformers can handle the memmap deletion internally.

tsalo:
Everything seems to be working (🤞), but the current cleanup removes the mem-mapped files, but leaves the empty temporary directories.

jdkent (changes requested):
It looks like the code additions could fit nicely into two functions: `_create_memmap` and `_delete_memmap`

```
def _create_memmap(prefix, dtype, shape, logger, dirname="temporary_files"):
        start_time = datetime.datetime.now().strftime("%Y%m%dT%H%M%S")
        dataset_dir = _get_dataset_dir("temporary_files", data_dir=None)
        temp_dir = mkdtemp(prefix=prefix, dir=dataset_dir)
        filename = os.path.join(temp_dir, f"{prefix}_{start_time}.dat")
        logger.info(f"Temporary file written to {filename}")

        # Use a memmapped array
        return np.memmap(
            filename, dtype=dtype, mode="w+", shape=shape
        )
```

```
def _delete_memmap(arr, logger):
        logger.info(f"Removing {arr.filename}")
            os.remove(arr.filename)
            temp_dir = os.path.dirname(arr.filename)
            if len(os.listdir(temp_dir)) == 0:
                logger.info(f"Removing empty directory {temp_dir}")
                os.rmdir(temp_dir)
      return arr.filename # don't know if this should return anything...
```

This should help with readability and maintainability. (these functions will probably go in utils? (that's a cop-out)

jdkent:
methods where memmap is being used:

- ALESubtraction
  - created: _fit
  - deleted: _fit
- SCALE
  - created: _fit
  - deleted: _fit
- KernelTransformer
  - created: _transform
  - deleted: transform
- meta.utils
  -  created: compute_kda_ma (called in KDAKernel._transform)

I think I write a decorator for this, I'll open a pull request to your branch.
A decorator will help with being DRY (Don't Repeat Yourself), but it will look a little weird/black magicy.

nimare/meta/kernel.py
```diff
@@ -159,7 +161,8 @@ def transform(self, dataset, masker=None, return_type="image"):
 
         if not isinstance(transformed_maps[0], (list, tuple)):
             if return_type == "array":
-                return transformed_maps[0][:, mask_data]
+                masked_ma_arr = transformed_maps[0][:, mask_data]
+                return masked_ma_arr
```
jdkent, line 165:
```suggestion
                return transformed_maps[0][:, mask_data]
```

This looks like it can be reverted to what it was before

nimare/utils.py
```diff
@@ -386,3 +390,36 @@ def uk_to_us(text):
         pattern = re.compile(r"\b(" + "|".join(SPELL_DICT.keys()) + r")\b")
         text = pattern.sub(lambda x: SPELL_DICT[x.group()], text)
     return text
+
+
+def use_memmap(logger):
```
jdkent, line 395: Might want a unit test for `use_memmap`, but I can do that in a separate pull request if you do not want to that here.

jdkent (commented):
Just a minor change and perhaps a unit test for `use_memmap`, but I can take on in another pull request, otherwise, LGTM!

jdkent (approved):

tsalo:
@jdkent Thanks for all your help with this!

PR #462, opened 2021-03-12
[REF] Use saved MA maps, when available, in CBMA estimators

### LINKED ISSUE #461: CBMA estimators do not leverage pre-generated MA maps
## Summary

CBMA estimators do not take advantage of pre-generated MA maps within a Dataset.

## Additional details

Run the CorrelationDecoder for the NiMARE paper 😭

When the CorrelationDecoder is fitted, it precomputes MA maps and saves them to files so that the massive number of meta-analyses that will be performed can skip the kernel transformation step. Therefore, I expected the MetaEstimator to use the precomputed MA maps instead of generating them from scratch.

### What actually happened?
The MetaEstimator re-generated the MA maps from scratch.

## Where is the problem?

The problem comes down to the following lines:

https://github.com/neurostuff/NiMARE/blob/955d7d909b8bf7a41cca311d80c715230aad804b/nimare/meta/cbma/base.py#L78-L80

https://github.com/neurostuff/NiMARE/blob/955d7d909b8bf7a41cca311d80c715230aad804b/nimare/meta/cbma/base.py#L648-L652

These use the `inputs_` attribute for the Estimator, which is set in:

https://github.com/neurostuff/NiMARE/blob/955d7d909b8bf7a41cca311d80c715230aad804b/nimare/base.py#L323-L324

tsalo:
Closes #461.

- Add a new `inputs_` field called `ma_maps` in which MA maps stored in the Dataset's images attribute are collected, should they exist.

### DIFF
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -6,6 +6,7 @@ import multiprocessing as mp
 import pickle
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
+from hashlib import md5
 
 import nibabel as nb
 import numpy as np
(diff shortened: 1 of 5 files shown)


tsalo:
My only concern with this approach is that it doesn't leverage the re-use code in KernelTransformer at all.

Here's the KernelTransformer reuse code:
https://github.com/neurostuff/NiMARE/blob/955d7d909b8bf7a41cca311d80c715230aad804b/nimare/meta/kernel.py#L126-L138

PR #464, opened 2021-03-15
[FIX] Assume non-symmetric null distribution in ALESubtraction

tsalo:
Closes None, but addresses an issue reported by @mriedel56 offline.

- Change `symmetric=True` to `symmetric=False`.
- Add comment about use of `symmetric=False` to code.
- Output p-value map for ALESubtraction in addition to z-value map.


tsalo:
The before and after, courtesy of @mriedel56:

image

tsalo:
@alexenge I want to let you know about this one given that you opened #442.

alexenge:
Thanks a lot for this change and for bringing it to my attention! Seems very reasonable.

PR #472, opened 2021-03-19
[FIX] Find common stem in find_stem instead of largest common substring

### LINKED ISSUE #471: Shared path detection in Dataset.update_path behaves inconsistently
## Summary

In `nimare.utils.validate_images_df`, which is called whenever [`Dataset.images` is set](https://github.com/neurostuff/NiMARE/blob/713a231527b3d4615a86161bcf174e0a52c96752/nimare/dataset.py#L194-L197), the absolute path and relative path are resolved to identify the path that is shared across the images.

Here's the relevant part:

https://github.com/neurostuff/NiMARE/blob/6b233477f7086acad21d70b84b763b3cb1895b0a/nimare/utils.py#L201-L205

To be honest, I feel like the whole process here should be revisited. The `Dataset.path` attribute should be included in `validate_images_df`, `Dataset.images` shouldn't be set (triggering the call to `validate_images_df`) so often in the CorrelationDecoder, since it should create the MA maps once at the beginning of the run.

## Additional details

Run a CorrelationDecoder.

The "shared path" that is detected should be the same across the full CorrelationDecoder run.

### What actually happened?

It vacillated between the actual path (`/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/data/ns_dset_maps/`) and the root (`/`).

## Reproducing the bug

This came up in [this script](https://github.com/NBCLab/nimare-paper/blob/master/listings_figures_and_tables_part3.py).

tsalo:
Closes #471. While we should still revisit how images are validated and paths to images are stored, this at least should fix the immediate issue.

- Refactor and clean up find_stem, so that it finds the common _stem_ (or beginning substring) instead of the largest common substring in any part of the set of strings.
- Add test for find_stem.

### DIFF
--- a/nimare/tests/test_utils.py
+++ b/nimare/tests/test_utils.py
@@ -9,6 +9,49 @@ import pytest
 from nimare import utils
 
 
+def test_find_stem():
+    """Test nimare.utils.find_stem."""
+    test_array = [
+        "/home/data/dataset/file1.nii.gz",
+        "/home/data/dataset/file2.nii.gz",
+        "/home/data/dataset/file3.nii.gz",
+        "/home/data/dataset/file4.nii.gz",
+        "/home/data/dataset/file5.nii.gz",
+    ]
+    stem = utils.find_stem(test_array)
+    assert stem == "/home/data/dataset/file"
+
+    test_array = [
+        "/home/data/dataset/subfolder1/file1.nii.gz",
+        "/home/data/dataset/subfolder1/file2.nii.gz",
+        "/home/data/dataset/subfolder2/file3.nii.gz",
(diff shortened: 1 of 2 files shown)


tsalo:
@jdkent I don't think the proposed changes will require a deep review, but I'd appreciate it if you could take a quick look when you have some time.

jdkent (commented):
Looks good to me (but do not entirely have all the context), does one of the tests cover where the function was previously failing?

tsalo:
>does one of the tests cover where the function was previously failing?

The third and fourth ones do. In both cases the longest matching substring doesn't come from the beginning of the strings.

PR #476, opened 2021-03-25
[REF] Add low_memory option to Estimators and add function for moving metadata from Dataset to DataFrame

### LINKED ISSUE #470: Recent changes to kernel transformers raise memory usage at masking step
## Summary

Our recent changes to the way kernel transformers operate under low-memory conditions (#453, #460, #462, #463) unfortunately have uncovered an issue with apply maskers to large imaging files (or lists of files) that seems to be driven by how `nilearn` loads the images into memory.

Before our recent changes, the MA maps were generated from scratch automatically, even when they _should_ have been loaded from the `Dataset.images` attribute instead. Somehow, using nilearn on files requires more memory at masking than using it on img objects? I'm not 100% sure.

## Additional details

Run a CorrelationDecoder on the full Neurosynth database (~14k studies).

I expected the CorrelationDecoder to run successfully, with MA maps (saved as files) loaded into memory without issue.

### What actually happened?
Here's the traceback from a CorrelationDecoder with MDKAChi2, run on only 5k studies from Neurosynth:

```bash
INFO:nimare.meta.cbma.mkda:Loading pre-generated MA maps for Dataset 1.
INFO:nimare.meta.cbma.mkda:Loading pre-generated MA maps for Dataset 2.
Traceback (most recent call last):
  File "listings_figures_and_tables_part3.py", line 56, in <module>
    decoder.fit(ns_dset_first5000)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/base.py", line 417, in fit
    self._fit(dataset)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/decode/continuous.py", line 188, in _fit
    self.meta_estimator.fit(feature_dset, nonfeature_dset)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/meta/cbma/base.py", line 671, in fit
    maps = self._fit(dataset1, dataset2)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/meta/cbma/mkda.py", line 219, in _fit
    ma_maps2 = self.masker.transform(self.inputs_["ma_maps2"])
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/input_data/base_masker.py", line 176, in transform
    return self.transform_single_imgs(imgs, confounds)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/input_data/nifti_masker.py", line 408, in transform_single_imgs
    dtype=self.dtype
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/joblib/memory.py", line 352, in __call__
    return self.func(*args, **kwargs)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/input_data/nifti_masker.py", line 58, in filter_and_mask
    dtype=dtype)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/input_data/base_masker.py", line 99, in filter_and_extract
    memory_level=memory_level)(imgs)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/joblib/memory.py", line 352, in __call__
    return self.func(*args, **kwargs)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/input_data/nifti_masker.py", line 31, in __call__
    dtype=img_data_dtype(imgs)), imgs.affine)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/masking.py", line 761, in apply_mask
    ensure_finite=ensure_finite)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/masking.py", line 804, in _apply_mask_fmri
    copy=True)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/_utils/numpy_conversions.py", line 110, in as_ndarray
    ret = _asarray(arr, dtype=dtype, order=order)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/nilearn/_utils/numpy_conversions.py", line 27, in _asarray
    ret = np.asarray(arr, dtype=dtype, order=order)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/nimare-paper/conda_env/lib/python3.7/site-packages/numpy/core/_asarray.py", line 102, in asarray
    return array(a, dtype, copy=False, order=order)
numpy.core._exceptions.MemoryError: Unable to allocate 33.3 GiB for an array with shape (91, 109, 91, 4956) and data type int64
```

## Reproducing the bug

This bug can be reproduced with the code at https://github.com/NBCLab/nimare-paper/blob/23c265098337d4d09e37c15bf42fdd12a2d6894d/listings_figures_and_tables_part3.py#L13-L58. To reproduce from scratch you'll need to download the Neurosynth dataset.

tsalo:
Closes #470 and helps with a problem @62442katieb's having with the metaCluster workflow.

EDIT: I know that this PR reflects two different (but related) sets of changes. My apologies for that.

- Add a `low_memory` option to Estimators for storing their MA maps.
- Add `n_files` argument to `use_memmap` for managing multiple memmapped arrays in one object.
- Add a new function, `add_metadata_to_dataframe`, which aggregates metadata from a Dataset to a DataFrame. Currently this is just used to collect `sample_sizes` from the Dataset and add a column called `sample_size` with the mean value to the coordinates DataFrame for ALE meta-analyses, but it will be useful for metaCluster, which generates MA maps outside of an Estimator.


jdkent:
I'm waffling on the responsibility of `use_memmap`, do we think the 2D arrays are going to be too big?
how big is `ma_maps1` with 5000 studies?

just thinking about what types of arrays we want to use `use_memmap` for and how to do that consistently across estimators.

tsalo:
Our default mask has 228453 voxels in the mask, so 228453 * 5000 = 1142265000, and with a `float64` array that would be 9.1 GB, I think. If we look at the full Neurosynth dataset (~14k studies), I believe we'd have an array that's ~25.6 GB. I think it's probably worth it to support a low_memory option in that case.

nimare/meta/cbma/mkda.py
```diff
@@ -204,7 +206,19 @@ def _fit(self, dataset1, dataset2):
         if "ma_maps1" in self.inputs_.keys():
```
jdkent, line 206: since this code is repeated (and now contains work to use a memmap file), it would be more readable as a function such as `get_ma_maps`, which could be defined in `CBMAEstimator` I think. (it would have to take `self.memmap_filenames[index]` as an argument)

nimare/tests/test_meta_mkda.py
```diff
@@ -53,7 +53,16 @@ def test_mkda_density(testdata_cbma):
     assert isinstance(cres, nimare.results.MetaResult)
 
 
-def test_mkda_chi2_fdr(testdata_cbma):
+def test_MKDADensity_low_memory(testdata_cbma):
+    """Smoke test for MKDADensity with low_memory option."""
+    meta = MKDADensity(null_method="empirical", n_iters=10, low_memory=True)
+    res = meta.fit(testdata_cbma)
+    assert meta.low_memory
+    assert not meta.kernel_transformer.low_memory
```
jdkent, line 61: is there a use case where we would want the estimator and kernel to have different settings? Or should low_memory be set in the kernel what the setting is in the estimator?

jdkent (changes requested):
could you add `@use_memmap` here instead? https://github.com/neurostuff/NiMARE/pull/476/files#diff-4cafb2c711a65e5b19450202efbd9250dc086c814fb97588c1fd88afcc1add0cR624

That way, all `PairwiseCBMAEstimator` classes would inherit the decorator.

nimare/tests/test_meta_mkda.py
tsalo: I'm not sure. I _do_ want to make sure there isn't unintentional bleedover of parameters, but it does seem reasonable to do it intentionally. We'll want to log something about overwriting the kernel transformer's parameter though.

tsalo:
The link you added doesn't go to a specific line. Could you add an inline comment?

nimare/meta/cbma/mkda.py
tsalo: Good point! Can do.

nimare/meta/cbma/mkda.py
```diff
@@ -195,6 +196,7 @@ def __init__(self, kernel_transformer=MKDAKernel, prior=0.5, **kwargs):
 
         self.prior = prior
 
+    @use_memmap(LGR)
```
jdkent, line 199: @use_memmap(LGR, n_files=2) (but I think this decorator should go in the `PairCBMAEstimator` class)

tsalo: Will it work if the decorator is applied to the whole class or `fit()` instead of `_fit()`?

jdkent: in `fit` instead of `_fit`, sorry that was not clear

tsalo: After digging into the CorrelationDecoder error a bit, it seems like the decorator impacts `inspect.getfullargspec`:

```python
import inspect
from nimare.meta.cbma import MKDAChi2, MKDADensity

print(inspect.getfullargspec(MKDAChi2.fit).args)
print(inspect.getfullargspec(MKDAChi2._fit).args)
print(inspect.getfullargspec(MKDADensity.fit).args)
print(inspect.getfullargspec(MKDADensity._fit).args)
```

I would expect to get:

```
['self', 'dataset1', 'dataset2']
['self', 'dataset1', 'dataset2']
['self', 'dataset']
['self', 'dataset']
```

However what I get is:

```
['self']
['self', 'dataset1', 'dataset2']
['self', 'dataset']
['self']
```

Since the decorator is applied to `fit` for `MKDAChi2` and `_fit` for `MKDADensity`, it looks like those are the ones that are affected.

tsalo: I actually need `n_files` to be 3 in ALESubtraction because the difference scores can be mem-mapped, so I think it's probably best to switch back to decorating `_fit`. Maybe I could decorate a new abstractmethod within PairwiseCBMAEstimator?

nimare/tests/test_meta_mkda.py
tsalo: I think I want to handle this in a separate PR. It should be fairly easy to implement, but since @62442katieb is waiting on these changes for the metaCluster workflow I'd like to get it merged sooner rather than later.

jdkent: sounds good to me

jdkent (approved):
LGTM, we will continue low_memory estimator/kernel discussion in another PR.

Issue #452, opened 2021-03-05
Memory issue with kernel transformation on large datasets


tsalo:
## Summary

The CorrelationDecoder works by generating MA maps for all studies in the Dataset. Those MA maps are saved to files to reduce memory usage, but the initial step is causing an out-of-memory issue with the full Neurosynth dataset.

## Additional details

- NiMARE version: 0.0.7

### What were you trying to do?
Run a CorrelationDecoder on the Neurosynth dataset.

### What did you expect to happen?
I would expect the decoder to be very slow, but it should run successfully.

### What actually happened?
```
Traceback (most recent call last):
  File "listings_figures_and_tables.py", line 620, in <module>
    decoder.fit(ns_dset)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/base.py", line 417, in fit
    self._fit(dataset)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/decode/continuous.py", line 183, in _fit
    dataset = kernel_transformer.transform(dataset, return_type="dataset")
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/meta/kernel.py", line 161, in transform
    transformed_maps = self._transform(mask, coordinates)
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/meta/kernel.py", line 292, in _transform
    dims, vox_dims, ijks, self.r, self.value, exp_idx, sum_overlap=self._sum_overlap
  File "/home/data/nbc/misc-projects/Salo_NiMARE/NiMARE/nimare/meta/utils.py", line 303, in compute_kda_ma
    kernel_data = np.zeros([n_studies] + list(shape), dtype=type(value))
numpy.core._exceptions.MemoryError: Unable to allocate 96.6 GiB for an array with shape (14371, 91, 109, 91) and data type int64
```

## Reproducing the bug

```python
import nimare

nimare.extract.fetch_neurosynth("data/", unpack=True)
ns_dset = nimare.io.convert_neurosynth_to_dataset(
    "data/database.txt",
    "data/features.txt",
)
decoder = nimare.decode.continuous.CorrelationDecoder(
    feature_group="Neurosynth_tfidf",
    frequency_threshold=0.001,
    meta_estimator=nimare.meta.cbma.mkda.MKDAChi2,
    target_image="z_desc-specificity",
)
decoder.fit(ns_dset)
```

[...]

PR #484, opened 2021-04-07
[WIP,ENH] reduce low memory bottlenecks

jdkent:
ref #452 .

Passing in images one at a time to transform takes a lot of time, especially if reading in neurosynth, and the machine can presumably load more than one file at a time, this pull request makes the use of `low_memory` more flexible to specify the memory limit (e.g., `1gb` or `1024mb`, etc). 
- add `determine_chunk_size` function
- use `determine_chunk_size` when transforming files.

This function could be applied to other bottlenecks as we identify them.


nimare/meta/cbma/base.py
```diff
@@ -150,8 +150,18 @@ def _collect_ma_maps(self, coords_key="coordinates", maps_key="ma_maps", fname_i
...
+                chunk_size = determine_chunk_size(limit, temp)
+                map_chunks = [
+                    self.inputs_[maps_key][i : i + chunk_size]
+                    for i in range(0, len(self.inputs_[maps_key]), chunk_size)
+                ]
+                idx = 0
+                for map_chunk in map_chunks:
+                    end_idx = idx + len(map_chunk)
+                    ma_maps[idx:end_idx, :] = self.masker.transform(map_chunk)
+                    idx = end_idx
```
tsalo, line 164: I'm having this issue with the `CorrelationDistributionDecoder`, and I think it could apply to any case where we're masking many images at once. WDYT of moving this into a new function?

Example signature:

```python
masked_data = safe_transform(imgs, masker, memory_limit="1gb", dtype="auto")
```

`memory_limit` would feed into `determine_chunk_size`, of course, and we could have `dtype` to let the function coerce the data to a given datatype.

nimare/meta/utils.py
```diff
@@ -321,6 +323,8 @@ def compute_kda_ma(
     cube = np.vstack([row.ravel() for row in np.mgrid[xx, yy, zz]])
     kernel = cube[:, np.sum(np.dot(np.diag(vox_dims), cube) ** 2, 0) ** 0.5 <= r]
 
+    if low_memory:
+        chunk_size = determine_chunk_size(limit=low_memory, arr=ijks[0])
     for i, peak in enumerate(ijks):
```
tsalo, line 328:
```suggestion
        chunk_size = determine_chunk_size(limit=low_memory, arr=ijks[0])

    for i, peak in enumerate(ijks):
```

For readability.

PR #486, opened 2021-04-08
[REF] Miscellaneous GCLDA cleanup

tsalo:
Closes None, but related to #130.

- Clean up comments.
- Convert loglikely attributes to a dictionary (`self.loglikelihood`).
- Remove drop steps since mismatched IDs should be addressed at initialization.
- Additional cleanup.

### DIFF
--- a/nimare/annotate/gclda.py
+++ b/nimare/annotate/gclda.py
@@ -438,14 +438,14 @@ class GCLDAModel(NiMAREBase):
         np.random.seed(randseed)
 
         # Loop over all word tokens
-        for i_wtoken in range(len(self.data["wtoken_word_idx"])):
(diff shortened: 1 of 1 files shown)


nimare/annotate/gclda.py
```diff
@@ -663,28 +645,61 @@ def _update_regions(self):
 
                 # Estimate Covariances
                 # Covariances are estimated independently
```
tsalo, line 647: @tyarkoni, this means that, while the centroids of the Gaussians will be symmetric, the actual shapes could be different, right? I've noticed that in fits in the past and it only just struck me recently that that was different from what I assumed would happen (i.e., I assumed that the symmetric distributions would be completely symmetric).

Issue #475, opened 2021-03-23
Evaluate bias of IBMA estimators under different masking methods


tsalo:
## Summary

In #466, @nicholst and @tyarkoni note that maskers which aggregated values across voxels _before_ fitting the meta-analytic model will likely produce biased results, depending on the meta-analytic model. We should systematically evaluate the different estimators across a range of datasets.

## Additional details

@tyarkoni has performed some simulations, and did not find large bias across approaches for the non-combination, non-likelihood estimators (e.g., Hedges, WeightedLeastSquares, DerSimonianLaird, and probably PermutedOLS). The combination test estimators (Fishers and Stouffers) are probably heavily biased. The likelihood-based estimators (SampleSizeBasedLikelihood and VarianceBasedLikelihood) may or may not be biased.

@nicholst proposed the following options:

1. OLS - We ignore ROI variances and the weighting is `tau^2+const` (no weighting), worst case is inefficiency and (as per Mumford & Nichols) no FPR risk for one-sample or balanced two-sample comparisons.  (However, the M&N result was calibrated against heterogeneity seen in task fMRI, and not N=10 <-> N=1200 differences)
1. GLS - We take average ROI variances as "correct", but they're actually too small, so weighting is `tau^2 + TooSmallVar_i`... I think this is OK as the estimated `tau^2` will make up for the vars being too small over all, so inferences are probably fine, but just not as efficient as they could be.  Another plus is that this approach will capture gross differences in sample size, something important if N's have a big range.

## Analysis plan (tentative)

1. Collect subject-level data (z, p, beta, and varcope maps) from a range of datasets.
    - We can collect these data from Neuroscout.
1. Generate a range dataset-level results with resampling.
1. Run voxel-wise image-based meta-analyses, then average results across ROIs.
1. Run ROI-wise image-based meta-analyses.
1. Compare results of both approaches. The former is the ground truth.
    - As the most basic test, we can perform pair-wise comparisons between the analysis-first and the aggregation-first results from the same estimators and datasets.
    - We can also dig into dataset parameters/characteristics, which might clarify what sources of bias there are.

PR #495, opened 2021-04-20
[ENH] Incorporate information about valid masking approaches into IBMA Estimators

### LINKED ISSUE #466: Passing certain maskers (e.g., NiftiLabelsMasker) will produce invalid results for IBMA estimators
## Summary
Currently I believe we let users pass any nilearn `Masker` in at estimator initialization, and then internally use that masker to reduce the data. This works fine for certain estimators, but will lead to wonky/invalid inferences in other cases.

## Additional details
Consider any of the PyMARE-based IBMA estimators that expect the input to be pairs of estimate and variance maps. Suppose the user passes a `NiftiLabelsMasker` as the mask. Then the expectation is presumably that the meta-analysis will be done separately for each discretely-labeled ROI in the mask. Naively, the masker will reduce the dataset by averaging over all voxels within each ROI—and it will do this (I think) for both the parameter estimate and the sampling variance maps. The former is not a problem, but the latter is problematic because (a) we need to scale the variance by the number of voxels in the ROI (which we could do ourselves, but it would be an extra step), and (b) voxels are dependent, and we have no idea what the covariance matrix looks like (nor any good way to estimate it). The net effect is that the variances being passed to PyMARE internally will be (heavily) biased, and hence so will p/z-values. (Credit to @nicholst for pointing this out in email a while ago.)

A related problem will arise for the combination tests (e.g., Fisher's or Stouffer's): I presume the naively applied masker will average z-scores within ROIs, but this is not at all what we want—we presumably would need to aggregate values within each ROI via some Fisher's-like procedure (but not using any of the methods currently in PyMARE, because they all assume independence).

## Reproducing the bug
Almost any ROI-based IBMA analysis will suffer from this issue. The exceptions are the estimators that take parameter estimates but not variances as inputs (I think that's currently only the `SampleSizeBasedLikelihood` estimator, which frankly is kind of crappy).

## Proposed solution
Disable ROI-based analyses entirely in the package until we can figure out a sensible solution to the problem (which is going to be... hard). Masking is fine so long as the user's just defining the valid voxel space, but as soon as there's aggregation of voxels, all bets are off.

tsalo:
Closes #466. The actual investigation into the effect of aggregation on IBMA results can continue in #475.

- Raise ValueError at `fit()` of Fishers and Stouffers Estimators if the masker is not a NiftiMasker.
- Log warning at `fit()` of WeightedLeastSquares, DerSimonianLaird, Hedges, and VarianceBasedLikelihood Estimators if the masker is not a NiftiMasker.
- Add test that checks for expected behavior of each Estimator when supplied a NiftiLabelsMasker.
- Add information about valid masking approaches to docstrings of affected Estimators.

### DIFF
--- a/nimare/meta/ibma.py
+++ b/nimare/meta/ibma.py
@@ -5,8 +5,8 @@ import logging
 
 import numpy as np
 import pymare
-from nilearn.mass_univariate import permuted_ols
 from nilearn.input_data import NiftiMasker
+from nilearn.mass_univariate import permuted_ols
 
 from ..base import MetaEstimator
 from ..transforms import p_to_z, t_to_z
@@ -27,6 +27,9 @@ class Fishers(MetaEstimator):
     Warning
     -------
     This method does not currently calculate p-values correctly. Do not use.
+
+    Masking approaches which average across voxels (e.g., NiftiLabelsMaskers)
+    will result in invalid results. It cannot be used with these types of maskers.
 
     All image-based meta-analysis estimators adopt an aggressive masking
     strategy, in which any voxels with a value of zero in any of the input maps
@@ -85,6 +88,9 @@ class Stouffers(MetaEstimator):
     Warning
     -------
     This method does not currently calculate p-values correctly. Do not use.
+
+    Masking approaches which average across voxels (e.g., NiftiLabelsMaskers)
+    will result in invalid results. It cannot be used with these types of maskers.
 
     All image-based meta-analysis estimators adopt an aggressive masking
     strategy, in which any voxels with a value of zero in any of the input maps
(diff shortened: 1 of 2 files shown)


tsalo:
The current issue is that, when the mask isn't set at the IBMA estimator's initialization, `IBMAEstimator.masker` is `None`, and the masker is taken from the `Dataset` in `MetaEstimator._preprocess_input()`. Which means that a check in `IBMAEstimator.__init__()` isn't sufficient. I could put a check in `__init__()` _and_ in `_fit()` though? Unfortunately that means some duplicate code. The alternative is to just have the check in `_fit()`, but that means postponing the masker check until when the user calls `fit()`.

PR #498, opened 2021-04-22
[FIX] Deal with extreme t-values in t_to_z by truncating associated p-values

tsalo:
Closes None. This partially addresses the problem identified in https://github.com/neurostuff/PyMARE/issues/61. 

Specifically, PyMARE was only responsible for some of the NaNs in the meta-analysis results (see picture below). A big part of the problem was that study 2621, which has a t-statistic map on Neurovault, has a very large sample size (1385) and very high t-statistics (max is ~85). When fed through [`nimare.transforms.t_to_z()`](https://nimare.readthedocs.io/en/latest/generated/nimare.transforms.t_to_z.html#nimare.transforms.t_to_z), the resulting z-values for those very-significant t-statistics became NaNs. Therefore, I think we need to crop extreme p-values (from t-statistic + dof --> p-value conversion) to epsilon before transforming _those_ to z-values.

## Before:
image

## After:
image

- Crop all extreme p-values to epsilon. Note that I used epsilon as the threshold instead of 0, because for some reason some values between 0 and epsilon were actually converting cleanly to z-values, and the resulting maps looked... weird.

### DIFF
--- a/nimare/transforms.py
+++ b/nimare/transforms.py
@@ -631,10 +631,12 @@ def t_to_z(t_values, dof):
 
     # Calculate p values for <=0
     p_values_t1 = stats.t.cdf(t1, df=dof)
+    p_values_t1[p_values_t1 < np.finfo(p_values_t1.dtype).eps] = np.finfo(p_values_t1.dtype).eps
     z_values_t1 = stats.norm.ppf(p_values_t1)
 
     # Calculate p values for > 0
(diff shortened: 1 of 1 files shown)


jdkent (approved):
This looks like a reasonable fix to me, thanks!

tsalo:
Thanks @jdkent!

PR #499, opened 2021-04-23
[ENH] Support symmetric GCLDA topics with more than two subregions

### LINKED ISSUE #491: Support symmetric option for even sets of regions in GCLDA
## Summary

I think it would be fairly easy to define symmetric regions as pairs instead of hardcoding region 1 and region 2 in GCLDA. This might make it easier to develop more network-based topics.

## Additional details

Network-based topics, probably?

## Next steps

1. Loop through pairs of regions in the `update_regions` method instead of matching region 1 and region 2 specifically. Each pair would have the same centroid.
2. Remove the restriction on symmetric.

tsalo:
Closes #491.

- Allow multiple symmetric pairs of subregions in GCLDA.

### DIFF
--- a/examples/03_annotation/plot_gclda.py
+++ b/examples/03_annotation/plot_gclda.py
@@ -66,6 +66,8 @@ model = annotate.gclda.GCLDAModel(
     dset.coordinates,
     mask=dset.masker.mask_img,
     n_topics=10,
+    n_regions=4,
+    symmetric=True,
 )
 model.fit(n_iters=100, loglikely_freq=20)
 model.save("gclda_model.pkl.gz")
--- a/nimare/annotate/gclda.py
+++ b/nimare/annotate/gclda.py
@@ -128,9 +128,9 @@ class GCLDAModel(NiMAREBase):
         coordinates_df = coordinates_df.loc[coordinates_df["id"].isin(ids)]
 
         # --- Checking to make sure parameters are valid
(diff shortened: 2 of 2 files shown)


tsalo:
@tyarkoni no pressure to review this PR in detail if you don't have the time, but I'd like to know if the general idea is, well, _bad_.

tyarkoni:
I don't have time to review this in detail, unfortunately, but I don't see any principled problem with this, so long as the results look sane for reasonable values. Do they look sane?

tsalo:
Thanks! The four-region results in the example look reasonable, but I will try running a more realistic version soon:

Screen Shot 2021-04-26 at 4 29 45 PM

One question- why are the paired regions' covariances estimated independently? For the "symmetric" option, wouldn't it make more sense for the regions to be symmetric not just in terms of centroid, but also in terms of skew and direction?

tyarkoni:
Yeah that looks fine to me.

Re: symmetric covariances, I don't remember the reasoning, but I'm guessing it was probably some combination of (a) the brain isn't actually completely symmetric, and (b) in a sense it's a nice sanity check, because if the covariances look pretty similar, despite being estimated separately, that's a good sign the model is behave reasonably. But I wouldn't have any objection to parameterizing that as an option too.

tsalo:
Thanks. I'll open an issue about a new option for symmetric covariances.

PR #501, opened 2021-04-26
[DOC] Add `versionadded` and `versionchanged` directives to docstrings

### LINKED ISSUE #489: Start using `versionadded` and `versionupdated` directives in docstrings
## Summary

I would like to use [Sphinx decorators (or directives)](https://deprecated.readthedocs.io/en/latest/sphinx_deco.html) in docstrings to indicate the NiMARE version in which the item was added or last meaningfully changed.

## Additional details

This will help users keep track of changes to NiMARE.

JulioAPeraza:
Closes #489 .

- Add `versionadded` and `versionchanged` directives to docstrings.

### DIFF
--- a/nimare/annotate/utils.py
+++ b/nimare/annotate/utils.py
@@ -1,10 +1,8 @@
 """Utility functions for ontology tools."""
 import numpy as np
 import pandas as pd
-from deprecated.sphinx import versionadded
 
 
-@versionadded(version="0.0.3", reason="version < 0.0.3: nimare.annotate.ontology.utils")
 def _generate_weights(rel_df, weights):
     """Create an IDxID DataFrame linking weight value to each relationship type.
 
--- a/nimare/annotate/utils.py
+++ b/nimare/annotate/utils.py
@@ -1,10 +1,18 @@
 """Utility functions for ontology tools."""
 import numpy as np
 import pandas as pd
+from deprecated.sphinx import versionadded
 
 
+@versionadded(version="0.0.3", reason="version < 0.0.3: nimare.annotate.ontology.utils")
 def _generate_weights(rel_df, weights):
     """Create an IDxID DataFrame linking weight value to each relationship type.
+
+    .. versionadded:: 0.0.2
+
+    .. versionchanged:: 0.0.3
+        [ENH] Submodule was reorginized (`nimare.annotate.ontology.utils` in previous
+        version)
 
     Parameters
     ----------
--- a/nimare/annotate/utils.py
+++ b/nimare/annotate/utils.py
@@ -8,10 +8,6 @@ def _generate_weights(rel_df, weights):
 
     .. versionadded:: 0.0.2
 
-    .. versionchanged:: 0.0.3
-        [ENH] Submodule was reorginized (`nimare.annotate.ontology.utils` in previous
-        version)
-
     Parameters
     ----------
     rel_df : (X x 3) :obj:`pandas.DataFrame`
--- a/nimare/decode/continuous.py
+++ b/nimare/decode/continuous.py
@@ -23,6 +23,8 @@ LGR = logging.getLogger(__name__)
 def gclda_decode_map(model, image, topic_priors=None, prior_weight=1):
     r"""Perform image-to-text decoding for continuous inputs using method from Rubin et al. (2017).
 
+    .. versionadded:: 0.0.1
+
     Parameters
     ----------
     model : :obj:`nimare.annotate.topic.GCLDAModel`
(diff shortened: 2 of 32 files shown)


JulioAPeraza:
@tsalo So far, I have added directives to functions and classes inside annotate. Please, let me know if you would like more detailed comments on `versionchanged` directive.

nimare/annotate/utils.py
```diff
@@ -1,11 +1,19 @@
 """Utility functions for ontology tools."""
 import numpy as np
 import pandas as pd
+from deprecated.sphinx import versionadded
 
 
+@versionadded(version="0.0.3", reason="version < 0.0.3: nimare.annotate.ontology.utils")
```
tsalo, line 7: I think you'll want to remove these two lines.

tsalo:
This is looking really good, but `build_docs` is failing due to updated dependencies. I think I fixed it in `master` in #502, so you probably just need to update your branch.

nimare/decode/continuous.py
```diff
@@ -220,6 +234,19 @@ def transform(self, img):
...
 
+    .. versionadded:: 0.0.1
+
+    .. versionchanged:: 0.0.2
+        [ENH] Change the order of the first two parameters (img, dataset) -> (dataset, img)
+
+    .. versionchanged:: 0.0.3
+        [ENH] New parameter *feature_group=None*.
+        [ENH] Convert decoder :func:`corr_dist_decode` to class
+        :class:`CorrelationDistributionDecoder`.
```
tsalo, line 245: It looks like newlines aren't respected, so this renders as a paragraph.

nimare/annotate/utils.py
```diff
@@ -6,6 +6,12 @@
 def _generate_weights(rel_df, weights):
     """Create an IDxID DataFrame linking weight value to each relationship type.
 
+    .. versionadded:: 0.0.2
+
+    .. versionchanged:: 0.0.3
+        [ENH] Submodule was reorginized (`nimare.annotate.ontology.utils` in previous
```
tsalo, line 12: Reorganized is misspelled as reorginized.

nimare/decode/continuous.py
JulioAPeraza: We could also use bullet points?
Screen Shot 2021-04-30 at 12 18 56 PM

tsalo: That does look much better!

JulioAPeraza:
Here is a list of some possible options:

1. Position in the docstring
a) At the top. 
b) At the bottom.
**Note:** The most common place is at the end, but they mix up with the following function in the docstrings. I think it looks good at the top.

2. `versionchanged`
a) Include changes from all versions.
b) Include changes from the last two releases (or the last two changes).
c) Include when there was a bug.
d) Change of parameters in input and output.
e) Internal enhancement of the function.
**Note:** I agree that `a)` will clutter up the docstrings. Most projects report `versionchanged` for any major change (in the input and output) in the previous 2-3 releases. I included `c)` here just to let the users know that the was a bug in the previous versions and it was fixed in that release.

3. Style
a) Paragraph.
b) Bullet points.
c) Include pull request titles (`FIX`, `ENH`, `REF`).
**Note:** Paragraph looks good too, it is the most common out there. I thought to include `c)` here so the user can easily identify if the change was just an `ENH` or a `FIX`.

I like the combination: 1(a), 2(b, c, d), 3(b, c); but I am probably missing something.

tsalo:
Thank you for describing the problem space. I think 1 depends on how long things end up getting, but I agree that (a) sounds good for now. Regarding the last two, I agree with 2(b, c, d) and 3(b). I'm less sure about 3(c), but if we don't end up liking it we can always remove those prefixes at a later date, so feel free to include the prefixes in this PR.

JulioAPeraza:
I forgot to split option `2b`. 
Should I add `b:` the last two changes (regardless of the version) or `b':` the changes in the last two versions?

tsalo:
I think `b'`, with the caveat that bullets shouldn't necessarily reflect PRs. If there are two PRs addressing related sets of changes, those should be in one bullet, so that the docstring doesn't get too long.

nimare/extract/utils.py
```diff
@@ -153,7 +159,12 @@ def readlinkabs(link):
 
 
 def _download_zipped_file(url, filename=None):
-    """Download from a URL to a file."""
+    """Download from a URL to a file.
+
+    .. versionadded:: 0.0.2
+
+    """
+
```
tsalo, line 167:
```suggestion
    """
```

tsalo (approved):
The changes LGTM. Thanks!

One caveat: I might adjust some of the notes in later PRs, but I don't have any specific things I want to change at the moment.

tsalo:
Not sure why I didn't merge this last month (sorry about that), but I'm merging now. Thanks for this.

PR #506, opened 2021-05-06
[DOC] Improve documentation of decoders

### LINKED ISSUE #436: Improve documentation of decoders
## Summary

We've had a couple of recent NeuroStars questions about the decoding methods in `NiMARE`, and I think they're due to poor documentation. For reference, see [here](https://neurostars.org/t/decoding-with-surface-based-data/18110) and [here](https://neurostars.org/t/nimare-decoding-results/18148).

## Next steps

1. Add a non-running example for the continuous decoders.
2. Describe the discrete decoders in _much_ more detail, with an emphasis on the outputs and their interpretation.

JulioAPeraza:
Closes #436.

- Improve the discrete decoders section (`methods/decoding.rst`).
- Add the continuous decoders section. Add a non-running example.

Open questions:

- Do we also want to add the encoders section as part of this pull request?


tsalo:
@JulioAPeraza Is this ready for review?

JulioAPeraza:
Yes, it is ready. Thanks!
In addition to the text and formulae from your NiMARE manuscript, I also incorporated text and a non-running example of the `CorrelationDecoder` with the `feature_group` argument, so the wording there may be a little confusing.

tsalo:
The math doesn't seem to be rendering properly. Are we missing the necessary plugin or is it maybe formatted for a different language (e.g., markdown)?

image

JulioAPeraza:
I don't know what is causing this. I built the documentation locally with the same version (`v4.2.0`) of Sphinx that CircleCI used, and the math looked fine.
To generate the latex images, I used the extension `"sphinx.ext.imgmath"`, which is already in the `conf.py` file.

tsalo:
Hm, okay. It could be that the CI build job is just not configured correctly. This is one good reason for us to switch to using the RTD CI tool instead of CircleCI for the docs build, though I had trouble setting it up the last time I tried.

Anyway, I'm leaning toward merging and then fixing if there are problems.

docs/methods/decoding.rst
```diff
@@ -1,38 +1,225 @@
...
+class object.
+
+
+The GCLDA approach
+`````````````````````
+:func:`nimare.decode.continuous.gclda_decode_map`
+
+Recently, it was introduced a new decoding framework based on Generalized Correspondence LDA 
+(GC-LDA, (`Rubin et al., 2017`_)), an extension to the LDA model. This method, in addition to the 
+two probability distributions from LDA (:math:`P(word|topic)`` and :math:`P(topic|article)``),
```
tsalo, line 123:
```suggestion
two probability distributions from LDA (:math:`P(word|topic)` and :math:`P(topic|article)`), 
```

tsalo (changes requested):
I think these extra backticks might be part of the problem. The other thing is that I'm a little worried about the hardcoded examples becoming out-of-date as NiMARE is updated, since the API is still not all that stable, but I'm happy to merge with them included and re-evaluate if it becomes a problem.

Anyway, if you fix those extra backticks, then I'll approve. Everything else looks good to me.

JulioAPeraza:
Do we need to install `latex` in the Docker image?

tsalo:
I'm thinking we should just merge and see how it goes. If there's a problem then we can fix later. Any concerns with me merging as-is?

JulioAPeraza:
Please, feel free to merge it. Thanks!

PR #519, opened 2021-06-07
[DOC] Add example for ALE subtraction and conjunction analysis

### LINKED ISSUE #515: Add conjunction analysis example using nilearn
## Summary

A question in today's OHBM NiMARE tutorial came up about conjunction analyses in NiMARE. These analyses should be possible with `nilearn.image.threshold_img` and `nilearn.image.math_img`, but it would be useful for NiMARE users if we had an example of how to do that in our documentation.

alexenge:
Closes #515.

- Added two Sleuth text files with data by Enge et al. (2021) on semantic cognition in children
- Added an example (based on these data) for a coordinate-based subtraction and conjunction analysis

This is my first PR, so please feel free to scrutinize rigorously. These three points might deserve special attention:

- Other examples don't have any references, so maybe they should be removed for consistency. However I, as a user, might find them helpful (especially the ones for the data and for the conjunction analysis which aren't referenced elsewhere).
- Other examples tend to present unthresholded images. This would be possible for the subtraction and conjunction examples as well; however, at least the two initial (group-specific) ALE images require thresholding so that the conjunction analysis becomes possible. Also, it might help to draw users' attention to the fact that single-group ALE and conjunction require *cluster-level* thresholding while subtraction requires *voxel-level* thresholding.
- I had to increase `n_iters` for the ALE FWE correction and for the subtraction analysis to `1000` or they would return empty masks. This, however, increases the total runtime of the script to ~4 min. If this is too long for an auto example, we could remove `plot_` from the file name and not have the script rendered at all (at the cost of not displaying the output figures).

Looking forward to your review. Thanks!


examples/02_meta-analyses/plot_cbma_subtraction_conjunction.py
```diff
@@ -0,0 +1,128 @@
+# emacs: -*- mode: python-mode; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 et:
+"""
+
+.. _metas8:
+
+========================================================
+ Run coordinate-based subtraction and conjunction analyses
+========================================================
```
tsalo, line 9: Can you extend the `=`s to match the length of the text?

examples/02_meta-analyses/plot_cbma_subtraction_conjunction.py
```diff
@@ -0,0 +1,128 @@
...
+# Individual group ALEs
+# --------------------------------------------------
+# Computing seperate ALE analyses for each group is not strictly necessary for
+# performing the subtraction analysis but will help to appreciate the
+# similarities and differences between the groups.
+ale = ALE(null_method="approximate")
+res = ale.fit(dset)
+res2 = ale.fit(dset2)
+
+corr = FWECorrector(method="montecarlo", voxel_thresh=0.001, n_iters=1000)
```
tsalo, line 50: GitHub and RTD can't handle anything that's too computationally intensive, so we need to minimize the number of iterations and limit to one core. I would try out `n_iters=100` and `n_cores=1`.

examples/02_meta-analyses/plot_cbma_subtraction_conjunction.py
```diff
@@ -0,0 +1,128 @@
...
+    title="Semantic relatedness",
+    threshold=2.326,  # cluster-level p < .01, one-tailed
+    cmap="RdBu_r",
+    vmax=4,
+)
+
+###############################################################################
+# Subtraction analysis
+# --------------------------------------------------
+sub = ALESubtraction(n_iters=1000, low_memory=False)
```
tsalo, line 79: We'll need to drop the number of iterations down a lot for this to run on GitHub/RTD. I think 100 would probably pass.

examples/02_meta-analyses/plot_cbma_subtraction_conjunction.py
```diff
@@ -0,0 +1,128 @@
...
+    vmax=4,
+)
+
+###############################################################################
+# Conjunction analysis
+# --------------------------------------------------
+# To determine the overlap of the meta-analytic results, a conjunction image
+# can be computed by (a) identifying voxels that were statistically significant
+# in *both* individual group maps and (b) selecting, for each of these voxels,
+# the smaller of the two group-specific *z* values. [4]_
```
tsalo, line 99: Could you also mention that, since this is simple image math, there is no NiMARE implementation for conjunction analysis, but it can be done easily with nilearn?

tsalo (changes requested):
This is really great, thanks! I have a couple of requests for wording, but the main thing is that the CI is failing because the analyses are just too intensive with their current settings. I have added a couple of recommendations for settings that should work better in CI.

examples/02_meta-analyses/plot_cbma_subtraction_conjunction.py
alexenge: Thanks a lot for your feedback! I've reduced `n_iters` and `n_cores` as suggested and the checks seem to pass now. Note, however, that I had to remove the thresholding for the subtraction image because `n_iters=100` didn't reliably detect anything at *p* < .01.

tsalo (approved):
This looks great, and it's passing! Going to merge now.

PR #520, opened 2021-06-09
[ENH] Replace low_memory with memory_limit and reduce memory bottlenecks

tsalo:
Closes #484. This PR adds some changes I made on top of @jdkent's #484. These changes might be useful, but they ended up causing problems, so I thought I'd open a PR that we could look at.

Changes proposed:
- Move masker transformation code in #484 into a separate function (`safe_transform`).
- Replace `low_memory` argument/attribute with `memory_limit`.
- Reduce memory usage of MKDAChi2.
- Replace Laird test dataset with json file, since there is a new error when loading old Datasets from pkl.gz files. The error probably stems from a new release of one of the dependencies, but I'm not going to dig into that here.


nimare/decode/continuous.py
```diff
@@ -238,13 +238,20 @@ class CorrelationDistributionDecoder(Decoder):
     evaluate results based on significance.
     """
 
+    @use_memmap(LGR, n_files=1)
```
tsalo, line 241: I don't think we can know how many memmaps are necessary, since there should probably be one file for each feature.

nimare/utils.py
```diff
@@ -454,6 +454,99 @@ def memmap_context(self, *args, **kwargs):
...
+    chunk_size = int(limit_bytes // arr_bytes)
+
+    if chunk_size == 0:
+        arr_size = arr_bytes // BYTE_CONVERSION["mb"]
+        raise RuntimeError(f"memory limit: {limit} too small for array with size {arr_size}mb")
+
+    return chunk_size
+
+
+def safe_transform(imgs, masker, memory_limit="1gb", dtype="auto", memfile=None):
```
tsalo, line 497: Here's the new function

nimare/decode/continuous.py
```diff
@@ -238,13 +238,20 @@ class CorrelationDistributionDecoder(Decoder):
...
     def __init__(
-        self, feature_group=None, features=None, frequency_threshold=0.001, target_image="z"
+        self,
+        feature_group=None,
+        features=None,
+        frequency_threshold=0.001,
+        target_image="z",
+        memory_limit="1gb",
```
tsalo, line 248: Maybe we should drop `low_memory` completely and replace with `memory_limit`. `memory_limit=None` == `low_memory=False`.

nimare/decode/continuous.py
tsalo: Dropping use_memmap from this class.

nimare/decode/continuous.py
tsalo: Done in 0892e75.

nimare/utils.py
```diff
@@ -478,6 +482,99 @@ def memmap_context(self, *args, **kwargs):
...
+        String representation of memory limit, can use:
+        kb, mb, gb, and tb as suffix (e.g., "4gb").
+    arr : :obj:`numpy.array`
+        Representative numpy array.
+    multiplier : :obj:`int`
+        Adjustment for processes that have more or
+        less overhead than expected.
+    """
+    limit = limit.lower()
+    size, representation = re.search(r"([0-9]+)([a-z]+)", limit).groups()
```
tsalo, line 510: It might be nice to validate the string here so that exceptions will be maximally informative.

tsalo:
The current problem seems to come up when a Dataset contains relative paths to images, but those paths contain the same folders/paths across all images. Then, `Dataset.update_path()` tries to update the absolute path, but leave the relative paths unchanged, while `Dataset.images` tries to maximize the base path, which would include the shared subfolders... Thus, the base path identified by `Dataset.images` and the one from `update_path` never match up and keep calling each other...

tsalo:
@jdkent do you mind if I merge this and close #484? We can revisit before the next release if there are any problems.

jdkent:
that is okay with me!

PR #527, opened 2021-06-11
[TST] Cover cases where some studies are missing coordinates

### LINKED ISSUE #408: Test mismatched images and coordinates in Dataset
We need to test behavior when a Dataset contains coordinates from only a subset of the studies.

This is related to #405 and #402. In #405, @jdkent identified a bug when there are _no_ coordinates, while in #402, I identified a bug in how MA maps were returned from a Dataset. If testing does identify a bug, then the solution may be to drop invalid studies automatically, as we do for images. Also see #204.

tsalo:
Closes #408.

- Remove coordinates from a subset of the studies in the _test_ NIDM pain dataset (not the real one). Specifically pain_01.nidm and pain_02.nidm.
- Hopefully fix any bugs associated with missing coordinates, or at least raise interpretable Exceptions.
- Do not overwrite `Estimator.inputs_` attribute when calling `Estimator._validate_input()`. This is necessary to retain fields in `Estimator.inputs_` for pairwise Estimators (which call `_validate_input()` and `_preprocess_input()` twice, with different Datasets).
- Do not overwrite coordinates input in `MetaEstimator._preprocess_input()`. This was undoing the important filtering done by `Dataset.get()` in `Estimator._validate_input()` to set `Estimator.inputs_["coordinates"]`.
- On that note, also actually filter coordinates DataFrame _and_ "ids" within Dataset.get based on which studies have coordinates.
- In ImageTransformer, replace NaNs with Nones to match our general missing data approach.


nimare/tests/test_transforms.py
```diff
@@ -166,71 +166,84 @@ def test_images_to_coordinates(tmp_path, caplog, testdata_ibma, kwargs, drop_dat
     # this transformation should retain the same number of unique ids
     # unless the merge_strategy was demolish
     if img2coord.merge_strategy == "demolish":
-        expected_studies_with_coordinates = set(tst_dset.coordinates["id"]) - set(
+        expected_studies_with_coordinates = set(tst_dset.ids) - set(
             studies_without_coordinates
         )
     else:
-        expected_studies_with_coordinates = set(tst_dset.coordinates["id"])
+        expected_studies_with_coordinates = set(tst_dset.ids)
```
tsalo, line 173: These are the sources of the current test failures.

tsalo: I just need a clean way to specify which studies should have coordinates. I _might_ just hardcode the IDs.

tsalo: Should be fixed in 3e06868.

tsalo:
Alright, I'm going to merge this. If anyone experiences any problems from it, please open an issue.

PR #533, opened 2021-06-30
[ENH] Add coordinate-based coactivation-based parcellation class

### LINKED ISSUE #260: Coactivation-based parcellation
Add coactivation-based parcellation algorithm, as described in Bzdok et al. (2013).

###References

Bzdok, D., Laird, A. R., Zilles, K., Fox, P. T., & Eickhoff, S. B. (2013). An investigation of the structural, connectional, and functional subspecialization in the human amygdala. Human brain mapping, 34(12), 3247-3266. https://doi.org/10.1002/hbm.22138

tsalo:
Closes #260.

- Add `n` option to `Dataset.get_studies_by_coordinate()`.
- Draft new `parcellate` module with `CoordCBP` class.

### DIFF
--- a/nimare/parcellate.py
+++ b/nimare/parcellate.py
@@ -1,0 +1,134 @@
+"""Parcellation tools."""
+import inspect
+import logging
+
+import numpy as np
+from sklearn.cluster import KMeans
+
+from .base import NiMAREBase
+from .meta.base import CBMAEstimator
+from .results import MetaResult
+from .utils import add_metadata_to_dataframe, check_type, listify, use_memmap, vox2mm
+
+LGR = logging.getLogger(__name__)
+
+
+class CoordCBP(NiMAREBase):
+    """Base class for parcellators in :mod:`nimare.parcellate`.
+
+    .. versionadded:: 0.0.10
+
+    """
+
+    _required_inputs = {"coordinates": ("coordinates", None)}
+
+    def __init__(
+        self,
+        target_mask,
+        meta_estimator,
+        target_image,
+        n_clusters,
+        r=None,
+        n=None,
+        *args,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+
+        # Get kernel transformer
+        meta_estimator = check_type(meta_estimator, CBMAEstimator)
+        self.meta_estimator = meta_estimator
+        self.target_image = target_image
+        self.n_clusters = listify(n_clusters)
+
+        if r and n:
+            raise ValueError("Only one of 'r' and 'n' may be provided.")
+        elif not r and not n:
+            raise ValueError("Either 'r' or 'n' must be provided.")
+        self.r = r
+        self.n = n
+
+    def _preprocess_input(self, dataset):
+        """Mask required input images using either the dataset's mask or the estimator's.
+
+        Also, insert required metadata into coordinates DataFrame.
+        """
+        super()._preprocess_input(dataset)
+
+        # All extra (non-ijk) parameters for a kernel should be overrideable as
+        # parameters to __init__, so we can access them with get_params()
+        kt_args = list(self.kernel_transformer.get_params().keys())
+
+        # Integrate "sample_size" from metadata into DataFrame so that
+        # kernel_transformer can access it.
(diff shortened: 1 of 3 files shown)


tsalo:
@mriedel56 @62442katieb if possible, I'd love it if you could check out the new class (especially the `_fit` method, which does the actual CBP) and give your thoughts. So far, I just have the most basic elements of the algorithm implemented, so I still need input on (1) the clustering algorithm options, (2) the metrics to use, and (3) the outputs to save.

Ultimately, I want this class to be fairly basic, meaning not including too many tunable parameters, with some documentation pointing toward [`cbptools`](https://cbptools.readthedocs.io/en/latest/index.html) for users who require more control.

nimare/parcellate.py
```diff
@@ -0,0 +1,166 @@
...
+            macm_data = self.meta_estimator.results.get_map(self.target_image, return_type="array")
+            if i_coord == 0:
+                data = np.zeros((target_xyz.shape[1], len(macm_data)))
+            data[i_coord, :] = macm_data
+
+        # Correlate voxel-wise MACM results
+        data = np.corrcoef(data)
+
+        # Convert voxel-wise correlation matrix to distances
+        data = 1 - data
```
tsalo, line 134: I'm not sure if [sklearn.cluster.KMeans](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html#sklearn.cluster.KMeans) automatically calculates the distance values or not. There's a `precompute_distances` parameter, but it's deprecated and has no effect.

tsalo: After looking through sklearn's examples a bit, I think it calculates distance internally, so we can feed it the raw ALE values.

nimare/parcellate.py
```diff
@@ -0,0 +1,223 @@
...
+                labels[i_filter, j_cluster, :] = kmeans.labels_
+
+        # Clean up MACM data memmap
+        LGR.info(f"Removing temporary file: {memmap_filename}")
+        os.remove(memmap_filename)
+
+        images = {"labels": labels}
+        return images
+
+    def _filter_selection(self):
```
tsalo, line 184: Chase 2020:
>We implemented a two-step procedure that involved a decision on those filter sizes to be included in the final analysis and subsequently a decision on the optimal cluster solution. In the first step, we examined the consistency of the cluster assignment for the individual voxels across the cluster solutions of the co-occurrence maps performed at different filter sizes. We selected a filter range with the lowest number of deviants, that is, number of voxels that were assigned differently compared with the solution from the majority of filters. In other words, we identified those filter sizes which produced solutions most similar to the consensus-solution across all filter sizes. For example, the proportion of deviants for the second parcellation is illustrated in Figure S1; this shows the borders of the filter range to be used for subsequent steps was based on the Z-scores of the number of deviants.

I interpret this to mean:

1. Derive mode array of label assignments for each cluster count across filter sizes.
    - I assume this means mode of each voxel determined independently, rather than mode of full set of assignments.
    - What if label numbers don't match? E.g., label 1 in filter size 1 is most similar to label 2 in filter size 2.
    - I assume we should do some kind of synchronization, unless there's some inherent order to KMeans labels?
2. Count number of voxels that don't match mode for each filter size.
3. Calculate proportion of deviants in each cluster solution and filter size.
3. Calculate weighted z-score for each filter size (across cluster solutions) somehow?
    - What is it weighted by?
4. Select range of filter sizes with lowest z-scores.
    - How? Is there some kind of threshold? Figure S1 grabs range with z-scores < -0.5. No clue if that's a meaningful threshold or something like 2 standard deviations from avg z-score or what.
    - What if there are multiple dips? Does amplitude (z-scores of filters below threshold) or width (number of filters below threshold) matter more?

adelavega:
https://github.com/neurosynth/neurosynth/blob/master/neurosynth/analysis/cluster.py

tsalo:
@62442katieb has some code from her naturalistic meta-analysis that may implement some of these metrics: https://github.com/62442katieb/meta-analytic-kmeans/blob/daf3904caad990aeadc89bc98769aaed32857e09/evaluating_clustering_solutions.ipynb

PR #537, opened 2021-07-12
[FIX] Update CircleCI config to fix recent bug

tsalo:
Closes None, but fixes recent bug.

- Run `apt-get update` before `apt-get install make` to fix recent bug.


tsalo:
Tests are taking >2X what they should be. I wonder if recent memory usage changes are adversely impacting performance, even when we don't want them to.

Issue #542, opened 2021-07-20
DerSimonianLaird failing due to SVD not converging


tsalo:
## Summary

DerSimonianLaird appears to be failing lately. I noticed this in the NHA tutorial (see neurohackademy/nh2021-curriculum#5).

I think this is due to some change made from `numpy` 1.21.0 (succeeds) to 1.21.1 (fails).

## Additional details

- NiMARE version: ~0.0.10rc1

### What were you trying to do?
Run the DerSimonianLaird IBMA Estimator on the pain dataset.

### What did you expect to happen?
The Estimator to run through without raising an exception.

### What actually happened?

Here's the full log from numpy 1.21.0 (some warnings, but no errors):

```
/Users/taylor/Documents/tsalo/nilearn/nilearn/datasets/__init__.py:92: FutureWarning: Fetchers from the nilearn.datasets module will be updated in version 0.9 to return python strings instead of bytes and Pandas dataframes instead of Numpy arrays.
  warn("Fetchers from the nilearn.datasets module will be "
INFO:nimare.utils:Shared path detected: '/Users/taylor/.nimare/nidm_21pain/'
INFO:nimare.utils:Shared path detected: '/Users/taylor/.nimare/nidm_21pain/'
/Users/taylor/Documents/tsalo/PyMARE/pymare/stats.py:23: RuntimeWarning: divide by zero encountered in true_divide
  w = 1. / (v + tau2)
/Users/taylor/Documents/tsalo/PyMARE/pymare/estimators/estimators.py:200: RuntimeWarning: divide by zero encountered in true_divide
  w = 1. / v
/Users/taylor/Documents/tsalo/nilearn/nilearn/datasets/struct.py:155: FutureWarning: Default resolution of the MNI template will change from 2mm to 1mm in version 0.10.0
  warnings.warn("Default resolution of the MNI template will change "
/Users/taylor/Documents/tsalo/nilearn/nilearn/_utils/niimg.py:61: UserWarning: Non-finite values detected. These values will be replaced with zeros.
  warn(
```

Here's the full log + traceback from numpy 1.21.1:

```
/Users/taylor/Documents/tsalo/nilearn/nilearn/datasets/__init__.py:92: FutureWarning: Fetchers from the nilearn.datasets module will be updated in version 0.9 to return python strings instead of bytes and Pandas dataframes instead of Numpy arrays.
  warn("Fetchers from the nilearn.datasets module will be "
[...]

Issue #274, opened 2020-06-25
Fill missing data in images with NaNs


tsalo:
This is related to https://github.com/neurostuff/PyMARE/issues/42, as well as to https://github.com/neurostuff/NiMARE/issues/231#issuecomment-633810456.

We can include this step in the `_preprocess_inputs` method in image-based meta-analysis estimators called by `fit()`.

@nicholst what do you think of this as a solution to your comment about varying degrees of freedom in IBMAs?

nicholst:
I’ve slipped into FSL’s sphere of avoiding implicit NA masking at all costs
and using explicit masks. But if you’re OK with NA masking go for it.
-- 
_________________________________________________________
Thomas Nichols, PhD
Professor of Neuroimaging Statistics
Nuffield Department of Population Health | University of Oxford
Big Data Institute | Li Ka Shing Centre for Health Information and Discovery
Old Road Campus | Headington | Oxford | OX3 7LF | United Kingdom
T: +44 1865 743590 | E: thomas.nichols@bdi.ox.ac.uk
W: http://nisox.org | http://www.bdi.ox.ac.uk

tsalo:
The only problem with explicit masks is that we have to pass a 2D array into PyMARE, which would increase complexity on NiMARE's side by a fair amount.

If we masked out studies with missing data for each voxel, then we would have to loop through unique combinations of studies _with_ data to identify matching voxels, build 2D arrays for those voxels, run the PyMARE estimator on each group separately, and then merge the results back into a map at the end.

nicholst:
I'm fine with it... so do what makes sense.

tsalo:
>Please forgive me if we've discussed this before, but have we considered (at least in the architecture if not the current implementation) the possibility of voxel-wise-varying sample size?
>
>It's now an edge case, but with larger and larger N (for an individual, non-meta analysis), you have the problem of ever eroding analysis mask since all voxels are required.  We are now building tools that allow the number of subjects contributing to an analysis vary by voxel (within limits, of course).
>
>_Originally posted by @nicholst in https://github.com/neurostuff/NiMARE/issues/231#issuecomment-633810456_

@nicholst I don't want to open a new issue about it, but is [Cutler, Radua, & Campbell-Meiklejohn (2018)](https://www.biorxiv.org/content/10.1101/457028v1) what you were talking about as a method for dealing with varying brain coverage across maps?

nicholst:
[...]

PR #544, opened 2021-07-22
[ENH] Do not apply IBMA methods to voxels with zeros or NaNs

tsalo:
Related to #542 and #274. The issue at hand is that PyMARE raises warnings when provided with arrays that are all zeros, while NiMARE's aggressive masking replaces any voxels with a zero in one or more studies with zeros across all studies. Our current goal is to eliminate that warning by only feeding PyMARE voxels with valid data. 

-
-

### DIFF
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -14,7 +14,7 @@ from nilearn._utils.niimg_conversions import _check_same_fov
 from nilearn.image import concat_imgs, resample_to_img
 
 from .results import MetaResult
-from .utils import get_masker
+from .utils import get_masker, reduce_masker
 
 LGR = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


nicholst:
I don't have enough insight to the potential side effects, but let me provide this input:  SPM & FSL (& I assume AFNI) work with a succession of masking requests and rules that, ultimately, produces an analysis mask.  (For example, in SPM, you can specify an absolute or relative % threshold that every subject's data must satisfy to be included; you can provide an explict analysis mask; and any zero-variance voxels are removed from the analysis; it is the intersection of all these rules that produces the fina analysis mask).

It seems to me this is simply an issue about at what stage the anlaysis mask is finalised.  If you find the present implimentation creating funny surprises for the user, maybe the analysis mask creation should be moved to an earlier stage.

tsalo:
Thanks @nicholst. I think these changes should operate similarly to the SPM approach then. I don't anticipate any issues, since the final mask should really be equivalent to the current implementation, with the small difference that voxels with zeros will be removed _before_ going to PyMARE instead of after, but if there do end up being issues I can focus on consolidating the masking earlier in the process.

nimare/base.py
```diff
@@ -389,6 +405,20 @@ def _preprocess_input(self, dataset):
...
+        # Further reduce image-based inputs to remove "bad" voxels
+        # (voxels with zeros or NaNs in any studies)
+        if "aggressive_mask" in self.inputs_.keys():
+            n_bad_voxels = (
+                self.inputs_["aggressive_mask"].size - self.inputs_["aggressive_mask"].sum()
+            )
+            if n_bad_voxels:
+                LGR.warning(
+                    f"Masking out {n_bad_voxels} additional voxels. "
+                    "The updated masker is available in the Estimator.masker attribute."
```
tsalo, line 417: This was a mistake. The masker is not updated directly.

PR #549, opened 2021-07-31
[DOC] Update citation for Enge et al. (2021)

alexenge:
Follow-up on  #515.

- Updated the citation for Enge et al. (2021) to the version now published in NeuroImage


tsalo:
Thanks!

PR #556, opened 2021-08-24
[FIX] Separate IJK calculation from coordinate space conversion

### LINKED ISSUE #555: Kernel Transformers use IJK blindly, without taking the masker's affine into account
## Summary

I was running an `ROIAssociationDecoder` on a Neurosynth Dataset built with v0.0.10rc2, using a target mask in MNI space with a (91, 109, 91) matrix. Because of recent changes to nilearn starting with 0.8.1, the default MNI brain mask is (99, 117, 95) instead of (91, 109, 91). This has caused problems with IBMAs (see #543), but I'm now seeing that it could be problematic for CBMAs when the target masker does not have the same affine as the Dataset's masker. 

Specifically, the kernel transformer just uses the IJK matrix indices, even though those only work for the Dataset's masker.

## Additional details

- NiMARE version: 0.0.10rc2

Run an ROIAssociationDecoder using a mask that doesn't have the same matrix size or affine as the Dataset's default mask.

The mask is of the amygdala, so I expected to see high correlations with terms that relate to the amygdala (e.g., "fear", "emotion").

### What actually happened?

The decoder ran, but the terms didn't make sense (e.g., "gradient", "35", "insights").

## Reproducing the bug

Load a Dataset with one mask, then apply a KernelTransformer to it using another mask, with a different affine. The results will be wrong.

tsalo:
Closes #555, though this may not be our permanent solution.

- Replace `np.bool` with `bool`, because it's being deprecated.
- Calculate IJK as needed within `KernelTransformer`, `Dataset.get_studies_by_mask`, and `MetaEstimator`.
- Fix bug in `Dataset.get_studies_by_label` in which, when the provided label was missing from the `Dataset.annotations` DataFrame, all Dataset IDs would be returned without so much as a warning.
- Replace basically every use of IJK with XYZ... hopefully I caught it all!
- Restructure the kernel tests to loop through parameters rather than having separate tests that are mostly duplicates of one another.
- Pin nilearn version to less than 0.8.0, to avoid the changes to the MNI152 template.

### DIFF
--- a/nimare/dataset.py
+++ b/nimare/dataset.py
@@ -648,7 +648,7 @@ class Dataset(NiMAREBase):
         mask_ijk = np.vstack(np.where(mask.get_fdata())).T
         distances = cdist(mask_ijk, dset_ijk)
         distances = np.any(distances == 0, axis=0)
-        found_ids = list(self.coordinates.iloc[distances]["id"].unique())
+        found_ids = list(self.coordinates.loc[distances, "id"].unique())
         return found_ids
 
     def get_studies_by_coordinate(self, xyz, r=20):
@@ -673,5 +673,5 @@ class Dataset(NiMAREBase):
         assert xyz.shape[1] == 3 and xyz.ndim == 2
         distances = cdist(xyz, self.coordinates[["x", "y", "z"]].values)
         distances = np.any(distances <= r, axis=0)
-        found_ids = list(self.coordinates.iloc[distances]["id"].unique())
+        found_ids = list(self.coordinates.loc[distances, "id"].unique())
         return found_ids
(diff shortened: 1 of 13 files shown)


nimare/dataset.py
```diff
@@ -638,11 +642,11 @@ def get_studies_by_mask(self, mask):
 
         dset_mask = self.masker.mask_img
         if not np.array_equal(dset_mask.affine, mask.affine):
-            from nilearn.image import resample_to_img
+            LGR.warning("Mask affine does not match Dataset affine. Assuming same space.")
 
-            mask = resample_to_img(mask, dset_mask, interpolation="nearest")
+        dset_ijk = mm2vox(self.coordinates[["x", "y", "z"]].values, mask.affine)
```
tsalo, line 647: `resample_to_img` assumes the images are already aligned (meaning we never had a method for using masks in different spaces), so I think this makes sense.

tsalo:
Most things look good, but the ROIAssociationDecoder is still giving me weird results when I use a target mask in the old 91x109x91 matrix. When I plot the overlay of the ROI on the default brain and on the old brain, it looks good in both cases, so it _is_ in the same space. I'm a little stumped.

PR #562, opened 2021-09-16
[REF] Reduce memory in CBMA Estimators

tsalo:
Closes None, but I'm trying to find a solution for Jennifer Barredo's memory issues, as reported in https://neurostars.org/t/nimare-allocating-resources-to-avoid-memory-crashes/20251

- Delete variables as they become unnecessary within CBMA (especially MKDAChi2) Estimators' FWE correction methods. I don't think this will make a huge difference, but it could help.


tsalo:
Local profiling indicated that at least the changes to MKDAChi2 helped.

PR #567, opened 2021-09-26
[TST] Test minimum versions of dependencies

### LINKED ISSUE #566: Test minimum versions of dependencies
## Summary

Nilearn has a cool CI step that runs its tests on the minimum allowed version of all of its requirements. We can do this in NiMARE by creating a new dependency group in `info.py` and pinning minimum versions in each. I'll probably just run it with the minimum supported Python version, since we already have a lot of tests being run.

tsalo:
Closes #566.

- Add new dependency group with core requirements pinned to the oldest versions I think we should support.
- Run tests using Python 3.6 with the new dependency group in CI.


tsalo:
`nilearn` requires `matplotlib` to access the `reporting` module, but it's not a NiMARE dependency. I'll probably add it to our info.py as a patch for now.

PR #571, opened 2021-09-30
[TST] Reorganize tests for speed

tsalo:
Closes None.

- Move the smoke tests in `test_estimator_performance` into a new job only run on Python 3.6.
- Move utility functions in `test_estimator_performance` into `tests.utils`.
- Take coverage reports from unit tests, performance tests, and performance smoke tests, and merge them before uploading.


tsalo:
Not a huge difference, but it's better than nothing.

PR #578, opened 2021-10-08
[ENH] Add test steps and explicit support for Python 3.9

### LINKED ISSUE #575: Add test steps and explicit support for Python 3.9
## Summary

We currently only explicitly support Python versions 3.6, 3.7, and 3.8. I'd like to extend that support to 3.9.

## Next steps

1. Add 3.9 to the classifiers within `info.py`.
2. Create new job for 3.9 in the CircleCI config file. You can just copy the job for 3.8 and change the relevant variables/names.
3. Add the new job to the CircleCI workflow section in the config file.

JulioAPeraza:
Closes #575.

-  Support Python version 3.9. Create and add a new job for 3.9 in the CircleCI config file.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
           paths:
               - src/NiMARE
       - restore_cache:  # ensure this step occurs *before* installing dependencies
-          key: deps9-{{ checksum "nimare/info.py" }}-{{ checksum "setup.py" }}
+          key: deps10-{{ checksum "nimare/info.py" }}-{{ checksum "setup.py" }}
       - run:  # will overwrite NiMARE installation each time
           name: Generate environment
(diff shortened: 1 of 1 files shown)


JulioAPeraza:
Updating the version fixed the issue. Should we use the latest release (`setuptools==58.2.0`)?

tsalo:
If it still works for Python 3.6, then yes. Otherwise, I'd like to determine the most recent version that works for all Python versions we support.

JulioAPeraza:
It seems like numpy has not released a wheel for Python 3.10.

JulioAPeraza:
I would add 3.9 support in this PR and tackle 3.10 later. The numpy wheel may be out by next month, but any issues with a different package in 3.10 may delay the 3.9 support.

Issue #576, opened 2021-10-08
convert_neurovault_to_dataset() - sample_size and IndexError 


gnodix:
Hi everyone, 

I'm trying to run a meta analysis strip to extract statistical map from NeuroVault. 

The script is : 
```python
from nimare.io import convert_neurovault_to_dataset

# The specific collections I would like to download group level
# statistical maps from

collection_ids = (2525, 9616, 3935, 3954, 3952, 6150)

# A mapping between what I want the contrast(s) to be
# named in the dataset and what their respective group
# statistical maps are named on neurovault
contrasts = {
    "dyad synchrony": (
    	"Lee 2017"
    	"van Baar 2021"
    	"Raykov 2018"
    	"Lee 2018"
    	"Lee 2017"
    	"Kerr 2020"
    )
}

# Convert how the statistical maps on neurovault are represented
# in a NiMARE dataset.
map_type_conversion = {"Z map": "z", "T map": "t"}

dset = convert_neurovault_to_dataset(
    collection_ids,
    contrasts,
    img_dir=None,
    map_type_conversion=None,
)

from nimare.transforms import ImageTransformer

# Not all studies have Z maps!
#print(dset.images["z"])

z_transformer = ImageTransformer(target="z")
dset = z_transformer.transform(dset)

# All studies now have Z maps!
#print(dset.images["z"])

from nilearn.plotting import plot_stat_map

from nimare.meta.ibma import Fishers

meta = Fishers()

meta_res = meta.fit(dset)

plot_stat_map(meta_res.get_map("z"), threshold=3.3)
# The result may look questionable, but this code provides
# a template on how to use neurovault in your meta analysis.
```

However, the terminal give me this response: 
```
Traceback (most recent call last):

  File "/Users/ginodiez/Desktop/Cogmaster/M1/LNC2/meta/statmap_neurovault.py", line 29, in <module>
    dset = convert_neurovault_to_dataset(

  File "/Users/ginodiez/opt/anaconda3/lib/python3.8/site-packages/nimare/io.py", line 544, in convert_neurovault_to_dataset
    sample_size = sample_sizes[0]

IndexError: list index out of range
```
I don't understand the "sample_size" error and how to solve it (there is no solution on the web for know!)

Best, 
Gino 

- NiMARE version: 0.0.9
- Mac OS Big Sur 11.2.3

gnodix:
Hi @jdkent 

Think you for your help! 
[...]

PR #580, opened 2021-10-12
[FIX] only download group maps when creating dataset and raise error if no images are found for a contrast

### LINKED ISSUE #579: single subject maps downloaded from neurovault
## Summary

when using `neurovault_to_dataset`, if the collection has single subject maps the function may use those maps in the dataset and not the group level map. example: https://neurovault.org/api/collections/11303/images/?format=json

## Additional details

- NiMARE version: 0.0.9

Run neuroscout meta-analyses

Robust meta-analyses with group level maps

### What actually happened?
bad meta-analysis maps

jdkent:
Closes #579 
ref #576 

- check if map is group-level
  - skip map if it is not group level
- check if any maps are found for a contrast
  - raise error if no images are found
- add test

### DIFF
--- a/nimare/tests/test_io.py
+++ b/nimare/tests/test_io.py
@@ -178,9 +178,7 @@ def test_convert_neurosynth_to_json_smoke():
         (
             {
                 "collection_ids": (8836,),
-                "contrasts": {
-                    "crab_people": "cannot hurt you because they do not exist"
-                },
+                "contrasts": {"crab_people": "cannot hurt you because they do not exist"},
             }
         ),
     ],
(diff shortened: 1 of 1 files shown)


tsalo:
There seems to be a new problem with isort. It also came up in https://github.com/ME-ICA/tedana/pull/815. I haven't looked into it yet, but it might be a matter of pinning the version of `flake8-isort`. The `build_docs` error is random and already documented in #418.

I will try to look at the actual code soon.

tsalo:
@jdkent it looks like the isort issue is fixed, although the current cache will have the bad version. If you update `info.py` or the CircleCI cache strings to trigger a new build, I think CI will pass.

tsalo:
@jdkent all of the downstream jobs from the make_env_py36 job need to have the cache updated too. However, now that #578 is merged, I think you can just rebase and it should pass CI.

PR #597, opened 2021-11-05
Replace multiprocessing with joblib for parallelization

tsalo:
Closes None.

- Use `joblib` instead of `multiprocessing` for parallelization.
- Drop if/else statements based on number of cores. Just use `joblib` regardless.
    - This may slow things down when `n_cores` is 1, but I want to be sure before using more complicated code.
- Change default number of cores to 1 in all tools that have `n_cores` as a parameter.
- Hopefully address issues with unclosed memmaps (see #631).

To do:

- [x] Drop tmpdir change now in #599.
- [x] Do some profiling.
    - At least on MKDADensity, without correction, joblib and multiprocessing perform the same.
- [x] Change default `n_cores` to 1 throughout library.


tsalo:
[`Pool.imap`'s documentation](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.imap) says the following:

>For very long iterables using a large value for _chunksize_ can make the job complete **much** faster than using the default value of 1.

I'm not sure what constitutes a "long iterable", but I can't help but wonder if using a larger chunksize value might be worth it.

tsalo:
`Joblib` can work with memmaps: https://joblib.readthedocs.io/en/latest/parallel.html#working-with-numerical-data-in-shared-memory-memmapping

PR #599, opened 2021-11-05
Use tmpdir for memmap files instead of the NiMARE data directory

### LINKED ISSUE #569: Use temporary directory for memory-mapped files
## Summary

This stems from #565. Currently, we write temporary files to the NiMARE data directory (`~/.nimare`), but using an actual temporary directory (defaulting to `$TMPDIR`) would leverage faster storage that is common on HPCs.

My only question is whether it would be better to adopt a straightforward `$TMPDIR` check _or_ look into how nilearn handles things, with `joblib` and caching (see [here](https://github.com/nilearn/nilearn/blob/e86ab53394674196493759c71f3ffbefd97da51a/nilearn/_utils/cache_mixin.py)).

@jdkent are you familiar with how nilearn handles this?

tsalo:
Closes #569, but I still want to look into nilearn's approach to memory management. This PR reverts parts of #460, and I'm hoping there won't be problems with the NiMARE manuscript code.

This mostly spun out from #597 as I figured it would help there as well.

- Use default temporary directory for memmap files instead of NiMARE data directory.


JulioAPeraza (approved):

tsalo:
I'm going to bite the bullet and merge this. Hopefully it won't cause problems for the Jupyter book.

PR #600, opened 2021-11-06
[REF] Add major classes/functions to parent namespaces

### LINKED ISSUE #388: Add main classes/functions to parent namespaces
It's a bit tedious to have to, e.g., `from nimare.meta.ale import ALE`. it forces users to have a grasp of the internal module structure, and there's probably no reason for it. I'd suggest including key objects located two levels below root in their parent namespaces. I.e., `from nimare.meta import MKDADensity`; `from nimare.decode import gclda_encode`, etc.

tsalo:
Closes #388.

- Add major classes and functions to parent module namespaces.

### DIFF
--- a/nimare/annotate/__init__.py
+++ b/nimare/annotate/__init__.py
@@ -1,5 +1,21 @@
 """Automated annotation tools."""
 
 from . import cogat, gclda, lda, text, utils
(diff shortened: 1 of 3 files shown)


tsalo:
The Sphinx interlinking issue is still there (and apparently has been for a while without me noticing), but the circular import issue seems to be resolved, so I'm going to merge this.

PR #609, opened 2021-12-07
Implement cluster mass-based cluster-level Monte Carlo correction for CBMA algorithms

### LINKED ISSUE #608: Add cluster mass option for Monte Carlo FWE correction methods
## Summary

I was reading the [FSL randomise documentation](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/Randomise/UserGuide) and they referred to cluster size-based inference as "so passé!" Apparently, cluster mass-based inference is preferable, based on the documentation and on [Zhang, Nichols, & Johnson (2009)](https://dx.doi.org/10.1016%2Fj.neuroimage.2008.08.017), which I, embarrassingly, hadn't heard of.

It should be pretty easy to calculate maximum cluster mass in each Monte Carlo permutation and retain it in a new null distribution, but I want to make sure that a cluster mass option would be useful before implementing it.

@nicholst would you be willing to weigh in on this?

## Additional details

Presumably, this would be a better metric for cluster-level inference.

## Next steps

1. Discuss. If the proposed option is not useful, then close this issue. Otherwise, implement the other steps.
2. Calculate "cluster mass" within the Monte Carlo FWE correction methods.
3. Retain maximum cluster mass in a new null distribution.
4. Calculate cluster masses for the statistical map.
5. Output the cluster mass-based significance map in the resulting MetaResult.

tsalo:
Closes #608.

- Add new cluster mass-based Monte Carlo FWE correction, which produces new `<z|logp>_desc-mass_level-cluster_corr-FWE_method-montecarlo` maps.
- Rename `<z|logp>_level-cluster_corr-FWE_method-montecarlo` maps to `<z|logp>_desc-size_level-cluster_corr-FWE_method-montecarlo`.

To do:

- [x] Add citation for cluster mass approach (Bullmore 1999).
- [x] Add versionchanged directive to FWE method.
- [x] Describe cluster mass approach in docstring.
- [x] Check results.


tsalo:
RTD is failing due to a problem with m2r and mistune.

tsalo:
Renaming the existing cluster size-based maps counts as a breaking change.

tsalo:
The results look the same as cluster size for an example Dataset with 17 studies and FWE correction with 10000 iterations.

image

Also, the cluster size and mass null distributions were correlated at about 0.9, and the cluster mass and max ALE value distributions were correlated at about 0.91. I think I can conclude that the results are what we'd expect.

PR #616, opened 2021-12-10
Flag peaks2maps for removal in 0.0.13/0.1.0

### LINKED ISSUE #615: Deprecate Peaks2MapsKernel
## Summary

I'm thinking that we might want to remove Peaks2Maps from NiMARE. I believe that @tyarkoni has brought this up in the past, but I pushed back because I thought it was cool. I still do think it's a cool tool, but it doesn't seem to get much use and it's just more code for me to maintain, so...

Before I do remove it, I want to make sure that it isn't important for anything. @adelavega are there any plans for Peaks2Maps in Neurosynth 2.0?

## Additional details

I've only ever gotten one NeuroStars question about Peaks2Maps (https://neurostars.org/t/issue-generate-modeled-activation-maps-with-peaks2maps/17656) and there haven't been any bug reports or feature requests relating to it from users. Therefore, I assume that Peaks2Maps doesn't get much use, but it does come with a maintenance cost. Tensorflow changes have caused problems in the past (e.g., #288). Those problems have been fairly straightforward to solve, but dealing with them might be completely unnecessary if no one is using this tool.

## Next steps

1. Discuss.
2. Deprecate `Peaks2MapsKernel` and all associated functions. Probably remove at 0.0.13 or 0.1- whichever comes first.

tsalo:
Closes #615, though I won't mark this as ready to review this until we have a consensus in that issue.

- Add "deprecated" directive to docstrings of functions and classes related to the Peaks2Maps algorithm.
- Add a deprecation warning to Peaks2MapsKernel.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -9,6 +9,9 @@ changelog:
     - title: 🎉 Exciting New Features
       labels:
         - enhancement
+    - title: 👎 Deprecations
+      labels:
+        - deprecation
     - title: 🐛 Bug Fixes
       labels:
         - bug
--- a/README.md
+++ b/README.md
@@ -26,11 +26,7 @@ pip install nimare
 
 ### Local installation (development version)
 ```
-pip install git+https://github.com/neurostuff/NiMARE.git#egg=nimare[peaks2maps-cpu]
-```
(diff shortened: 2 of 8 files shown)


tsalo:
No objections have been raised, so I'm merging this.

PR #618, opened 2021-12-13
Create Annotation and Annotator classes

### LINKED ISSUE #617: Refactor storage of annotations in Datasets
## Summary

Datasets currently store annotations as an attribute `.annotations`, which is a pandas DataFrame with one column for each label. Groups of annotations (e.g., Neurosynth TF-IDF values vs. LDA topics) are distinguished with prefixes in the column names. Per today's call, a better structure would be to have `Dataset.annotations` be a dictionary containing `Annotation` objects. The keys to the dictionary would be the annotation group names.

`Annotation` objects will have the following attributes/methods (names are up for debate):

- `.study_term_weights`: A pandas DataFrame containing the annotation group's labels and weights. This is the main attribute of interest for NeuroStore.
- `.other_distributions`: A dictionary containing other distributions (e.g., topic-term weights, topic-voxel weights), whether in numpy.ndarray or pandas.DataFrame form.
- `.metadata`: BIDS-like metadata, possibly with top terms for topic models. Also software info and provenance.
- `.to_filename()`: To save the study_term_weights attribute as a CSV file. We discussed having JSON-based storage, but it seems like CSVs would be easier for NeuroStore to work with here.

Annotators will then operate more like Estimators than Transformers. Namely, they will ingest Datasets and return Annotation objects (as proposed by @jdkent in #488). Users will have to add the Annotations to their Datasets themselves.

## Next steps

1. Create new Annotation class in #607.
2. Refactor Dataset.annotations and associated conversion functions.
3. Refactor Dataset search methods and decoding functions/classes.

tsalo:
Closes #617, but far from ready for review. Some of this PR will overlap with #607.

- Replace Dataset.annotations DataFrame with a dictionary of name-Annotation pairs.
- Create Annotation class for storing annotations.
- Create Annotator class for creating Annotations from Datasets.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -330,3 +330,4 @@ For more information about fetching data from the internet, see :ref:`fetching t
    base.MetaEstimator
    base.Transformer
    base.Decoder
+   base.Annotation
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -574,20 +574,48 @@ class Decoder(NiMAREBase):
 
 
 class Annotator(NiMAREBase):
-    """Base class for annotators."""
+    """Base class for annotators in :mod:`nimare.annotate`.
 
-    def transform(self, dset):
-        """Apply an annotation method to a Dataset.
+    Annotators operate like Transformers in that they ingest Datasets and output modified Datasets.
+    One difference between Annotators and Transformers is that Annotators retain extra information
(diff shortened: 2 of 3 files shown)


tsalo:
Closing this. We can circle back to it after the NIMADS PR is merged.

Issue #399, opened 2020-11-16
Dataset cannot be copied if masker comes from a gzipped file


tsalo:
When I try to use the `ale_2mm` target (e.g., when loading a Sleuth dataset), the resulting Dataset cannot be copied due to the following error:

```python
TypeError: cannot serialize 'IndexedGzipFile' object
```

After digging a bit, if I just uncompress the associated mask file, everything works fine. I believe it's an issue I came across a while ago with `nibabel` (see https://github.com/cosanlab/nltools/issues/281). I thought the issue was resolved in the newest versions of `nibabel`, but I may have just misunderstood what was causing it. In any case, I think we need to either enforce uncompressed files for our masks or find some safer way of loading them in our masker objects.

BTW, converting to a masker with `utils.get_masker()` doesn't help.

tsalo:
As a temporary solution, we could incorporate a check into the `get_masker()` function that determines if an input is from a gzipped file and raises a warning if so?

tsalo:
@tyarkoni Based on some testing in #400 and the original cosanlab issue, it looks like the bug is eliminated by not having `indexed_gzip` installed. This is why I encountered the bug on my laptop and not in CircleCI.

Do you have any thoughts on what we should do? Add `indexed_gzip` as a dependency and work around the bug by default or add some kind of check to see if it's installed and raise a warning if the right conditions are met?

tyarkoni:
I don't feel I know enough about the packages in question to have an informed opinion, so I'll defer to your judgment on this. 
The only thing I do feel pretty strongly about is that things should still work fine without any user intervention—a warning is fine, but no additional configuration should be necessary. Assuming that's what's on the table (i.e., implicitly unpack the gzipped mask to a temporary location and then use that instead, with a warning), that seems fine. But if this isn't already a known issue on the NiBabel repo, might be worth raising it for discussion there.

tsalo:
@rmarkello shared the following trick that we can use:

```python
import gzip
import nibabel as nib

nib.openers.HAVE_INDEXED_GZIP = False
nib.openers.IndexedGzipFile = gzip.GzipFile
```

[...]

PR #621, opened 2021-12-14
Replace Nilearn templates with ones stored in resources folder and unset maximum Nilearn version

### LINKED ISSUE #611: Replace fetched templates/masks with ones downloaded from TemplateFlow
## Summary

The main reason I've pinned an earlier version of nilearn is that in 0.8.0 they replaced the MNI template with a new one that is less compatible with the 21 pain dataset (see #543). If we use our own MNI templates (downloaded from TemplateFlow), then there shouldn't be a problem with newer versions of nilearn.

## Next steps

1. Identify the appropriate template(s) from TemplateFlow.
2. Download the templates.
3. Update `nimare.utils.get_template` to use the files stored in `resources` instead of using nilearn.
4. Remove the restrictions on nilearn version from `info.py` introduced in #556.
5. Remove the restrictions on matplotlib version introduced in #610.

tsalo:
Closes #611. The breaking change here is that the `mask="gm"` option no longer works. TemplateFlow did not have tissue type probability maps for the desired templates. Additionally, while the new templates should be essentially the same as the old ones, the affines are slightly different, so some tests and examples had to be adjusted.

- Download `tpl-MNI152NLin6Asym` templates from TemplateFlow and store them in the resources folder. Use these templates instead of nilearn functions for `nimare.utils.get_template`.
- Remove the restriction on maximum nilearn version. The main reason we had a maximum version pinned is that the MNI template introduced in ~0.8 (corresponding to `MNI152NLin2009cAsym`, I believe) had different dimensions than the one currently used in NiMARE, and those dimension differences meant that images from the NIDM pain dataset had to be resampled, which in turn caused problems for our examples and tests. 
    - We originally set the maximum version in #556.
- gzip the existing ALE brain mask image, since #399 has been resolved.


tsalo:
The NeuroVault example is currently failing, and I think the problem is that the new `mni152_2mm` template has the following affine:

```
[[   2.    0.    0.  -90.]
 [   0.    2.    0. -126.]
 [   0.    0.    2.  -72.]
 [   0.    0.    0.    1.]]
```

While the old nilearn one has the following affine:

```
[[  -2.    0.    0.   90.]
 [   0.    2.    0. -126.]
 [   0.    0.    2.  -72.]
 [   0.    0.    0.    1.]]
```

I believe this corresponds to a simple flip in the internal array that is cancelled out by the shifted negative value, which means that the maps should be completely compatible. However, `_check_same_fov` is too restrictive to allow it.

I'm not sure what to do about it...

nimare/utils.py
```diff
@@ -33,53 +32,63 @@ def get_resource_path():
...
     space : {'mni152_1mm', 'mni152_2mm', 'ale_2mm'}, optional
-        Template to load. Default is 'mni152_1mm'.
-    mask : {None, 'brain', 'gm'}, optional
-        Whether to return the raw template (None), a brain mask ('brain'), or
-        a gray-matter mask ('gm'). Default is None.
+        Template to load. Default is 'mni152_2mm'.
+        The options are:
+
+            -   mni152_1mm: The MNI152NLin6Asym template at 1mm3 resolution,
+                downloaded from TemplateFlow. The shape of this template is X voxels.
```
tsalo, line 46:
```suggestion
                downloaded from TemplateFlow. The shape of this template is 182x218x182 voxels.
```

I don't want to wait around for CI to pass again, so once it passes I'm going to commit this change and force the merge through.

nimare/utils.py
```diff
@@ -33,53 +32,63 @@ def get_resource_path():
...
+            -   mni152_2mm: The MNI152NLin6Asym template at 2mm3 resolution,
+                downloaded from TemplateFlow. The shape of this template is 91x109x91 voxels.
+            -   ale_2mm: The template used is the MNI152NLin6Asym template at 2mm3 resolution,
+                but if ``mask='brain'``, then a brain mask taken from GingerALE will be used.
+                The brain mask corresponds to GingerALE's "more conservative" mask.
+                The shape of this template is 91x109x91.
```
tsalo, line 52:
```suggestion
                The shape of this template is 91x109x91 voxels.
```

nimare/utils.py
```diff
@@ -33,53 +32,63 @@ def get_resource_path():
 def get_template(space="mni152_2mm", mask=None):
     """Load template file.
 
+    .. versionchanged:: 0.0.11
+        Remove the mask="gm" option.
+        Replace the nilearn templates with ones downloaded directly from TemplateFlow.
```
tsalo, line 37:
```suggestion
    .. versionchanged:: 0.0.11
        - Remove the ``mask="gm"`` option.
        - Replace the nilearn templates with ones downloaded directly from TemplateFlow.
```

PR #649, opened 2022-02-22
Add FocusCounter diagnostic tool

### LINKED ISSUE #595: Focus-count post-meta-analysis method
## Summary

We need to implement the focus-count method as a class in `nimare.diagnostics`.

## Additional details

This should be equivalent to what GingerALE does.

## Next steps

1. Class
2. Tests
3. Documentation

tsalo:
Closes #595.

- Add a new class to `nimare.diagnostics`, `FocusCounter`, that characterizes CBMA results based on the number of foci in each experiment within the Dataset that contributed to each cluster.
- Work around bug reported in #651 specifically for `Jackknife` diagnostic by operating on a copy of the Estimator within the study loop.


tsalo:
Currently failing because Jackknife has issues with the IBMA estimator, which I'm guessing is due to changes in the aggressive mask based on the subset of experiments in the meta-analysis.

tsalo:
The problem is that, when you fit the same IBMA estimator multiple times, each time the aggressive mask is updated based on the new data _and_ the existing aggressive mask using a logical or, meaning that good voxels in an older dataset will be retained when fitting to the new dataset. This causes problems when some of those older voxels are not good for the new dataset.

This problem wasn't already apparent because all of our Jackknife tests on IBMAs had good data in all of the voxels, I believe. Or at least the aggressive mask didn't change from one subset of studies to another.

I'm not sure what the best solution is. Here are some ideas:

- Automatically replace the aggressive mask when fitting an estimator to a new dataset. I think this would just be a step in `_preprocess_inputs` to delete that key from the `inputs_` dictionary if it exists at the beginning of the preprocessing.
- Add an argument to amend or replace the mask to the IBMA estimator initialization, so users can decide.
- Just make a copy of the estimator within the Jackknife `_transform` method so that it amends the original aggressive mask, but doesn't iteratively update masks from different subsets of the dataset.

tsalo:
For the sake of this PR, I will probably just make a copy, since that's probably a good idea in general. I will also open a separate issue about the aggressive mask bug.

PR #650, opened 2022-02-26
Support cluster-level Monte Carlo FWE correction in the MKDAChi2 Estimator

tsalo:
Closes None.

- Implement cluster-level FWE correction in the MKDAChi2 estimator.
- Fix voxel-level FWE correction in the MKDAChi2 estimator.
     - It looks like the FWE null distributions really shouldn't be treated as two-sided tests, which is how the vFWE was being done. The FWE null distribution reflect maximum values across both negative and positive values/clusters, so we need to take the absolute values and compare those in a one-sided way to the null distributions before re-signing them.


tsalo:
For some reason the `threshold` parameter isn't having any effect on the `plot_stat_map` figures, but I don't think it's a problem for the new feature.

tsalo:
I may have to revisit this based on https://github.com/neurostuff/NiMARE/pull/655#issuecomment-1073029709.

PR #657, opened 2022-03-21
Stop storing MetaResults as attributes of fitted Estimators

### LINKED ISSUE #634: Stop storing MetaResult as attribute of fitted Estimator
## Summary

In working on #633, I have come to the conclusion that it's a little odd to store MetaResults in Estimators, when we're also storing the Estimators in the MetaResults. I think it would make more sense for MetaResults to retain their Estimators as attributes (which is necessary for Estimator-specific correction methods, like `FWECorrector(method="montecarlo")`), but to drop the MetaResult attribute from the fitted Estimator.

## Additional details

The main benefit is that it would simplify how users interact with NiMARE's primary classes. On the dev side of things, it would make it easier to keep track of where and when modified versions of each of these objects are being used.

## Next steps

1. Remove the `.results` attribute from Estimators.
2. Update all examples and tests.
3. Make it extremely clear to users that they will need to collect the _output_ of the Estimator when it's being fitted instead of grabbing the `.results` attribute later.

tsalo:
Closes #634.

- Return `results` from `Estimator.fit`, but do not store the object as an attribute of the `Estimator`.
- Ditto for `Decoder`s.
- Update tests and examples accordingly.

### DIFF
--- a/examples/02_meta-analyses/06_plot_compare_ibma_and_cbma.py
+++ b/examples/02_meta-analyses/06_plot_compare_ibma_and_cbma.py
@@ -47,9 +47,12 @@ dset = coord_gen.transform(dset)
 # ALE (CBMA)
 # -----------------------------------------------------------------------------
 meta_cbma = ALE()
-meta_cbma.fit(dset)
+cbma_results = meta_cbma.fit(dset)
 plot_stat_map(
-    meta_cbma.results.get_map("z"), cut_coords=[0, 0, -8], draw_cross=False, cmap="RdBu_r"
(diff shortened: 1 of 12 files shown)


jdkent (approved):
sorry if this was waiting on me, I agree dropping the metaresult from the estimator is cleaner, especially when iterating different datasets with the same estimator. and the code looks good to me!

jdkent:
this also appears to help with a memory leak issue where the deepcopy of the estimator is being added to memory every time the result is created.

tsalo:
Thanks @jdkent. Merging.

PR #659, opened 2022-03-24
Reduce memory usage of CBMAEstimator's FWE correction

tsalo:
Closes None. A user reported memory errors when running an ALE with voxel-level FWE correction on a ~2k-study subset of the NeuroQuery database. I've been trying to figure out what the problem is and how to fix it, and this is the best I've come up with so far.

- Fully support a `vfwe_only` option in `correct_fwe_montecarlo`, even when the null method isn't `montecarlo`.
- Determine the random coordinates for each permutation _within_ the permutation function, rather than defining all of them at once and then splitting them up for the loop.
    - I am pretty sure that this will slow things down, since the previous approach was something we optimized for speed. Unfortunately, it just doesn't scale well for very large Datasets.

### DIFF
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -464,7 +464,14 @@ class CBMAEstimator(MetaEstimator):
             "histweights_level-voxel_corr-fwe_method-montecarlo"
         ] = histweights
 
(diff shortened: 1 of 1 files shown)


jdkent:
could be helpful for profiling: https://github.com/bloomberg/memray

tsalo:
Yeah, I think it can be closed.

PR #662, opened 2022-04-19
Remove extraneous `copy()` statements

### LINKED ISSUE #661: Memory usage: Copying ma_maps to create ma_values
## Summary

copying `ma_maps` to create `ma_values` is a high memory operation, is this being done as a precaution or because `ma_values` is actually transformed in place at some point?

## Additional details

- NiMARE version: 0.11
example:
https://github.com/neurostuff/NiMARE/blob/470b8d351b3f6905147f61035d5235f8698fa251/nimare/meta/cbma/mkda.py#L1077-L1078

jdkent:
Closes #661

- try not to copy numpy arrays when not necessary

### DIFF
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -504,7 +504,7 @@ class CBMAEstimator(MetaEstimator):
         iter_max_value = np.max(iter_ss_map)
 
         # Cluster-level inference
-        iter_ss_map = self.masker.inverse_transform(iter_ss_map).get_fdata().copy()
+        iter_ss_map = self.masker.inverse_transform(iter_ss_map).get_fdata()
         iter_max_size, iter_max_mass = _calculate_cluster_measures(
             iter_ss_map, voxel_thresh, conn, tail="upper"
         )
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -563,7 +563,6 @@ class CBMAEstimator(Estimator):
             iter_max_size, iter_max_mass = _calculate_cluster_measures(
                 iter_ss_map, voxel_thresh, conn, tail="upper"
             )
-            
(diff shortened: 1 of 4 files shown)


nimare/meta/kernel.py
```diff
@@ -152,7 +152,7 @@ def transform(self, dataset, masker=None, return_type="image"):
                     elif return_type == "image":
                         return [nib.load(f) for f in files]
                     elif return_type == "dataset":
-                        return dataset.copy()
+                        return dataset
```
tsalo, line 155: In this case, I believe we do want to return a copy of the Dataset.

adelavega:
@jdkent I fixed a merge conflict this had because of #676 

Looks like if you return the copy that @tsalo requests, we can merge this and it may further help memory

nimare/meta/kernel.py
```diff
@@ -154,7 +154,7 @@ def transform(self, dataset, masker=None, return_type="image"):
...
```
tsalo, line 157:
```suggestion
                        return dataset.copy()
```

nimare/meta/cbma/base.py
```diff
@@ -559,11 +559,11 @@ def _correct_fwe_montecarlo_permutation(
             iter_max_size, iter_max_mass = None, None
         else:
             # Cluster-level inference
-            iter_ss_map = self.masker.inverse_transform(iter_ss_map).get_fdata().copy()
+            iter_ss_map = self.masker.inverse_transform(iter_ss_map).get_fdata()
             iter_max_size, iter_max_mass = _calculate_cluster_measures(
                 iter_ss_map, voxel_thresh, conn, tail="upper"
             )
-
+
```
tsalo, line 566:
```suggestion

```

tsalo (approved):
I think this is ready to merge now.

PR #674, opened 2022-04-26
Replace relative imports with absolute ones

tsalo:
Closes None. Absolute imports are recommended by PEP8, and I think this will make it easier to identify any circular import problems as we start reorganizing base classes.

### DIFF
--- a/nimare/annotate/cogat.py
+++ b/nimare/annotate/cogat.py
@@ -5,11 +5,11 @@ import re
 import numpy as np
 import pandas as pd
 
-from .. import references
-from ..due import due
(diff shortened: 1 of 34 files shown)


tsalo:
@jdkent @adelavega @JulioAPeraza do any of you have any issues with me merging this?

jdkent (approved):
LGTM, I do not think any of the import statements are overly verbose written out explicitly
[relevant PEP8 section](https://peps.python.org/pep-0008/#imports)

adelavega:
Fine w/ me

Issue #671, opened 2022-04-22
RF Suggestion: Instance methods vs helper functions


adelavega:
Here's a potential suggestion (/discussion/note to self) regarding the complex abstractions in NiMARE:

I noticed there are many cases where at a level of abstraction there are instance methods that take `self` as an argument but don't use it. In other words, they are pure static methods that make sense for all children to inherit and hence are put at the level of abstraction.

One example (although I'm sure there are better ones): https://github.com/neurostuff/NiMARE/blob/83d9f5575e8c90031fd27734fdea1278ae3861c0/nimare/base.py#L77

One idea would be to remove these from classes and make them pure helper functions which may make it easy to know which methods are called by who and when. 

This would be more akin to how nilearn/sklearn avoid excessive abstractions

PR #675, opened 2022-04-26
Simplify organization of base classes

tsalo:
Closes None, but does touch on #671.

- Make `_check_ncores` a function in `nimare.utils` (see #671).
- Remove the `Transformer` base class.
    - It only had the `transform` abstract method.
- Move the `Decoder` base class to `nimare.decode.base`.
- Create a new `IBMAEstimator` base class, and move all IBMA-related code from `MetaEstimator` into that new class.
- Move all CBMA-related code from `MetaEstimator` into `CBMAEstimator`.
- Remove `MetaEstimator`.
- Rename the `_validate_input` method to `_collect_inputs`.
- Convert generic kwargs in `CBMAEstimator` and `IBMAEstimator` into actual keyword arguments, and raise a warning for any extraneous kwargs.


adelavega:
I will take a look as soon as I get good chance!

adelavega:
- +1 on moving `_check_ncores` to `utils`
- I'm on the fence about removing the `Transformer` class since it does serve the purpose of enforcing the `transform` method on all children.

adelavega (changes requested):
Looks good!

- +1 on moving `_check_ncores` to `utils`
- I'm on the fence about removing the `Transformer` class since it does serve the purpose of enforcing the `transform` method is implemented by all children. It does add one more layer of abstraction but it sort of makes sense to me to keep this one. I could go either way honestly, maybe we can err on the side of removing it since it doesn't add much.  
- +1 on moving `Decoder`
- +1 on separating CBMA and IBMA. If there is a concern that there is common logic that could crop up, we could always add add the `MetaEstimator` class back in, and use that to control common logic. But if it as its stands there's almost none, then I think its more parsimonious to remove.
- Should `_preprocess_input` be an abstractmethod in `Estimator`? It seems like all children implement it, and probably should. I suppose there could be an instance of an estimator that doesn't, but since they all do lets enforce it. 

An observation:

What simplifies understanding `Estimators` to me is that they all have a `fit` method which calls: `_validate_input`, then `_preprocess_input`, and then `_fit`-- in this order. These methods must always be implemented. 

To further simplify reading the code, lets make sure all children implement them in this order (i.e. in `CBMAEstimator` `_fit` comes before `_preprocess_input`, so let's reverse that), and that these three functions are the top three after `__init__`. 

This at least makes the major code flow easier to read. 

The other issue for me is that `CBMAEstimator` implements so many other methods that are called within `_fit` and other "primary" methods. It's possible we could pull some of these out of the class entirely if they can be defined as pure functions (i.e. don't need access to that many class internals), or otherwise simplified. Let's hold off for another PR until we profile though since it has implications for that.

adelavega:
Great. LGTM then!

PR #676, opened 2022-04-28
Reduce memory usage of `KernelTransformer.transform` and `meta.utils.compute_kda_ma`

JulioAPeraza:
Closes None.

- Replace the 4D array in the (M)KDA with a list of 3D sparse arrays (one for each experiment) to avoid the high memory peak in the `KernelTransformer.transform` and `meta.utils.compute_kda_ma`.

There is probably a better way to introduce the sparse array here, but the current proposal works well with almost no modification to the kernel transformer.

### DIFF
--- a/nimare/meta/cbma/mkda.py
+++ b/nimare/meta/cbma/mkda.py
@@ -4,6 +4,7 @@ import logging
 
 import nibabel as nib
 import numpy as np
+import sparse
 from joblib import Parallel, delayed
 from scipy import ndimage
 from scipy.stats import chi2
@@ -322,9 +323,19 @@ class MKDAChi2(PairwiseCBMAEstimator):
             maps_key="ma_maps1",
             coords_key="coordinates1",
             fname_idx=0,
+            return_type="sparse",
         )
         n_selected = ma_maps1.shape[0]
-        n_selected_active_voxels = np.sum(ma_maps1.astype(bool), axis=0)
+        n_selected_active_voxels = ma_maps1.sum(axis=0)
+
+        if isinstance(n_selected_active_voxels, sparse._coo.core.COO):
+            masker = dataset1.masker if not self.masker else self.masker
+            mask = masker.mask_img
+            mask_data = mask.get_fdata().astype(bool)
+
+            # Indexing the sparse array is slow, perfor masking in the dense array
+            n_selected_active_voxels = n_selected_active_voxels.todense().reshape(-1)
+            n_selected_active_voxels = n_selected_active_voxels[mask_data.reshape(-1)]
 
         # Close the memmap.
         # Deleting the variable should be enough, but I'd prefer to be explicit.
@@ -340,9 +351,13 @@ class MKDAChi2(PairwiseCBMAEstimator):
             maps_key="ma_maps2",
             coords_key="coordinates2",
             fname_idx=1,
+            return_type="sparse",
         )
         n_unselected = ma_maps2.shape[0]
-        n_unselected_active_voxels = np.sum(ma_maps2.astype(bool), axis=0)
+        n_unselected_active_voxels = ma_maps2.sum(axis=0)
+        if isinstance(n_unselected_active_voxels, sparse._coo.core.COO):
+            n_unselected_active_voxels = n_unselected_active_voxels.todense().reshape(-1)
+            n_unselected_active_voxels = n_unselected_active_voxels[mask_data.reshape(-1)]
 
         # Close the memmap.
         # Deleting the variable should be enough, but I'd prefer to be explicit.
@@ -399,6 +414,8 @@ class MKDAChi2(PairwiseCBMAEstimator):
         )
 
         del n_selected, n_unselected
+
+        # print(cells)
 
         pFgA_chi2_vals = two_way(cells)
 
--- a/nimare/meta/cbma/mkda.py
+++ b/nimare/meta/cbma/mkda.py
@@ -401,8 +401,6 @@ class MKDAChi2(PairwiseCBMAEstimator):
 
         del n_selected, n_unselected
 
-        # print(cells)
-
         pFgA_chi2_vals = two_way(cells)
 
         del n_selected_active_voxels, n_unselected_active_voxels
--- a/nimare/meta/cbma/mkda.py
+++ b/nimare/meta/cbma/mkda.py
@@ -322,6 +322,7 @@ class MKDAChi2(PairwiseCBMAEstimator):
         ma_maps1 = self._collect_ma_maps(
             maps_key="ma_maps1",
             coords_key="coordinates1",
+            return_type="sparse",
         )
         n_selected = ma_maps1.shape[0]
         n_selected_active_voxels = ma_maps1.sum(axis=0)
@@ -331,7 +332,7 @@ class MKDAChi2(PairwiseCBMAEstimator):
             mask = masker.mask_img
             mask_data = mask.get_fdata().astype(bool)
 
-            # Indexing the sparse array is slow, perfor masking in the dense array
+            # Indexing the sparse array is slow, perform masking in the dense array
             n_selected_active_voxels = n_selected_active_voxels.todense().reshape(-1)
             n_selected_active_voxels = n_selected_active_voxels[mask_data.reshape(-1)]
 
@@ -342,6 +343,7 @@ class MKDAChi2(PairwiseCBMAEstimator):
         ma_maps2 = self._collect_ma_maps(
             maps_key="ma_maps2",
             coords_key="coordinates2",
+            return_type="sparse",
         )
         n_unselected = ma_maps2.shape[0]
         n_unselected_active_voxels = ma_maps2.sum(axis=0)
--- a/nimare/meta/cbma/mkda.py
+++ b/nimare/meta/cbma/mkda.py
@@ -1,4 +1,5 @@
 """CBMA methods from the multilevel kernel density analysis (MKDA) family."""
+import gc
 import logging
 
 import nibabel as nib
@@ -332,6 +333,7 @@ class MKDAChi2(PairwiseCBMAEstimator):
             ma_maps1._mmap.close()
 
         del ma_maps1
+        gc.collect()
 
         # Generate MA maps and calculate count variables for second dataset
         ma_maps2 = self._collect_ma_maps(
@@ -349,6 +351,7 @@ class MKDAChi2(PairwiseCBMAEstimator):
             ma_maps2._mmap.close()
 
         del ma_maps2
+        gc.collect()
 
         n_mappables = n_selected + n_unselected
 
--- a/nimare/meta/utils.py
+++ b/nimare/meta/utils.py
@@ -356,7 +356,6 @@ def compute_kda_ma(
     # Preallocate coords and data array, np.concatenate is too slow
     n_coords = ijks.shape[0] * kernel.shape[1]  # = n_peaks * n_voxels
     coords = np.zeros((4, n_coords), dtype=int)
-    data = np.zeros(n_coords, dtype=type(value))
     temp_idx = 0
     for i, peak in enumerate(ijks):
         sphere = np.round(kernel.T + peak)
@@ -376,12 +375,15 @@ def compute_kda_ma(
 
             coords[0, chunk_idx] = np.full((1, n_brain_voxels), exp)
             coords[1:, chunk_idx] = sphere.T
-            data[chunk_idx] = np.full(n_brain_voxels, value)
 
             temp_idx += n_brain_voxels
+    coords = coords[:, :temp_idx]
 
     if not memmap_filename:
-        kernel_data = sparse.COO(coords, data, shape=kernel_shape, has_duplicates=sum_overlap)
+        if not sum_overlap:
+            coords = np.unique(coords, axis=1)
+        data = np.full(coords.shape[1], value)
+        kernel_data = sparse.COO(coords, data, shape=kernel_shape)
 
     return kernel_data
 
--- a/nimare/meta/utils.py
+++ b/nimare/meta/utils.py
@@ -5,8 +5,8 @@ import os
 import nibabel as nib
 import numpy as np
 import numpy.linalg as npl
-from scipy import ndimage
 import sparse
+from scipy import ndimage
 
 from .. import references
 from ..due import due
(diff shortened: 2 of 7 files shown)


nimare/meta/utils.py
```diff
@@ -359,19 +360,33 @@ def compute_kda_ma(
         idx = (np.min(sphere, 1) >= 0) & (np.max(np.subtract(sphere, shape), 1) <= -1)
         sphere = sphere[idx, :].astype(int)
         exp = exp_idx[i]
-        if sum_overlap:
-            kernel_data[exp][tuple(sphere.T)] += value
+        if memmap_filename:
```
tsalo, line 363: If we're going to start using sparse arrays, should we just drop memory-mapping entirely?

jdkent: I was leaning that to be another PR since that could be a decent amount of work, maybe the removal could be a PR to this PR?

tsalo: I'm fine tackling that separately. I'm just glad to know that we're planning to simplify things eventually.

JulioAPeraza:
# Profiler results:
## Using dense arrays:

| Dataset (`n_studies`) |  Peak Memory (MB)         | Time  (s)         |
| -------- | --------- | ----------- |
| Small (10)     | 107.87 | 2 |
| Medium (100) | 913.96 | 3.3 |
| Large (1000)     | 9064.67 | 27 |

## Using sparse arrays:

| Dataset (`n_studies`) |  Peak Memory (MB)         | Time (s)          |
| -------- | --------- | ----------- |
| Small (10)     | 64.13 | 3.6 |
| Medium (100) | 774.28 | 24 |
| Large (1000)     | 3677.94 | 212 |

tsalo:
The change in peak memory for large datasets is promising, but I'm a little surprised by how much slower it ends up being.

JulioAPeraza:
Yeah, It is also much slower than memory-mapped arrays:

| Dataset (`n_studies`) |  Peak Memory (MB)         | Time (s)          |
| -------- | --------- | ----------- |
| Small (10)     | 52.74 | 2 |
| Medium (100) | 210.08 | 5 |
| Large (1000)     | 2065.77 | 103 |

JulioAPeraza:
@adelavega It looks like the compression performed by `pydata/sparse` takes a lot of time. An alternative approach could be to reshape the MRI data arrays and use `scipy/sparse`, which seems to be more efficient in terms of memory and computation time.

nimare/meta/utils.py
```diff
@@ -344,7 +347,8 @@ def compute_kda_ma(
         # Use a memmapped 4D array
         kernel_data = np.memmap(memmap_filename, dtype=type(value), mode="w+", shape=kernel_shape)
     else:
-        kernel_data = np.zeros(kernel_shape, dtype=type(value))
+        # Use a list of 3D sparse arrays
+        kernel_data_lst = [None] * n_studies
```
adelavega, line 351: If I'm reading correctly the idea here it to use a list of 3D sparse arrays? 

Isn't it possible with `sparse` to construct a 4D array? In a way I believe that's one of the main advantages of `sparse`

nimare/meta/utils.py
```diff
@@ -359,19 +363,33 @@ def compute_kda_ma(
...
+            if i % chunk_size == 0:
+                kernel_data.flush()
         else:
-            kernel_data[exp][tuple(sphere.T)] = value
-
-        if memmap_filename and i % chunk_size == 0:
-            # Write changes to disk
-            kernel_data.flush()
+            if kernel_data_lst[exp] is None:
+                kernel_data = np.zeros(shape, dtype=type(value))
```
adelavega, line 375: I don't think we want to create a zeros array, as this will take up the memory footprint of `shape` in memory, even if its all zeros.

nimare/meta/utils.py
```diff
@@ -359,19 +363,33 @@ def compute_kda_ma(
...
+            else:
+                kernel_data = kernel_data_lst[exp].todense()
+            if sum_overlap:
+                kernel_data[tuple(sphere.T)] += value
+            else:
+                kernel_data[tuple(sphere.T)] = value
 
-    if squeeze:
-        kernel_data = np.squeeze(kernel_data, axis=0)
+            kernel_data_lst[exp] = sparse._compressed.GCXS.from_numpy(
```
adelavega, line 383: `kernel_data` is a dense_matrix composed of zeros at this point right? 
 
I think those use `sparse` matrices correctly we want to avoid dense matrices at any point, and instead build a sprase matrix from the bottom up. 

See: https://sparse.pydata.org/en/stable/construct.html

I think we want to create from coordinates and data, rather than from an empty numpy matrix. This should save memory & processing time.

adelavega (changes requested):
I think we need to build up sparse arrays from coordinates directly rather than using numpy dense arrays as an intermediary.

When you construct a `np.zeros` arrray, you are still taking up the memory footprint of the size of that array. I think if we want to use sparse matrices, it's most efficient if we avoid dense arrays entirely. 

This might also help w/ the computation time but no promises.
I would imagine that to convert a dense array to a sparse one you have to go through all zero values, which would add computation time (and would scale linearly w/ # of studies)

adelavega:
@tsalo @jdkent will you double check my review?

nimare/meta/utils.py
JulioAPeraza: > If I'm reading correctly the idea here it to use a list of 3D sparse arrays?

Yes, the list of 3D sparse arrays follows the same principle as it was previously implemented. The only advantage is that we avoid the memory peak by accumulating sparse arrays instead of the whole 3D array. But the dense to sparse and sparse to dense formatting is slowing everything.

> Isn't it possible with sparse to construct a 4D array? In a way I believe that's one of the main advantages of sparse

Yes, I think it is possible.

nimare/meta/utils.py
JulioAPeraza: Agreed. It is only 7MB for the peak memory, but it will be better to start using sparse arrays from coordinates directly.

nimare/meta/utils.py
JulioAPeraza: Thanks for the review. I agree with this. Now, I can see how to implement it. 
I think in the `KernelTrnasformer` we do need to convert the sparse arrays to dense since `nib.Nifti1Image()` expects a dense array.

nimare/meta/utils.py
adelavega: Ah okay, it's probably optimized to not take up as much memory as it could when empty

nimare/meta/utils.py
adelavega: Okay, yes, I think that's correct, although at that point we will have one less dimension (i.e. studies) so the dense array should have a much smaller footprint. 

Hopefully this speeds things up. If not I'm OK to try scipy.sparse also

nimare/meta/utils.py
JulioAPeraza: I was able to construct the 4D sparse array from coordinates, but this object type `sparse._coo.core.COO` is not iterable. To get this working we would need to modify the `KernelTrnasformer`. Alternatively, we could try again the list of 3D sparse arrays but build the sparse array from coordinates instead of compressing a dense array.

JulioAPeraza:
I modified the kernel transformer to index the MA maps in the 4D sparse array instead of iterating over it. The new changes are working, but I get a huge memory peak while and a high computation time. It seems that indexing the sparse array takes time and memory. I will try creating a list of 3D sparse arrays to avoid indexing.

JulioAPeraza:
Deleting these two variables (https://github.com/neurostuff/NiMARE/blob/main/nimare/meta/cbma/mkda.py#L334, https://github.com/neurostuff/NiMARE/blob/main/nimare/meta/cbma/mkda.py#L351) using `del` does not free up the memory. This is causing a high memory peak when the MA map of the second dataset is generated.
Calling the Garbage Collector (`gc.collect()`) after each `del` solved the issue.

JulioAPeraza:
Latest profiler results:
| Method | Dataset (`n_studies`) |  Peak Memory (MB)         | Time (s)          |
| :-: | - | - | - |
|  | Small (10)     | 108 | 2 |
| 4D Dense Array | Medium (100) | 914 | 3 |
| (Current Default) | Large (1000)     | 9,065 | 33 |
|  | Very Large (10000)     | 90,571 | 1,166 |
| | | | |
|  | Small (10)     | 53 | 2 |
| 4D Memory-Mapped Array | Medium (100) | 210 | 5 |
|  | Large (1000)     | 2,065 | 106 |
|  | Very Large (10000)     | 20,621 | 2,139 |
| | | | |
|  | Small (10)     | 66 | 11 |
| 4D Sparse Array | Medium (100) | 382 | 14 |
| (Proposed Default) | Large (1000)     | 3,679 | 48 |
|  | Very Large (10000)     | 36,649 | 993 |
| | | | |
|  | Small (10)     | 52 | 3 |
| List of 3D Sprse Arrays | Medium (100) | 369 | 7 |
| | Large (1000)     | 3,666 | 67 |
|  | Very Large (10000)     | 36,635 | 1,177 |

adelavega:
Nice.Those numbers look more reasonable to me. I didn't see why 3D vs 4D sparse should be any different unless there was a bug. Since 3D and 4D sparse seem equivalent, I would only keep one (4D) to keep the code simple. 

I would say thought that this is and of itself is a satisfactory improvement in memory and warrants removing the memmap approach since its not that much better, but is much slower. Are we doing this in a separate PR?

I'm still a little bit surprised we can't get this down a bit more though. 
What is the object size of the 4D object? For example for 1000 studies, what's the size of just the sparse array?
That would help to know if there's other places where we are using a lot of memory.

JulioAPeraza:
I think @jdkent  suggested removing the `memmap` approach in a separate PR.

For 1000 studies, for example, the size of the sparse array is ~500MB. However, we need to transform experiment-wise sparse arrays (`transformed_maps[0][idx]`) to dense to create the `nib.Nifti1Image` object (`nib.Nifti1Image` expects a numpy array), and also to mask the data (n-dimensional indexing is not supported in [pydata/sparse](https://sparse.pydata.org/en/stable/operations.html#advanced-indexing)). So, I think we end up accumulating a lot of memory in the variable [imgs](https://github.com/JulioAPeraza/NiMARE/blob/enh-sparse/nimare/meta/kernel.py#L228-L230) (~2MB per image = 2GB for 1000 studies) after each loop, plus the 500MB of the sparse array.
One solution could be to return the 4D sparse array directly and perform the sum here:

https://github.com/JulioAPeraza/NiMARE/blob/enh-sparse/nimare/meta/cbma/mkda.py#L327 

https://github.com/JulioAPeraza/NiMARE/blob/enh-sparse/nimare/meta/cbma/mkda.py#L345 

using the sparse array sum operation which I think is supported: https://sparse.pydata.org/en/stable/operations.html#reductions

JulioAPeraza:
Returning the sparse array by the transformer and using the sparse array in the `mkda` fit shows a considerable improvement in the memory peak:

| Method | Dataset (`n_studies`) |  Peak Memory (MB)         | Time (s)          |
| :-: | - | - | - |
|  | Small (10)     | 64 | 5 |
| 4D Sparse Array | Medium (100) | 172 | 9 |
| (Use sparse in `mkda`) | Large (1000)     | 1,607 | 60 |
|  | Very Large (10000)     | 15,947 | 1,117 |

The computation time increased because I had to run `np.unique` to the coordinates for the COO.sum(axis=0) to work well.

JulioAPeraza:
The code has gotten a little messy trying to support both sparse and `memap` arrays.
I think it would be better to just drop memory-mapping arrays in this PR. Thoughts?

adelavega:
Nice! That's a huge improvement now. That's the difference between running on your laptop and now. I think the performance hit is acceptable as its not huge and we can tackle improving performance in a different PR. 

I support dropping memmap. If you wanted to keep it clean you could start a new PR to drop memmap, confirm tests pass etc, we can merge that to master then you can rebase your current PR with those changes.

I assuming dropping memmap would be fairly easy so let's try that approach if that works for you.

JulioAPeraza:
I'm profiling the CorrelationDecoder now to see if we get any improvement there. Is there anything else we should include in this PR?

nimare/meta/cbma/mkda.py
```diff
@@ -383,6 +401,8 @@ def _fit(self, dataset1, dataset2):
 
         del n_selected, n_unselected
 
+        # print(cells)
```
adelavega, line 404:
```suggestion
```

setup.cfg
```diff
@@ -46,12 +46,14 @@ install_requires =
...
     nibabel>=3.0.0
     nilearn>=0.7.1
+    numba
     numpy
     pandas
     pymare>=0.0.3rc2
     requests
     scikit-learn
     scipy
+    sparse
```
adelavega, line 56: add version

setup.cfg
```diff
@@ -46,12 +46,14 @@ install_requires =
     matplotlib<3.5  # this is for nilearn, which doesn't include it in its reqs
     nibabel>=3.0.0
     nilearn>=0.7.1
+    numba
```
adelavega, line 49: is this used?

nimare/meta/utils.py
```diff
@@ -316,39 +317,50 @@ def compute_kda_ma(
...
+
+        temp_idx += n_brain_voxels
+
+    # Usually coords.shape[1] < n_coords, since n_brain_voxels < n_voxels sometimes
+    coords = coords[:, :temp_idx]
+
+    if not sum_overlap:
+        coords = np.unique(coords, axis=1)
+    data = np.full(coords.shape[1], value)
+    kernel_data = sparse.COO(coords, data, shape=kernel_shape)
```
adelavega, line 363: Would we want to make it parameterizable if sparse or dense arrays are used?

cc: @tsalo

adelavega:
No, I think this PR looks good! I'd like to give @tsalo a chance to review though.

setup.cfg
JulioAPeraza: I think it is used by sparse. I got an error in the testing when numba wasn't included.

tsalo: Can you add a comment to that effect?
```suggestion
    numba  # used by sparse
```

setup.cfg
```diff
@@ -46,12 +46,14 @@ install_requires =
...
     numpy
     pandas
     pymare>=0.0.3rc2
     requests
     scikit-learn
     scipy
+    sparse>=0.13.0
```
tsalo, line 56:
```suggestion
    sparse>=0.13.0  # for kernel transformers
```

tsalo: Can you also add `sparse==0.13.0` to the minimum requirements?

tsalo (changes requested):
I have a few small requests, but overall I think this is great. Thanks!

JulioAPeraza:
I have added the suggestions. Thanks for the review!

PR #677, opened 2022-04-28
Add DatasetSearcher class with search methods from Dataset

### LINKED ISSUE #669: Create DatasetSearcher class for transforming Datasets
## Summary

Per today's discussion, one thing that would make it easier to simplify the Dataset object would be to move the Dataset's searching methods into a new Transformer-like class.

## Additional details

It would make it easier to convert Dataset into a more NIMADS-like structure.

tsalo:
Closes #669.

- Create a `nimare.dataset.DatasetSearcher` class with all of the `get` methods originally in the `Dataset` class.
- Move several functions around to deal with circular import issues.
    - We probably need to enforce some kind of import rules for our modules to avoid this kind of situation.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -17,6 +17,7 @@ API
    :template: class.rst
 
    dataset.Dataset
+   dataset.DatasetSearcher
 
 
 .. _api_meta_ref:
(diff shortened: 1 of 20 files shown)


jdkent:
@tsalo I'm going through open pull requests, I'm sorry I haven't followed up on this.

I like the idea of having the internals move all the search functionality to a class that is separate from the data container. 

From a public API perspective, I think having dataset/studyset object handle simple introspection/search should be accessible through the dataset/studyset API.

PR #679, opened 2022-05-05
Refactor Correctors and remove statsmodels requirement

### LINKED ISSUE #667: Replace statsmodels dependency with internal multiple comparisons correction implementations
## Summary

NiMARE has a lot of dependencies, and the only thing we use statsmodels for is multiple comparisons correction. And even then, it's the methods that we don't normally use on imaging data. I think we could just implement Bonferroni and Benjamini-Hochberg correction internally.

## Next steps

1. Implement a subset of multiple comparisons correction methods as functions in NiMARE.
    - Bonferroni
    - Benjamini-Hochberg, with "indep" and "negcorr" options
2. Remove the statsmodels dependency.

tsalo:
Closes #667.

- Replace calls to statsmodels multiple comparisons correction functions with PyMARE functions.
- Remove `nimare.stats.fdr`.
- Refactor Correctors somewhat.
    - Generic correction methods are now actual instance methods, with proper docstrings and everything!
    - Estimator methods explicitly override generic ones.
- Rename `MKDAChi2.correct_fwe_bh` to `MKDAChi2.correct_fwe_indep` (breaking change).
- Replace "fdr_bh" option with "bh" in decoder multiple comparisons correction (breaking change).

To do:

- [x] Call the appropriate correction method within the Corrector based on the method parameter.
- [x] Version directives.


nimare/stats.py
```diff
@@ -275,27 +275,90 @@ def nullhist_to_p(test_values, histogram_weights, histogram_bins):
...
-        Array of p-values.
+    p_values : :obj:`numpy.ndarray`
+        Array of p values.
     q : :obj:`float`, optional
-        False discovery rate in fraction form. Default is 0.05 (5%).
+        Alpha value. Default is 0.05.
+    method : {"bh", "by"}, optional
+        Method to use for correction.
+        Either "bh" (Benjamini-Hochberg :footcite:p:`benjamini1995controlling`) or
+        "by" (Benjamini-Yekutieli :footcite:p:`benjamini2001control`).
```
jdkent, line 294: could you explain "by" in the other function signatures?

jdkent (changes requested):
Looks good! 

Main comment is I'm curious if the method parameter documentation could be defined once and referenced in other places, Don't consider it a blocking review comment, but it's nice not to have to repeat yourself, maybe rst substitutions would do what you want? https://stackoverflow.com/questions/8320909/repeated-parameters-with-python-sphinx

docs/conf.py
```diff
@@ -175,6 +175,7 @@
     "nibabel": ("https://nipy.org/nibabel/", None),
     "nilearn": ("http://nilearn.github.io/", None),
     "pymare": ("https://pymare.readthedocs.io/en/latest/", None),
+    "statsmodels": ("http://www.statsmodels.org/stable/", None),
```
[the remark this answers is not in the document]
tsalo: This is just so I can link back to statsmodels in the `fdr()` docstring, since the new code is taken from there.

tsalo:
>Main comment is I'm curious if the method parameter documentation could be defined once and referenced in other places, Don't consider it a blocking review comment, but it's nice not to have to repeat yourself, maybe rst substitutions would do what you want? https://stackoverflow.com/questions/8320909/repeated-parameters-with-python-sphinx

I think that's what nilearn does, and I certainly like it. Would you mind if I tackled that in a separate PR? I might open an issue about it.

nimare/stats.py
tsalo: Good point. I'll do that, although I'll mostly be parroting statsmodels' documentation there, since I'm not very familiar with the different approaches.

adelavega (commented):
Looks good to me, although I wonder if these non-imaging basic functions should perhaps live in pyMARE?

tsalo:
>Looks good to me, although I wonder if these non-imaging basic functions should perhaps live in pyMARE?

That seems reasonable to me, but PyMARE doesn't use these methods at the moment, so we would probably only want to add them to that package if we also create a Corrector class within PyMARE as well.

nimare/stats.py
tsalo: I've added a bit more info to the associated method docstrings.

adelavega (approved):
LGTM!

tsalo:
I've opened an issue about the parameter definitions (#687), so I'm going to resolve the conflicts and merge this once CI passes.

PR #685, opened 2022-05-28
Optimize numpy operations in MKDADensity Estimator and (M)KDAKernel

adelavega:
- Speeds in KDA meta-analysis, especially in the context of monte-carlo correction
- Main gains came from optimizing numpy operations, especially switchingout `np.unique` -> `unique_rows`
- Also chunked kernel operations (i.e. convolution & masking), by experiment ID, which is more efficient

Profile summary: Around a ~2.5-3x speed up for `correct_fwe_montecarlo` on a dataset with ~1000 coordinates


adelavega:
From this issue: https://github.com/numpy/numpy/issues/11136
I found an implementation that uses pandas and hashing to get unique rows wihout sorting for 2D arrays.

It seems to be significantly faster. All in all, for a dataset w/ ~100 studies, it went from ~0.53 to ~0.13s.

Still need to test more, but this should be helpful at least.

adelavega:
Replaced with scikit image version: https://github.com/scikit-image/scikit-image/blob/64103e6c90917fcfdef8343fd7dd4df32c910446/skimage/util/unique.py#L47

adelavega:
Profiling results using the "neurosynth_laird_studies.json" dataset (1117 coordinaes in 17 studies).

`correct_fwe_montecarlo, with 10 iterations`:

`main` branch: 9.67 s / ~1.1 iterations/s
This PR: 4.19s / ~3 iterations/s

It's looking like a 2.5x-3x speed up, just from speeding up `compute_kda_ma`

adelavega:
Ugh, `black` made changes in tons of files. Shouldn't black already have been run on `main`?

nimare/meta/cbma/mkda.py
```diff
@@ -166,12 +166,7 @@ def _compute_weights(self, ma_values):
         return weight_vec
 
     def _compute_summarystat_est(self, ma_values):
-        # Note: .dot should be faster, but causes multiprocessing to stall
-        # on some (Mac) architectures. If this is ever resolved, we can
-        # replace with the commented line.
-        # return ma_values.T.dot(self.weight_vec_).ravel()
-        weighted_ma_vals = ma_values * self.weight_vec_
-        return weighted_ma_vals.sum(0)
+        return ma_values.T.dot(self.weight_vec_).ravel()
```
adelavega, line 169: This seems like an edge case that is not worth the slow down, so I went w/ the faster version

nimare/meta/utils.py
```diff
@@ -275,14 +275,50 @@ def generate_input_fn():
     return niis
 
 
+def unique_rows(ar):
```
adelavega, line 278: This is straight from skimage, so I could import but  I don't want that as a dependency

tsalo:
I think you must have used a different version of black, because all files _should_ have already had black run on them and the linter is currently failing.

nimare/meta/utils.py
tsalo: Is their license amenable to using their code directly? If so, then I'm all for just using their function without adding a dependency.

adelavega:
```
Copyright (C) 2019, the scikit-image team
All rights reserved.

Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are
met:

 1. Redistributions of source code must retain the above copyright
    notice, this list of conditions and the following disclaimer.
 2. Redistributions in binary form must reproduce the above copyright
    notice, this list of conditions and the following disclaimer in
    the documentation and/or other materials provided with the
    distribution.
 3. Neither the name of skimage nor the names of its contributors may be
    used to endorse or promote products derived from this software without
    specific prior written permission.
```

Looks like it just need to add the copyright

adelavega:
By the way, I think the other major speed up we could potentially get is related to `sparse` memory usage.

Currently, `compute_kda_ma` returns a sparse matrix, which `correct_fwe_montecarlo` then converts to image because it calls:

```
self.kernel_transformer.transform(
            iter_df, masker=self.masker, return_type="array"
        )
```

with `return_type` == 'array`.

This means we spend a lot of time going from sparse --> dense. We save a tiny bit of memory doing this because we loop over experiments and only keep the masked voxels, but its slow. 

Starting from `dense` and staying that would should speed things up, but it's not clear to me by how much. It also seems like memory is a bigger problem than speed, so it may not be worth it.

@JulioAPeraza

nimare/meta/kernel.py
```diff
@@ -202,13 +206,11 @@ def transform(self, dataset, masker=None, return_type="image"):
         if return_type == "sparse":
             return transformed_maps[0]
 
+        # Optimize this, possily make own function
         imgs = []
         # Loop over exp ids since sparse._coo.core.COO is not iterable
         for i_exp, id_ in enumerate(transformed_maps[1]):
-            if isinstance(transformed_maps[0][i_exp], sparse._coo.core.COO):
```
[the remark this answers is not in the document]
adelavega: Apparently this check is necessary, because without it tests fails. However, it _shouldn't_ be necessary because the output should be sparse here. 

It's not in `ALEKernel` however.

This is inconsistent w/ the API because if you say "return_type='sparse'" you could get back a dense matrix w/ an ALEKernel

nimare/meta/utils.py
tsalo: Awesome! I'm happy to just use the code then.

nimare/meta/cbma/mkda.py
tsalo: I can test it out tomorrow.

tsalo: `tests.test_meta_mkda.test_MKDADensity_approximate_null` went from 7.37s to 6.49s after making the change on my MacBook Pro (OSX 12.3.1). Maybe the architecture Tal was referring to is the M1 chip? Another possibility is that the problem was unique to multiprocessing, and when I switched to joblib it resolved the issue without anyone noticing.

In case anyone wants to dig further into it, the commit where the workaround was added is 68c515a5511f9baaa265bc24026a12defe86d4fb.

adelavega:
@tsalo I believe this is ready to go as soon as you approve

nimare/meta/kernel.py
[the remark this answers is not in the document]
tsalo: Moved to #692.

nimare/meta/kernel.py
```diff
@@ -125,14 +125,16 @@ def transform(self, dataset, masker=None, return_type="image"):
...
             assert (
                 return_type != "dataset"
             ), "Input dataset must be a Dataset if return_type='dataset'."
 
             # Calculate IJK. Must assume that the masker is in same space,
             # but has different affine, from original IJK.
-            coordinates[["i", "j", "k"]] = mm2vox(coordinates[["x", "y", "z"]], mask.affine)
+            coordinates[["i", "j", "k"]] = mm2vox(
+                dataset[["x", "y", "z"]], mask.affine
+            )  # Possibily optimize this
```
tsalo, line 137:
```suggestion
            )  # TODO: Possibly optimize this
```

nimare/meta/kernel.py
```diff
@@ -181,7 +183,9 @@ def transform(self, dataset, masker=None, return_type="image"):
 
         # Generate the MA maps if they weren't already available as images
         if return_type == "array":
-            mask_data = mask.get_fdata().astype(bool)
+            mask_data = mask.get_fdata().astype(
+                bool
+            )  # Possibilty allow this to be passed in to not recompute
```
tsalo, line 188:
```suggestion
            )  # TODO: Possibly allow this to be passed in to not recompute
```

nimare/meta/kernel.py
```diff
@@ -202,6 +206,7 @@ def transform(self, dataset, masker=None, return_type="image"):
         if return_type == "sparse":
             return transformed_maps[0]
 
+        # Optimize this, possily make own function
```
tsalo, line 209:
```suggestion
        # TODO: Optimize this, possibly make own function
```

nimare/meta/kernel.py
```diff
@@ -412,7 +417,7 @@ class Peaks2MapsKernel(KernelTransformer):
 
     Parameters
     ----------
-    model_dir : :obj:`str`, optional
+    model_dir : :obj:`str`, optionalf
```
tsalo, line 420:
```suggestion
    model_dir : :obj:`str`, optional
```

nimare/meta/utils.py
```diff
@@ -337,30 +388,43 @@ def compute_kda_ma(
     cube = np.vstack([row.ravel() for row in np.mgrid[xx, yy, zz]])
     kernel = cube[:, np.sum(np.dot(np.diag(vox_dims), cube) ** 2, 0) ** 0.5 <= r]
 
-    # Preallocate coords array, np.concatenate is too slow
-    n_coords = ijks.shape[0] * kernel.shape[1]  # = n_peaks * n_voxels
-    coords = np.zeros((4, n_coords), dtype=int)
+    def _convolve_sphere(kernel, peaks):
+        # Convolve spheres
```
tsalo, line 392:
```suggestion
        """Convolve peaks with a spherical kernel.

        Parameters
        -----------
        kernel : 2D numpy.ndarray
            IJK coordinates of a sphere, relative to a central point (not the brain template).
        peaks : 2D numpy.ndarray
            The IJK coordinates of peaks to convolve with the kernel.

        Returns
        --------
        sphere_coords : 2D numpy.ndarray
            All coordinates that fall within any sphere.
            Coordinates from overlapping spheres will appear twice.
        """
```

Hopefully this is correct.

nimare/meta/utils.py
```diff
@@ -275,6 +275,57 @@ def generate_input_fn():
...
+
+    Raises
+    ------
+    ValueError : if `ar` is not two-dimensional.
+
+    Notes
+    -----
+    The function will generate a copy of `ar` if it is not
+    C-contiguous, which will negatively affect performance for large
+    input arrays.
```
tsalo, line 302: Can you add a reference to the scikit-image function? E.g., ```This is taken from skimage. See :func:`skimage.util.unique_rows`.``` We'll need to add skimage to the inter sphinx mapping in `docs/conf.py`.

tsalo (changes requested):
I _think_ I understand the changes to the code, and they look good to me.

I have some notes on the comments though- mostly just adding TODO to comments we might want to tackle in a future PR.

nimare/meta/kernel.py
adelavega: Oh, sure. I forgot to remove these so I think it's probably better to just remove them.

tsalo: That sounds good. I assumed they were just things you wanted to deal with later.

nimare/meta/utils.py
adelavega: Does that work w/o adding skimage as a dependency?

adelavega: yep, I figured it out I think.

tsalo:
I totally forgot- can you also add `unique_rows` to `docs/api.rst`? Sorry about that!

adelavega:
Yes, and actually I moved it to `nimare.utils` which makes more sense. We could presumably reuse it throughout the package, not just in cbma. Agree?

tsalo:
Agreed! I'm sure there'll be a few places where it could come in handy.

tsalo (approved):
The changes look good to me. I might write a test or two for the new function (or just copy them from skimage) in a future PR, but I don't want to block this one any longer. Once CI passes, feel free to merge.

I'm just going to change the title slightly, to match the pattern in other PRs.

EDIT: I also realized the changes were made to the kernel and the _MKDA_ (rather than KDA) Estimator.

nimare/meta/kernel.py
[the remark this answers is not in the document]
JulioAPeraza: Sorry that I missed this comment. I created a new folder for GitHub email and I didn't get notified of this one. I added a comment to the new issue.

PR #693, opened 2022-06-01
Do not zero out one-tailed z-statistics for p-values > 0.5

### LINKED ISSUE #511: Cluster-corrected logp and z-statistic maps do not completely match clusters
## Summary

This mismatch was reported by a user via email. The issue seems to result from the fact that, in a one-tailed p-to-z transform, p values greater than 0.5 result in z-statistics of 0, while in the -log10(p) transform, those same p-values result in very low, but still non-zero, values. As such, the set of non-zero voxels do not match up between maps, _but_ those mismatched voxels are all very nonsignificant and will disappear when the images are thresholded with equivalent thresholds.

Still, it would be nice to address this.

## Additional details

- NiMARE version: ~0.0.8 (dev)

JulioAPeraza:
Closes #511.

- Do not zero out one-tailed z-statistics for p-values > 0.5. **Note:** p-value = 0.5 will still be set to 0, so logp and z-statistic maps will not completely match in that case.


tsalo (changes requested):
The changes make sense to me. Can you add a new test that covers this case though (or expand an existing one)?

JulioAPeraza:
@tsalo if we don't zero out one-tailed z-statistics for p-values > 0.5, should we: (1) leave the negative value, (2) apply the abs() function since the values returned by this function should be unsigned z-statistics, or (3) replace the negative values with a small number `1.0e-300`?

tsalo:
I'm not sure... maybe someone else who has a better handle on stats can answer that?

tsalo:
@yifan0330 seemed to have a good handle on z-to-p conversion (and probably p-to-z conversion), given #748 and #749. @yifan0330 would you be willing to take a look at this?

nimare/tests/test_transforms.py
```diff
@@ -247,18 +247,22 @@ def test_ddimages_to_coordinates_merge_strategy(testdata_ibma):
...
 @pytest.mark.parametrize(
-    "z,tail,expected_p",
+    "expected_z,tail,expected_p",
     [
         (0.0, "two", 1.0),
         (0.0, "one", 0.5),
         (1.959963, "two", 0.05),
         (1.959963, "one", 0.025),
         (-1.959963, "two", 0.05),
+        (-1.959963, "one", 0.025),
```
yifan0330, line 257: I think it should be (-1.959963, "one", 0.975). I've corrected it in issue #748

yifan0330, line 257: This test case should be (-1.959964, "one", 0.0975). I think I've corrected in issue #748

nimare/tests/test_transforms.py
```diff
@@ -247,18 +247,22 @@ def test_ddimages_to_coordinates_merge_strategy(testdata_ibma):
...
         ([0.0, 1.959963, -1.959963], "two", [1.0, 0.05, 0.05]),
+        ([0.0, 1.959963, -1.959963], "one", [0.5, 0.025, 0.025]),
```
yifan0330, line 259: ([0.0, 1.959963, -1.959963], "one", [0, 0.025, 0.975])

yifan0330:
> @yifan0330 seemed to have a good handle on z-to-p conversion (and probably p-to-z conversion), given #748 and #749. @yifan0330 would you be willing to take a look at this?

@JulioAPeraza @tsalo I saw you replace negative z value with values estimated by cdf, I am wondering if it's the same as taking absolute value of z (since `z` and `zval_cdf` have same value but opposite sign)? Also, I left two comments on testing case (I think z=-1.959964 corresponds to p=0.975 in one tail test).

JulioAPeraza:
Thanks so much, @yifan0330! I've incorporated your suggestions.
@tsalo Do you think this is ready to merge?

tsalo (approved):
@JulioAPeraza if @yifan0330 is happy with the changes then absolutely, I think it's good to merge.

yifan0330:
> @JulioAPeraza if @yifan0330 is happy with the changes then absolutely, I think it's good to merge.

Yeah I am happy with Julio's changes :)

JulioAPeraza:
Thanks! Merging now.

PR #699, opened 2022-06-07
Add warning when coordinates dataset contains both positive and negative z_stats

### LINKED ISSUE #465: Raise warning if coordinates dataset contains both positive and negative z_stats
## Summary

Seems weird to perform a meta-analysis with peaks being reported at both the positive and negative thresholds

## Additional details

- NiMARE version: 0.0.7 dev

Thinking about how users may use the output of #446 

Raise warning/error when both positive/negative values are detected in z_stat column?

### What actually happened?

The dataset will run on any estimator

JulioAPeraza:
Closes #465.

- Check for `z_stat` in `Dataset.coordinates.columns`.
- Check if there are values with `z_stat >= 0` AND `z_stat < 0` in `Dataset.__init__()` and when the `ImagesToCoordinates.transform()` is applied.

### DIFF
--- a/nimare/dataset.py
+++ b/nimare/dataset.py
@@ -105,11 +105,15 @@ class Dataset(NiMAREBase):
         self.basepath = None
 
         if "z_stat" in self.coordinates.columns:
-            # Raise warning if coordinates dataset contains both positive and negative z_stats
-            if ((self.coordinates["z_stat"].values >= 0).any()) and (
-                (self.coordinates["z_stat"].values < 0).any()
-            ):
-                warnings.warn("Coordinates dataset contains both positive and negative z_stats")
+            # "z_stat" column may contain Nones
+            if not self.coordinates["z_stat"].isna().any():
+                # Raise warning if coordinates dataset contains both positive and negative z_stats
+                if ((self.coordinates["z_stat"].values >= 0).any()) and (
+                    (self.coordinates["z_stat"].values < 0).any()
+                ):
+                    warnings.warn(
+                        "Coordinates dataset contains both positive and negative z_stats"
+                    )
 
     def __repr__(self):
         """Show basic Dataset representation.
--- a/nimare/dataset.py
+++ b/nimare/dataset.py
@@ -4,6 +4,7 @@ import inspect
 import json
 import logging
 import os.path as op
+import warnings
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


tsalo:
It looks like the additions to `transforms.py` is covered by tests, but the changes to `dataset.py` are not.

nimare/dataset.py
```diff
@@ -103,6 +104,17 @@ def __init__(self, source, target="mni152_2mm", mask=None):
...
 
+        if "z_stat" in self.coordinates.columns:
+            # "z_stat" column may contain Nones
+            if not self.coordinates["z_stat"].isna().any():
+                # Raise warning if coordinates dataset contains both positive and negative z_stats
+                if ((self.coordinates["z_stat"].values >= 0).any()) and (
+                    (self.coordinates["z_stat"].values < 0).any()
+                ):
+                    warnings.warn(
+                        "Coordinates dataset contains both positive and negative z_stats"
```
tsalo, line 115: I think we should expand the warning a bit, to give users some context. Perhaps something like "The algorithms currently implemented in NiMARE are designed for one-sided tests, so results may not make sense."

@jdkent WDYT?

JulioAPeraza:
> It looks like the additions to `transforms.py` is covered by tests, but the changes to `dataset.py` are not.

I think this happens because `db_file = op.join(get_test_data_path(), "neurosynth_dset.json")` contains Nones in the "z_stat" column, so it doesn't get to check for positive and negative values.

tsalo:
Do you know how it's checked in `transforms.py`? Do we can any tests that actually cover the case where there are positive and negative z-stat values? My concern is that we don't currently have any methods that _use_ z-stat values, so I think the only place we're going to catch any bugs is by testing the warning directly, rather than downstream in any meta-analyses.

JulioAPeraza:
> Do you know how it's checked in `transforms.py`?

I think it is checked when an images dataset is transformed to coordinates here: https://github.com/neurostuff/NiMARE/blob/main/nimare/tests/test_transforms.py#L131. We don't initialize a Dataset class here, so the changes to `dataset.py` are not covered.

I think to test the changes to `dataset.py` we would need to initialize the Dataset class with a `source` that contains a `z_stat `with negative and positive values.

tsalo:
I re-ran that step and it worked this time. I assume it was just a random upload issue. I'll take another quick look and approve if it looks good.

tsalo (approved):
LGTM!

tsalo:
@JulioAPeraza do you want to merge this?

JulioAPeraza:
Sure!

PR #701, opened 2022-06-10
Fix CHANGELOG formatting issues

JulioAPeraza:
Closes None.

- Fix the admonitions that were reformatted by update-changelog Action. 
- Remove bullet points from multiline elements by having everything on a single line.
- Reorder the release versions by switching the latest version 0.0.12rc6 with 0.0.12rc3. 
- Add links to the full changelog to the headings to match the format of the update-changelog Action.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,6 +1,9 @@
 name: "Run Tests"
 
 on:
+  push:
+    branches:
+      - "main"
   pull_request:
     branches:
       - '*'
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,9 +1,6 @@
 name: "Run Tests"
 
 on:
-  push:
-    branches:
-      - "main"
   pull_request:
     branches:
       - '*'
--- a/.github/workflows/update-changelog.yml
+++ b/.github/workflows/update-changelog.yml
@@ -25,5 +25,5 @@ jobs:
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
           branch: main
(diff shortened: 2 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -1,9 +1,6 @@
 name: "Run Tests"
 
 on:
-  push:
-    branches:
-      - "main"
```
tsalo, line 6: Is this a change you want merged?

tsalo (commented):
Thanks for fixing the changelog. I think we need to add some info about making releases and/or updating the changelog to the contributing guidelines, but that can come in a later PR.

.github/workflows/testing.yml
JulioAPeraza: I deleted these lines to skip the test when the CHANGELOG is updated (there is a push to main). Do you think this is the best way to do that? I believe we don't usually have direct commits pushes to main, and the Docs build is not affected by skipping those tests.

tsalo: I'm thinking there might be a way to filter based on tags or the commit message/PR title, so we could run it generally, but skip commits from the changelog Action.

tsalo:
Ah, we _do_ have the skip CI action, so if we can include any of the following in the changelog commits, that would work: `[skip ci];[ci skip];[skip github]`.

JulioAPeraza:
That's a very nice solution. Thanks!

tsalo:
Darn, it looks like there's a problem in the YAML file: https://github.com/neurostuff/NiMARE/actions/runs/2490490055/workflow

PR #703, opened 2022-06-11
Add information about maintaining NiMARE to developer's guide

tsalo:
Closes None, but is related to #701.

- Add section on project maintenance to developer's guide.

### DIFF
--- a/docs/dev_guide.rst
+++ b/docs/dev_guide.rst
@@ -73,3 +73,41 @@ this will add the full list of changes in the new release based on our template.
 Once the release notes have been completed, you can publish the release.
 This will make the release on GitHub and will also trigger GitHub Actions to
 (1) publish the new release to PyPi and (2) update the changelog file.
+
+Updating the Changelog
+``````````````````````
+
+The Changelog is a Markdown file that is updated automatically by a GitHub Action.
+It is a list of all releases, with a copy of the release notes for each release.
+Since the Changelog is updated automatically, you should not have to edit it manually very often;
+however, there are some cases where you may need to do so.
+One common situation is a mistake in the release notes.
+Another possibility is that someone identifies a bug in a past release,
+in which case it may be useful to add an admonition to that release's release notes.
+
+The auto-generated Changelog follows a specific format,
+so it is important to understand its idiosyncrasies before making any manual changes.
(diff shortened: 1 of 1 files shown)


tsalo:
@JulioAPeraza, I'd love to get your thoughts on this one. I want to expand the maintenance documentation, and you're the one who will use that documentation the most.

JulioAPeraza (commented):
Should we also add the guidelines for testing with local environments using `pytest -v nimare` and `make html`? The step will look like this:
```
conda create -p /path/to/nimare_env pip python=3.9
conda activate /path/to/nimare_env
pip install -e .[all]
```

docs/dev_guide.rst
```diff
@@ -43,3 +43,71 @@ Once inside the container, you can install NiMARE:
...
+
+          - The ALESubtraction class from this release should not be used, as it uses a symmetric null distribution,
+            which does not work properly for comparisons between Datasets with different sizes.
+
+    Here is a valid single-line version of that admonition:
+
+    .. code-block:: text
+
+      .. warning:: Known Bugs This version contains some bugs that were identified after it was released.
+      \   - The ALESubtraction class from this release should not be used, as it uses a symmetric null distribution, which does not work properly for comparisons between Datasets with different sizes.
```
JulioAPeraza, line 113: This looks great!
From the latest release, it looks like this line is not being reformated by update-changelog Action. In the future, I will look for a way to reproduce the typical format using an equivalent of newline `\n` and horizontal tab `\t`.

tsalo:
>Should we also add the guidelines for testing with local environments

I think that's a good idea. That's important for any devs, rather than just maintainers, so I'll add it to the overall development guide.

docs/dev_guide.rst
tsalo: I agree that that would be optimal, but I don't think it's a priority as long as we have something that works. We can always update these docs with more info as things change.

JulioAPeraza: Agreed!

tsalo:
@JulioAPeraza is there anything else you'd like to see in this PR?

JulioAPeraza:
No, I think this looks good.

PR #644, opened 2022-02-17
Remove Peaks2Maps from NiMARE

tsalo:
Closes None, but finishes what was proposed in #615 and started in #616.

Changes proposed in this pull request:

- Remove the `Peaks2MapsKernel` and associated functions, documentation, and examples.

PR #709, opened 2022-06-13
Use sparse array in ALEKernel

### LINKED ISSUE #692: Handle sparse/dense uniformly for different kernels
https://github.com/neurostuff/NiMARE/blob/e0d7cf5dc939729d18c637f1d73ee694acdc6bac/nimare/meta/kernel.py#L202

This line is buggy, because only MKDA return a sparse matrix.

For ALE, this line returns a dense matrix.

In addition, it's worth thinking what we can do to speed things up, as going from data -> sparse then sparse -> costs time

tsalo:
Closes #692.

**NOTE**: This isn't optimized at all. I just tacked the conversion on to the end of `compute_ale_ma`.

- Convert MA map to sparse array in `ALEKernel._transform`.
- Convert sparse arrays from `compute_ale_ma` back into dense arrays within `generate._create_foci`.

### DIFF
--- a/nimare/generate.py
+++ b/nimare/generate.py
@@ -270,7 +270,7 @@ def _create_foci(foci, foci_percentage, fwhm, n_studies, n_noise_foci, rng, spac
         for peak in ground_truth_foci_ijks
         if peak.size
     }
-    foci_prob_maps = {k: v.todense() for k, v in foci_prob_maps.items()}
+    # foci_prob_maps = {k: v.todense() for k, v in foci_prob_maps.items()}
 
     # get study specific instances of each foci
     signal_studies = int(round(foci_percentage * n_studies))
--- a/nimare/generate.py
+++ b/nimare/generate.py
@@ -270,6 +270,7 @@ def _create_foci(foci, foci_percentage, fwhm, n_studies, n_noise_foci, rng, spac
         for peak in ground_truth_foci_ijks
         if peak.size
     }
+    foci_prob_maps = {k: v.todense() for k, v in foci_prob_maps.items()}
 
     # get study specific instances of each foci
     signal_studies = int(round(foci_percentage * n_studies))
--- a/nimare/meta/kernel.py
+++ b/nimare/meta/kernel.py
@@ -339,9 +339,14 @@ class ALEKernel(KernelTransformer):
                 else:
                     kern = kernels[sample_size]
 
-            kernel_data = compute_ale_ma(mask.shape, ijk, kern)
+            ma_values = compute_ale_ma(mask.shape, ijk, kern)
 
-            transformed.append(kernel_data)
+            # Convert dense array to sparse
+            nonzero_idx = np.vstack(np.where(ma_values))
+            nonzero_values = ma_values[nonzero_idx[0, :], nonzero_idx[1, :], nonzero_idx[2, :]]
+            ma_values = sparse.COO(nonzero_idx, nonzero_values, shape=mask.shape)
+
+            transformed.append(ma_values)
 
         return transformed, exp_ids
 
--- a/nimare/meta/kernel.py
+++ b/nimare/meta/kernel.py
@@ -340,14 +340,14 @@ class ALEKernel(KernelTransformer):
                 else:
                     kern = kernels[sample_size]
 
-            ma_values = compute_ale_ma(mask.shape, ijk, kern)
+            kernel_data = compute_ale_ma(mask.shape, ijk, kern)
 
             # Convert dense array to sparse
-            nonzero_idx = np.vstack(np.where(ma_values))
-            nonzero_values = ma_values[nonzero_idx[0, :], nonzero_idx[1, :], nonzero_idx[2, :]]
-            ma_values = sparse.COO(nonzero_idx, nonzero_values, shape=mask.shape)
+            nonzero_idx = np.vstack(np.where(kernel_data))
+            nonzero_values = kernel_data[nonzero_idx[0, :], nonzero_idx[1, :], nonzero_idx[2, :]]
+            kernel_data = sparse.COO(nonzero_idx, nonzero_values, shape=mask.shape)
 
-            transformed.append(ma_values)
+            transformed.append(kernel_data)
 
         return transformed, exp_ids
 
--- a/nimare/meta/kernel.py
+++ b/nimare/meta/kernel.py
@@ -309,30 +309,31 @@ class ALEKernel(KernelTransformer):
         self.sample_size = sample_size
 
     def _transform(self, mask, coordinates):
-        kernels = {}  # retain kernels in dictionary to speed things up
         exp_ids = coordinates["id"].unique()
+
+        use_dict = True
+
+        if self.sample_size is not None:
+            _, kern = get_ale_kernel(mask, sample_size=self.sample_size)
+            use_dict = False
+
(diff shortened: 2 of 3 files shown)


JulioAPeraza:
> once ALEKernel supports sparse arrays, do we need to keep these lines at all?

I think that once `ALEKernel` supports sparse arrays and `Peaks2MapsKernel` is removed, `transformed_maps[0][i_exp]` should always be `sparse._coo.core.COO`. So, I think we don't need to keep those lines.

tsalo:
Thanks @JulioAPeraza. I totally forgot about Peaks2MapsKernel! I will remove that line in #644 instead.

adelavega:
It's hard to say, because if covering to sparse adds time without saving any space then maybe we should just take the time to use sparse from the start and maximize memory.

Might be worth profiling at least that line to see if it's bad or not

tsalo:
Here's what I got from testing ALEKernel:

Memory:
- `main`: 
    - Mem usage: 343.7 MiB 
    - Increment: 309.1 MiB
- `sparsity-2`: 
    - Mem usage: 286.2 MiB 
    - Increment: 184.7 MiB

Time: 
- `main`: 175 ms ± 1.79 ms per loop (mean ± std. dev. of 4 runs, 50 loops each)
- `sparsity-2`: 430 ms ± 3.82 ms per loop (mean ± std. dev. of 4 runs, 50 loops each)

Here's the code I used to do it:

```python
# Cell 1
%load_ext memory_profiler
import os.path as op
import nimare
from nimare.tests.utils import get_test_data_path

dset = nimare.dataset.Dataset(op.join(get_test_data_path(), 'nidm_pain_dset.json'))
kern = nimare.meta.kernel.ALEKernel()

# Cell 2
%mprun -T ale -f nimare.meta.utils.compute_ale_ma kern.transform(dset)

# Cell 3
%timeit -r 4 -n 50 kern.transform(dset)
```

The gist seems to be that it _does_ slow the kernel down a bit, but it also uses less memory, as expected.
I'd be comfortable merging as-is, releasing 0.0.12, and then optimizing it afterwards. WDYT?

adelavega:
Hmm, I'm not sure. It cuts down memory by 0.6xf but time increases 2-3x.

Wouldn't this severely affect monte-carlo Correctors? If so then I'm concerned. Otherwise, it's fast enough.

tsalo:
That's a good point. It will cause problems for Monte Carlo-based correction.

I started trying to implement a pure-sparse version of `compute_ale_ma` in https://github.com/tsalo/NiMARE/tree/sparsity-3, but it ended up being very slow. Perhaps someone can make some progress with that, but working on it gave me a headache.

We could also try reorganizing `compute_ale_ma` to be more like `compute_kda_ma` (i.e., create the MA maps for all studies and return a 4D sparse array), but @JulioAPeraza might be better able to do that.

JulioAPeraza:
I am happy to work on that! I think we can make the kernel transformer to return the sparse array for ALE here too:

https://github.com/neurostuff/NiMARE/blob/cb542af17c4f969afd05e4fd381a1200a1e4817c/nimare/meta/kernel.py#L202-L203

As @adelavega mentioned here https://github.com/neurostuff/NiMARE/issues/692, that may save computation time.

JulioAPeraza:
I was able to create the 4D sparse array in `compute_ale_ma`, and return it by kernel transformer. That step is fast and saves more memory. However,  it has been difficult to work with the sparse array in the `CBMAEstimator._fit()`, in particular for the default `null_method="approximate"`:

https://github.com/neurostuff/NiMARE/blob/ac57801b288675e97b91d4356ab8aa83ced9edfc/nimare/meta/cbma/base.py#L168-L169

where the `np.histogram()` expects a numpy array. The only feasible solution I could think of is to loop over the experiment's MA maps and convert them to dense, which would impact the computation time, and it's pretty much the same that is already implemented in this PR.

tsalo:
That's unfortunate. The approximate method is the default, so we want it to be fast. Do you think we can add a new histogram function for sparse arrays? I imagine there's something floating around online. We'd need to make sure it's faster than converting the arrays to dense, of course.

JulioAPeraza:
> Can scipy.ndimage.histogram handle sparse arrays?

I think `scipy.ndimage.histogram` calls `numpy.asarray` to convert the input data to a numpy array, which doesn't 
 support sparse arrays.

> Do you think we can add a new histogram function for sparse arrays?

I have seen some examples online using `numba`, which is faster than `np.histogram()`. They loop over all the elements of the array (avoiding the memory issue), which in our case would be possible to do since the sparse array can be indexed.

JulioAPeraza:
Unfortunately, we cannot pass a sparse array (COO object) to Numba’s JIT compiler.

adelavega:
Take a look at this: https://www.quansight.com/post/extending-numba-types-for-clean-fast-code

Looks difficult though

tsalo:
What about running `numpy.histogram` on the sparse array's `.data` attribute? I believe that `data` is a numpy array of all of the values.

If we can loop through volumes in the 4D sparse array (assuming that doesn't involve any intensive under-the-hood operations), then we could run the histogram function directly on the nonzero data. It should be trivial to calculate the number of zero elements as well, in order to add those manually to the appropriate bin in the resulting histogram.

So around:
https://github.com/neurostuff/NiMARE/blob/ac57801b288675e97b91d4356ab8aa83ced9edfc/nimare/meta/cbma/ale.py#L222

We'd have the following:
```python
for i_study in range(ma_values.shape[3]):
    study_ma_values = ma_values[..., i_study].data
    n_nonzero_voxels = study_ma_values.shape[0]
    n_zero_voxels = n_mask_voxels - n_nonzero_voxels  # assuming we have the number of mask voxels already
    ma_hist = np.histogram(study_ma_values , bins=bin_edges, density=False)
    ma_hist[0] += n_zero_voxels
```

JulioAPeraza:
> What about running `numpy.histogram `on the sparse array's `.data` attribute?

Yeah, I tried using `.data` and the computation time was at the order of using `.todense()`, around 10x slower.

JulioAPeraza:
However, using `.data` only one time (just before the for loop) to get the whole data is fast enough. We will only need to index the data array using the `.coords` to get study_ma_values.

```python
data = ma_values.data
coords = ma_values.coords
for i_study in range(ma_values.shape[0]):
    study_ma_values = data[coords[0, :] == exp_idx]
    n_nonzero_voxels = study_ma_values.shape[0]
    n_zero_voxels = n_mask_voxels - n_nonzero_voxels  # assuming we have the number of mask voxels already
    ma_hist = np.histogram(study_ma_values , bins=bin_edges, density=False)
    ma_hist[0] += n_zero_voxels
```

tsalo:
This is really weird, since, AFAICT, `COO.data` is an attribute, not a property (see [here](https://github.com/pydata/sparse/blob/30b7172b16a6b333cdda5afc8266e3c1ae067d26/sparse/_coo/core.py#L229)). It doesn't seem like there is any processing going on when we access the `.data` attribute, and yet there is this weird slow-down. Is there some kind of caching happening?

JulioAPeraza:
I have been working on this branch: https://github.com/JulioAPeraza/NiMARE/tree/ale-sparse. I think it's ready to run some tests and profiling. Should I add the commits to this PR or open a new PR?

tsalo:
I think you should open a new PR.

PR #722, opened 2022-06-18
Pin minimum version of pandas

jdkent:
using pandas 1.0.3, and 1.0.4 throw errors, but 1.1.0 does not when assigning values to columns that do not yet exist in a pandas dataframe: 
(this line)
https://github.com/neurostuff/NiMARE/blob/cb542af17c4f969afd05e4fd381a1200a1e4817c/nimare/meta/cbma/base.py#L127

Closes # None.

- make minimal pandas version 1.1.0
-

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,7 @@ install_requires =
     nilearn>=0.7.1
     numba  # used by sparse
     numpy
-    pandas
+    pandas>=1.1.0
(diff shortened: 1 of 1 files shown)


tsalo (approved):
Thanks for this. Do you mind if I update the PR title before merging? We switched away from using the prefixes in favor of labels.

PR #725, opened 2022-07-01
Use sparse array in ALE, ALESubtraction, SCALE, KDA, and MKDADensity

### LINKED ISSUE #692: Handle sparse/dense uniformly for different kernels

JulioAPeraza:
Closes #692.

- Create the MA maps for all studies in `compute_ale_ma` and return a 4D sparse array.
- Return sparse array by `ALEKernel._transform` and `KernelTransformer.transform`.
- Add support for `ma_maps` 4D sparse array in `ALE` and `CBMAEstimator`. This will involve adding support for sparse array in `ALE._compute_summarystat_est`, `ALE._determine_histogram_bins`, and `ALE._compute_null_approximate`.

### DIFF
--- a/nimare/meta/cbma/ale.py
+++ b/nimare/meta/cbma/ale.py
@@ -383,13 +383,7 @@ class ALESubtraction(PairwiseCBMAEstimator):
         n_voxels = diff_ale_values.shape[0]
 
         # Combine the MA maps into a single array to draw from for null distribution
-        if isinstance(ma_maps1, sparse._coo.core.COO) and isinstance(
-            ma_maps2, sparse._coo.core.COO
-        ):
-            ma_arr = sparse.concatenate((ma_maps1, ma_maps2))
-        else:
-            # For ma_map_reuse
-            ma_arr = np.vstack((ma_maps1, ma_maps2))
+        ma_arr = sparse.concatenate((ma_maps1, ma_maps2))
 
         del ma_maps1, ma_maps2
 
--- a/nimare/meta/cbma/ale.py
+++ b/nimare/meta/cbma/ale.py
@@ -384,15 +384,11 @@ class ALESubtraction(PairwiseCBMAEstimator):
         ma_maps1 = self._collect_ma_maps(
             maps_key="ma_maps1",
             coords_key="coordinates1",
-            return_type="array",
         )
         ma_maps2 = self._collect_ma_maps(
             maps_key="ma_maps2",
             coords_key="coordinates2",
-            return_type="array",
         )
-
-        n_grp1, n_voxels = ma_maps1.shape
 
         # Get ALE values for the two groups and difference scores
         grp1_ale_values = self._compute_summarystat_est(ma_maps1)
@@ -400,8 +396,17 @@ class ALESubtraction(PairwiseCBMAEstimator):
         diff_ale_values = grp1_ale_values - grp2_ale_values
         del grp1_ale_values, grp2_ale_values
 
+        n_grp1 = ma_maps1.shape[0]
+        n_voxels = diff_ale_values.shape[0]
+
         # Combine the MA maps into a single array to draw from for null distribution
-        ma_arr = np.vstack((ma_maps1, ma_maps2))
+        if isinstance(ma_maps1, sparse._coo.core.COO) and isinstance(
+            ma_maps2, sparse._coo.core.COO
+        ):
+            ma_arr = sparse.concatenate((ma_maps1, ma_maps2))
+        else:
+            # For ma_map_reuse
+            ma_arr = np.vstack((ma_maps1, ma_maps2))
 
         del ma_maps1, ma_maps2
 
@@ -458,6 +463,13 @@ class ALESubtraction(PairwiseCBMAEstimator):
 
     def _compute_summarystat_est(self, ma_values):
         stat_values = 1.0 - np.prod(1.0 - ma_values, axis=0)
+
+        if isinstance(stat_values, sparse._coo.core.COO):
+            mask_data = self.masker.mask_img.get_fdata().astype(bool)
+
+            stat_values = stat_values.todense().reshape(-1)  # Indexing a .reshape(-1) is faster
+            stat_values = stat_values[mask_data.reshape(-1)]
+
         return stat_values
 
     def _run_permutation(self, i_iter, n_grp1, ma_arr, iter_diff_values):
@@ -480,8 +492,8 @@ class ALESubtraction(PairwiseCBMAEstimator):
         gen = np.random.default_rng(seed=i_iter)
         id_idx = np.arange(ma_arr.shape[0])
         gen.shuffle(id_idx)
-        iter_grp1_ale_values = 1.0 - np.prod(1.0 - ma_arr[id_idx[:n_grp1], :], axis=0)
-        iter_grp2_ale_values = 1.0 - np.prod(1.0 - ma_arr[id_idx[n_grp1:], :], axis=0)
+        iter_grp1_ale_values = self._compute_summarystat_est(ma_arr[id_idx[:n_grp1], :])
+        iter_grp2_ale_values = self._compute_summarystat_est(ma_arr[id_idx[n_grp1:], :])
         iter_diff_values[i_iter, :] = iter_grp1_ale_values - iter_grp2_ale_values
 
     def _alediff_to_p_voxel(self, i_voxel, stat_value, voxel_null):
@@ -676,15 +688,11 @@ class SCALE(CBMAEstimator):
 
         stat_values = 1.0 - np.prod(1.0 - ma_values, axis=0)
 
-        # np.array type is used by _determine_histogram_bins to calculate max_poss_ale
         if isinstance(stat_values, sparse._coo.core.COO):
             mask_data = self.masker.mask_img.get_fdata().astype(bool)
 
             stat_values = stat_values.todense().reshape(-1)  # Indexing a .reshape(-1) is faster
             stat_values = stat_values[mask_data.reshape(-1)]
-
-            # This is used by _compute_null_approximate
-            self.n_mask_voxels = stat_values.shape[0]
 
         return stat_values
 
--- a/nimare/meta/cbma/ale.py
+++ b/nimare/meta/cbma/ale.py
@@ -162,7 +162,7 @@ class ALE(CBMAEstimator):
             stat_values = stat_values[mask_data.reshape(-1)]
 
             # This is used by _compute_null_approximate
-            self.n_mask_voxels = stat_values.shape[0]
+            self.__n_mask_voxels = stat_values.shape[0]
 
         return stat_values
 
--- a/nimare/meta/cbma/ale.py
+++ b/nimare/meta/cbma/ale.py
@@ -171,21 +171,15 @@ class ALE(CBMAEstimator):
 
         Parameters
         ----------
-        ma_maps : list of imgs or sparse._coo.core.COO
+        ma_maps : :obj:`sparse._coo.core.COO`
             MA maps.
-            The ma_maps can be:
-            (1) a list of imgs containing MA values
-            or (2) a 4d sparse array of MA maps,
+            The ma_maps can be a 4d sparse array of MA maps,
 
         Notes
         -----
         This method adds one entry to the null_distributions_ dict attribute: "histogram_bins".
         """
-        if isinstance(ma_maps, list):
-            ma_values = self.masker.transform(ma_maps)
-        elif isinstance(ma_maps, sparse._coo.core.COO):
-            ma_values = ma_maps
-        else:
+        if not isinstance(ma_maps, sparse._coo.core.COO):
             raise ValueError(f"Unsupported data type '{type(ma_maps)}'")
 
         # Determine bins for null distribution histogram
@@ -194,7 +188,7 @@ class ALE(CBMAEstimator):
         INV_STEP_SIZE = 100000
         step_size = 1 / INV_STEP_SIZE
         # Need to convert to dense because np.ceil is too slow with sparse
-        max_ma_values = ma_values.max(axis=[1, 2, 3]).todense()
+        max_ma_values = ma_maps.max(axis=[1, 2, 3]).todense()
 
         # round up based on resolution
         max_ma_values = np.ceil(max_ma_values * INV_STEP_SIZE) / INV_STEP_SIZE
@@ -208,11 +202,9 @@ class ALE(CBMAEstimator):
 
         Parameters
         ----------
-        ma_maps : list of imgs or sparse._coo.core.COO
+        ma_maps : :obj:`sparse._coo.core.COO`
             MA maps.
-            The ma_maps can be:
-            (1) a list of imgs containing MA values
-            or (2) a 4d sparse array of MA maps,
+            The ma_maps can be 4D sparse array of MA maps,
 
         Notes
         -----
@@ -221,11 +213,7 @@ class ALE(CBMAEstimator):
             - "histogram_bins"
             - "histweights_corr-none_method-approximate"
         """
-        if isinstance(ma_maps, list):
-            ma_values = self.masker.transform(ma_maps)
-        elif isinstance(ma_maps, sparse._coo.core.COO):
-            ma_values = ma_maps
-        else:
+        if not isinstance(ma_maps, sparse._coo.core.COO):
             raise ValueError(f"Unsupported data type '{type(ma_maps)}'")
 
         assert "histogram_bins" in self.null_distributions_.keys()
@@ -237,11 +225,11 @@ class ALE(CBMAEstimator):
         bin_edges = bin_centers - (step_size / 2)
         bin_edges = np.append(bin_centers, bin_centers[-1] + step_size)
 
-        n_exp = ma_values.shape[0]
+        n_exp = ma_maps.shape[0]
         n_bins = bin_centers.shape[0]
         ma_hists = np.zeros((n_exp, n_bins))
-        data = ma_values.data
-        coords = ma_values.coords
+        data = ma_maps.data
+        coords = ma_maps.coords
         for exp_idx in range(n_exp):
             # The first column of coords is the fourth dimension of the dense array
             study_ma_values = data[coords[0, :] == exp_idx]
@@ -665,7 +653,7 @@ class SCALE(CBMAEstimator):
         return images
 
     def _compute_summarystat_est(self, data):
-        """Generate ALE-value array and null distribution from list of contrasts.
+        """Generate ALE-value array and null distribution from a list of contrasts.
 
         For ALEs on the original dataset, computes the null distribution.
         For permutation ALEs and all SCALEs, just computes ALE values.
@@ -675,8 +663,6 @@ class SCALE(CBMAEstimator):
             ma_values = self.kernel_transformer.transform(
                 data, masker=self.masker, return_type="sparse"
             )
-        elif isinstance(data, list):
-            ma_values = self.masker.transform(data)
         elif isinstance(data, (np.ndarray, sparse._coo.core.COO)):
             ma_values = data
         else:
--- a/nimare/meta/cbma/ale.py
+++ b/nimare/meta/cbma/ale.py
@@ -243,7 +243,7 @@ class ALE(CBMAEstimator):
             n_mask_voxels = np.count_nonzero(mask_data)
 
             n_exp = ma_values.shape[0]
-            n_bins = bin_edges.shape[0] - 1
+            n_bins = bin_centers.shape[0]
             ma_hists = np.zeros((n_exp, n_bins))
             data = ma_values.data
             coords = ma_values.coords
--- a/nimare/meta/cbma/ale.py
+++ b/nimare/meta/cbma/ale.py
@@ -179,7 +179,7 @@ class ALE(CBMAEstimator):
         """
         if isinstance(ma_maps, list):
             ma_values = self.masker.transform(ma_maps)
-        elif isinstance(ma_maps, sparse._coo.core.COO):
+        elif isinstance(ma_maps, (np.ndarray, sparse._coo.core.COO)):
             ma_values = ma_maps
         else:
             raise ValueError(f"Unsupported data type '{type(ma_maps)}'")
@@ -189,7 +189,11 @@ class ALE(CBMAEstimator):
         # Assuming values of 0, .001, .002, etc., bins are -.0005-.0005, .0005-.0015, etc.
         INV_STEP_SIZE = 100000
         step_size = 1 / INV_STEP_SIZE
-        max_ma_values = ma_values.max(axis=[1, 2, 3]).todense()
+        if isinstance(ma_values, sparse._coo.core.COO):
+            # Need to convert to dense because np.ceil is too slow with sparse
+            max_ma_values = ma_values.max(axis=[1, 2, 3]).todense()
+        else:
+            max_ma_values = np.max(ma_values, axis=1)
 
         # round up based on resolution
         max_ma_values = np.ceil(max_ma_values * INV_STEP_SIZE) / INV_STEP_SIZE
@@ -215,12 +219,16 @@ class ALE(CBMAEstimator):
         """
         if isinstance(ma_maps, list):
             ma_values = self.masker.transform(ma_maps)
-        elif isinstance(ma_maps, sparse._coo.core.COO):
+        elif isinstance(ma_maps, (np.ndarray, sparse._coo.core.COO)):
             ma_values = ma_maps
         else:
             raise ValueError(f"Unsupported data type '{type(ma_maps)}'")
 
         assert "histogram_bins" in self.null_distributions_.keys()
+
+        def just_histogram(*args, **kwargs):
+            """Collect the first output (weights) from numpy histogram."""
+            return np.histogram(*args, **kwargs)[0].astype(float)
 
         # Derive bin edges from histogram bin centers for numpy histogram function
         bin_centers = self.null_distributions_["histogram_bins"]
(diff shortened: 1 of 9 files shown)


nimare/meta/cbma/ale.py
```diff
@@ -219,7 +236,31 @@ def just_histogram(*args, **kwargs):
         bin_edges = bin_centers - (step_size / 2)
         bin_edges = np.append(bin_centers, bin_centers[-1] + step_size)
 
-        ma_hists = np.apply_along_axis(just_histogram, 1, ma_values, bins=bin_edges, density=False)
+        if isinstance(ma_values, sparse._coo.core.COO):
+            masker = self.dataset.masker if not self.masker else self.masker
+            mask = masker.mask_img
+            mask_data = mask.get_fdata().astype(bool)
+            n_mask_voxels = np.count_nonzero(mask_data)
```
tsalo, line 243: I'm not sure how well this generalizes to other masker types, but I've opened an issue requesting a new attribute for masker objects with this info: https://github.com/nilearn/nilearn/issues/3296

nimare/meta/cbma/ale.py
```diff
@@ -219,7 +236,31 @@ def just_histogram(*args, **kwargs):
...
+
+            n_exp = ma_values.shape[0]
+            n_bins = bin_edges.shape[0] - 1
+            ma_hists = np.zeros((n_exp, n_bins))
+            data = ma_values.data
+            coords = ma_values.coords
+            for exp_idx in range(n_exp):
+                study_ma_values = data[coords[0, :] == exp_idx]
```
tsalo, line 251:
```suggestion
                # The first column of coords is the fourth dimension of the dense array
                study_ma_values = data[coords[0, :] == exp_idx]
```

nimare/meta/cbma/base.py
```diff
@@ -156,6 +157,7 @@ def _fit(self, dataset):
         ma_values = self._collect_ma_maps(
             coords_key="coordinates",
             maps_key="ma_maps",
+            return_type="sparse",
```
tsalo, line 160: If we are enforcing a sparse array here, will `_compute_null_approximate`, `_compute_summarystat_est`, or `_determine_histogram_bins` ever actually encounter any of the other possible datatypes? If not, then I think there's some spaghetti code we can simplify. If yes, then I'd like to try to document the circumstances where that would happen.

nimare/meta/kernel.py
```diff
@@ -309,40 +311,27 @@ def __init__(self, fwhm=None, sample_size=None):
...
-                sample_size = data.sample_size.astype(float).values[0]
-
-            if self.fwhm is not None:
-                assert np.isfinite(self.fwhm), "FWHM must be finite number"
-                if self.fwhm not in kernels.keys():
-                    _, kern = get_ale_kernel(mask, fwhm=self.fwhm)
-                    kernels[self.fwhm] = kern
-                else:
-                    kern = kernels[self.fwhm]
+        use_dict = True
```
tsalo, line 317: Thanks for including this! This is the only improvement from #709 that I wanted to get merged, so I'm glad I don't need to open a separate PR.

tsalo (commented):
I noticed a couple of commented-out lines that will to be removed before this is merged. I had a few comments/suggestions as well. Thanks for this!

nimare/meta/utils.py
```diff
@@ -397,7 +397,7 @@ def _convolve_sphere(kernel, peaks):
     return kernel_data
 
 
-def compute_ale_ma(shape, ijk, kernel):
+def compute_ale_ma(mask, ijks, exp_idx, sample_sizes, kernels, use_dict):
```
tsalo, line 400: I'd like to compare this to https://github.com/tsalo/NiMARE/blob/49191a0a4b10a2ac522828c9e558f279d5c6d6ea/nimare/meta/utils.py#L400-L438 before we merge.

nimare/meta/cbma/base.py
JulioAPeraza: Yeah, I think once we get all tests passing using the sparse array I will remove support for the other datatypes.

tsalo: Is the PR ready for this, or are you planning to do more before removing the other datatypes?

JulioAPeraza: I think it's almost ready. The [example figures](https://nimare--725.org.readthedocs.build/en/725/auto_examples/02_meta-analyses/01_plot_cbma.html#sphx-glr-auto-examples-02-meta-analyses-01-plot-cbma-py) in the documentation are similar to the ones in the [master](https://nimare.readthedocs.io/en/latest/auto_examples/02_meta-analyses/01_plot_cbma.html#sphx-glr-auto-examples-02-meta-analyses-01-plot-cbma-py) branch but are not the same. I need to look at the code to find what I'm missing, then rerun some profiling checks for ALE, KDA, and MKDADensity.

tsalo: Ah, thanks for the clarification. In that case, I'll wait until you request my review before looking through everything.

JulioAPeraza: I think I am ready to remove the support for other data types. I started doing that this morning but some tests were failing because they were created to work with numpy array. Should we remove those tests in that case?

I can also provide a table with the profiling results `ALE`, `KDA`, and `MKDADensity` against the master branch. At first glance, the use of sparse directly in the `CBMAEstimator` makes everything more efficient in terms of memory and computation time.
Also, `_run_fwe_permutation` efficiency got improved by using sparse.

JulioAPeraza: @tsalo to remove return_type="array" in favor of return_type="sparse" I would need to implement sparse in ALESubtraction too. Do you think that can be done here or in another PR?

tsalo: I think it's worth implementing it in this PR.

JulioAPeraza:
@tsalo I believe this is ready for review. Now, all CBMA methods support sparse arrays.

nimare/meta/cbma/ale.py
```diff
@@ -151,6 +152,18 @@ def __init__(
...
+        # np.array type is used by _determine_histogram_bins to calculate max_poss_ale
+        if isinstance(stat_values, sparse._coo.core.COO):
+            masker = self.dataset.masker if not self.masker else self.masker
+            mask_data = masker.mask_img.get_fdata().astype(bool)
+
+            stat_values = stat_values.todense().reshape(-1)  # Indexing a .reshape(-1) is faster
+            stat_values = stat_values[mask_data.reshape(-1)]
+
+            # This is used by _compute_null_approximate
+            self.n_mask_voxels = stat_values.shape[0]
```
tsalo, line 165:
```suggestion
            self.__n_mask_voxels = stat_values.shape[0]
```
I'm thinking this should either be private (leading underscores) or considered a fitted, data-dependent parameter (trailing underscore).

nimare/meta/cbma/ale.py
```diff
@@ -176,7 +189,12 @@ def _determine_histogram_bins(self, ma_maps):
         # Assuming values of 0, .001, .002, etc., bins are -.0005-.0005, .0005-.0015, etc.
         INV_STEP_SIZE = 100000
         step_size = 1 / INV_STEP_SIZE
-        max_ma_values = np.max(ma_values, axis=1)
+        if isinstance(ma_values, sparse._coo.core.COO):
+            # Need to convert to dense because np.ceil is too slow with sparse
+            max_ma_values = ma_values.max(axis=[1, 2, 3]).todense()
+        else:
+            max_ma_values = np.max(ma_values, axis=1)
```
tsalo, line 196: Can you add a comment with info about _when_ this would be reached? I assume it would be reached when `ma_maps` is a list above, but if there are other cases then I'd like those documented as well.

nimare/meta/cbma/ale.py
```diff
@@ -584,25 +632,23 @@ def _fit(self, dataset):
...
         rand_xyz = self.xyz[rand_idx, :]
         iter_xyzs = np.split(rand_xyz, rand_xyz.shape[1], axis=1)
 
-        perm_scale_values = np.memmap(
-            self.memmap_filenames[1],
-            dtype=stat_values.dtype,
-            mode="w+",
-            shape=(self.n_iters, stat_values.shape[0]),
-        )
+        perm_scale_values = np.zeros((self.n_iters, stat_values.shape[0]), dtype=stat_values.dtype)
```
tsalo, line 651: I think we're going to need to profile this.

nimare/meta/cbma/mkda.py
```diff
@@ -465,19 +487,34 @@ def _run_fwe_permutation(self, iter_xyz1, iter_xyz2, iter_df1, iter_df2, conn, v
...
+        if isinstance(n_selected_active_voxels, sparse._coo.core.COO):
+            masker = self.masker
+            mask_data = masker.mask_img.get_fdata().astype(bool)
+
+            # Indexing the sparse array is slow, perform masking in the dense array
+            n_selected_active_voxels = n_selected_active_voxels.todense().reshape(-1)
+            n_selected_active_voxels = n_selected_active_voxels[mask_data.reshape(-1)]
+
         del temp_ma_maps1
+        gc.collect()
```
tsalo, line 504: Should this be called after every `del`?

nimare/meta/kernel.py
```diff
@@ -206,9 +206,9 @@ def transform(self, dataset, masker=None, return_type="image"):
         # Loop over exp ids since sparse._coo.core.COO is not iterable
         for i_exp, id_ in enumerate(transformed_maps[1]):
             if isinstance(transformed_maps[0][i_exp], sparse._coo.core.COO):
+                # This step is slow, but it is here just in case user want a
+                # return_type = "array", "image", or "dataset"
```
tsalo, line 210: Hm... should it be `if isinstance(...) and return_type != "sparse"`?

nimare/meta/utils.py
```diff
@@ -407,55 +427,130 @@ def compute_ale_ma(shape, ijk, kernel):
...
-        Shape of brain image + buffer. Typically (91, 109, 91) + (30, 30, 30).
-    ijk : array-like
+    mask : img_like
+        Mask to extract the MA maps shape (typically (91, 109, 91)) and voxel dimension.
+        The mask is applied to the coordinates before creating the kernel_data.
+    ijks : array-like
         Indices of foci. Each row is a coordinate, with the three columns
         corresponding to index in each of three dimensions.
-    kernel : array-like
+    kernels : array-like, or None, optional
```
tsalo, line 436: I believe this is a dict now.

nimare/meta/cbma/ale.py
```diff
@@ -166,7 +179,7 @@ def _determine_histogram_bins(self, ma_maps):
         """
         if isinstance(ma_maps, list):
             ma_values = self.masker.transform(ma_maps)
-        elif isinstance(ma_maps, np.ndarray):
+        elif isinstance(ma_maps, (np.ndarray, sparse._coo.core.COO)):
```
jdkent, line 182: could you add the possible types `ma_maps` can be in the docstring of `_determine_histogram_bins`? and maybe how many dimensions the array can have.

nimare/meta/cbma/ale.py
```diff
@@ -151,6 +152,18 @@ def __init__(
 
     def _compute_summarystat_est(self, ma_values):
         stat_values = 1.0 - np.prod(1.0 - ma_values, axis=0)
+
+        # np.array type is used by _determine_histogram_bins to calculate max_poss_ale
+        if isinstance(stat_values, sparse._coo.core.COO):
+            masker = self.dataset.masker if not self.masker else self.masker
+            mask_data = masker.mask_img.get_fdata().astype(bool)
```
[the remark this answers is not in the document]
JulioAPeraza: Here I try to follow the way the `KernelTransformer.transform()` creates the mask_data. Do you think we can easily add support for other makers here, or that could be addressed in a separate PR?

nimare/meta/cbma/mkda.py
```diff
@@ -202,21 +229,18 @@ def _compute_null_approximate(self, ma_maps):
         This method adds two entries to the null_distributions_ dict attribute:
```
jdkent, line 229: This doc string may be out of date now, `ma_maps` can be sparse arrays as well.

nimare/meta/cbma/mkda.py
```diff
@@ -1047,7 +1084,21 @@ def _compute_summarystat_est(self, ma_values):
             OF values. One value per voxel.
```
jdkent, line 1084: this doc string should be updated too to include sparse arrays as input

nimare/meta/cbma/mkda.py
```diff
@@ -1064,7 +1115,7 @@ def _determine_histogram_bins(self, ma_maps):
         """
```
jdkent, line 1115: could you update this docstring to include what datatypes `ma_maps` can be?

nimare/meta/cbma/mkda.py
```diff
@@ -1121,7 +1174,7 @@ def _compute_null_approximate(self, ma_maps):
...
```
jdkent, line 1174: this doc string should be updated to include sparse arrays

nimare/meta/cbma/mkda.py
JulioAPeraza: For MKDAChi2 I had to call this because `del` wasn't freeing up the memory there. I will profile this line to see if it's needed here.

jdkent (changes requested):
Looking great! I have a couple places where docstrings should likely be updated, and I'm testing locally to if some of the functions are ever called with `np.array` since the switch to sparse arrays.

nimare/meta/kernel.py
JulioAPeraza: I think if `return_type == "sparse"` this line won't be reached, the transformer will return the transformed_maps in line 202:

https://github.com/JulioAPeraza/NiMARE/blob/bb9815891727cac66cbda5a80da60d7e0ad42c9b/nimare/meta/kernel.py#L202-L203

nimare/meta/utils.py
JulioAPeraza: We're creating the empty dictionary inside `compute_ale_ma` if `use_dict is True`. I think `kernels` will be a 3D array when `use_dict is False` and `sample_sizes is None`

nimare/meta/cbma/mkda.py
JulioAPeraza: I believe `ma_maps` aren't used by the `MKDADensity._compute_null_approximate()`. I think the `ma_maps` parameter is here just to provide compatibility when the `_compute_null_approximate()` is called in the `CBMAEstimator`. 
What would you recommend doing in this case?

JulioAPeraza: How does this look?

```
Parameters
----------
ma_maps
    Modeled activation maps. Unused for this estimator.
```

nimare/meta/cbma/base.py
```diff
@@ -206,8 +207,10 @@ def _collect_ma_maps(self, coords_key="coordinates", maps_key="ma_maps", return_
...
         -------
-        ma_maps : :obj:`numpy.ndarray`
-            2D numpy array of shape (n_studies, n_voxels) with MA values.
+        ma_maps : :obj:`numpy.ndarray`, or :obj:`sparse._coo.core.COO`
+            Return a 2D numpy array of shape (n_studies, n_voxels) with MA values,
+            if pre-generated MA maps exist. Otherwise, a 4D sparse array of shape
+            (n_studies, mask.shape) with MA maps is returned.
         """
         if maps_key in self.inputs_.keys():
             LGR.debug(f"Loading pre-generated MA maps ({maps_key}).")
```
jdkent, line 216: I think if this is a simple voxelwise niftimasker (not a labels masker or some sort of atlas masker), this should still return a sparse matrix, not a dense one. Inconsistent behavior to re-run the `fit` method twice and have the data types be different for each run.

jdkent:
these lines no longer need to be in the code: https://github.com/neurostuff/NiMARE/blob/ac57801b288675e97b91d4356ab8aa83ced9edfc/nimare/tests/test_meta_ale.py#L48-L53

since we have removed memory limit

nimare/meta/cbma/ale.py
JulioAPeraza: Do you think we should keep the memap array here? 
I profiled the computation time is faster and the memory usage is not that different.

nimare/meta/cbma/base.py
jdkent: I'm looking into it, we can potentially extend/catch the niftimasker class to allow for a custom transformation, here is some draft code for the masking procedure (don't know if it currently prevents entire array from being loaded at once)
```
nonzero_idx = imgs.dataobj.nonzero()
all_data = imgs.dataobj[nonzero_idx]
all_coords = np.vstack(nonzero_idx)
sparse_data = sparse.COO(all_coords, all_data, shape=imgs.shape)
masked_sparse_data = sparse_data[mask_img_.dataobj.nonzero()].T
```

nimare/meta/cbma/mkda.py
jdkent: okay, that sounds good to me for now

nimare/meta/cbma/base.py
JulioAPeraza: I ended up using the same approach we have in `compute_ale_ma` and in `compute_kda_ma`. Now loading the pre-generated MA maps is faster and memory efficient. 
Thanks!

nimare/meta/cbma/mkda.py
tsalo: Thanks!

nimare/meta/kernel.py
tsalo: Ah, that makes sense.

nimare/meta/cbma/ale.py
JulioAPeraza: I'm deleting these lines since `_collect_ma_maps` always returns a sparse array now.

nimare/meta/cbma/ale.py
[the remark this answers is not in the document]
JulioAPeraza: I have added the check for the masker to `_preprocess_input` where the masker is called for the first time. In the initialization of the classes `dataset` doesn't exist yet, so I didn't know where to add the check.

nimare/meta/cbma/ale.py
JulioAPeraza: Done!

nimare/meta/cbma/mkda.py
JulioAPeraza: Done!

nimare/meta/cbma/ale.py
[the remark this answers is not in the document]
tsalo: I would recommend putting it in `_fit`, like we have for the IBMAEstimators:

https://github.com/neurostuff/NiMARE/blob/ac57801b288675e97b91d4356ab8aa83ced9edfc/nimare/meta/ibma.py#L169-L175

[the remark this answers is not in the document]
tsalo: But yes, I agree that it shouldn't be checked at initialization, since we don't have access to the Dataset yet.

nimare/meta/cbma/mkda.py
JulioAPeraza: I'm removing the `gc.collect()`, `del` is working fine here.

tsalo: That is so confusing! I have no clue why it's necessary for some things, but not others... Thanks for checking it in this case though.

nimare/meta/cbma/ale.py
JulioAPeraza:
```python
from nimare.dataset import Dataset
from nimare.meta.cbma.ale import SCALE
from nimare.utils import get_resource_path, vox2mm

dset_file = os.path.join(get_resource_path(), "nidm_pain_dset.json")
dset = Dataset(dset_file)

xyz = vox2mm(
    np.vstack(np.where(dset.masker.mask_img.get_fdata())).T,
    dset.masker.mask_img.affine,
)

meta = SCALE(xyz=xyz, n_iters=10)
results = meta.fit(dset)

```

This is what I got for 10 iterations:

- `master`:
    - 10/10 [00:03<00:00,  2.95it/s]
    - 228483/228483 [04:32<00:00, 839.13it/s]
    - Current memory usage is 18.818837MB; Peak was 231.680765MB
    - Computation time: 276.62519907951355

- `ale-sparse`:
    - 10/10 [00:06<00:00,  1.65it/s]
    - 228483/228483 [04:11<00:00, 908.33it/s]
    - Current memory usage is 29.558014MB; Peak was 198.31731MB
    - Computation time: 260.77495074272156

nimare/meta/cbma/ale.py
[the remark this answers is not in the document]
JulioAPeraza: Done!

nimare/meta/cbma/ale.py
```diff
@@ -151,32 +156,44 @@ def __init__(
...
 
     def _determine_histogram_bins(self, ma_maps):
         """Determine histogram bins for null distribution methods.
 
         Parameters
         ----------
-        ma_maps
+        ma_maps : :obj:`sparse._coo.core.COO`
+            MA maps.
+            The ma_maps can be a 4d sparse array of MA maps,
```
tsalo, line 180: I think this line can be removed, since it's the only possible type now.

nimare/meta/cbma/ale.py
```diff
@@ -189,8 +206,9 @@ def _compute_null_approximate(self, ma_maps):
 
         Parameters
         ----------
-        ma_maps : list of imgs or numpy.ndarray
+        ma_maps : :obj:`sparse._coo.core.COO`
             MA maps.
+            The ma_maps can be 4D sparse array of MA maps,
```
tsalo, line 211: Can be removed as well.

nimare/meta/cbma/base.py
```diff
@@ -151,6 +154,13 @@ def _fit(self, dataset):
         """
         self.dataset = dataset
         self.masker = self.masker or dataset.masker
+
+        if not isinstance(self.masker, NiftiMasker):
```
tsalo, line 158: Can you cover this in one of the tests?

nimare/meta/cbma/mkda.py
```diff
@@ -166,57 +169,80 @@ def _compute_weights(self, ma_values):
...
+            study_ma_values = data[coords[0, :] == exp_idx]
+
+            n_nonzero_voxels = study_ma_values.shape[0]
+            n_zero_voxels = self.__n_mask_voxels - n_nonzero_voxels
+
+            prop_active[exp_idx] = np.mean(np.hstack([study_ma_values, np.zeros(n_zero_voxels)]))
+
         self.null_distributions_["histogram_bins"] = np.arange(len(prop_active) + 1, step=1)
+        # To speed things up in _compute_null_approximate, we save the means too,
+        self.null_distributions_["histogram_means"] = prop_active
```
tsalo, line 219: This new null distribution needs to be documented in a few docstrings, I think.

nimare/meta/cbma/mkda.py
```diff
@@ -1034,39 +1076,49 @@ def _compute_summarystat_est(self, ma_values):
...
     def _determine_histogram_bins(self, ma_maps):
         """Determine histogram bins for null distribution methods.
 
         Parameters
         ----------
-        ma_maps
-            Modeled activation maps. Unused for this estimator.
+        ma_maps : :obj:`sparse._coo.core.COO`
+            MA maps.
+            The ma_maps can be a 4d sparse array of MA maps,
```
tsalo, line 1115: Line can be removed.

nimare/meta/cbma/mkda.py
```diff
@@ -1111,33 +1165,41 @@ def _compute_null_approximate(self, ma_maps):
 
         Parameters
         ----------
-        ma_maps : list of imgs or numpy.ndarray
+        ma_maps : :obj:`sparse._coo.core.COO`
             MA maps.
+            The ma_maps can be a 4d sparse array of MA maps,
```
tsalo, line 1170: Line can be removed.

nimare/meta/utils.py
```diff
@@ -358,104 +364,201 @@ def _convolve_sphere(kernel, peaks):
...
     """Generate ALE modeled activation (MA) maps.
 
     Replaces the values around each focus in ijk with the contrast-specific
     kernel. Takes the element-wise maximum when looping through foci, which
     accounts for foci which are near to one another and may have overlapping
     kernels.
 
+    .. versionchanged:: 0.0.12
+
+        * This function now Return 4D sparse array.
```
tsalo, line 432:
```suggestion
        * This function now returns a 4D sparse array.
```

tsalo (changes requested):
I think the core changes are ready. There are just a few docstring changes that I think need to be made, as well as a small improvement to one of the CBMA tests (i.e., covering the new masker type check).

nimare/meta/cbma/base.py
JulioAPeraza: Do you know of any case in which this line is reached? If I provide a non-NiftiMasker to the estimator, it will get converted to a NiftiMasker by `get_masker(mask)`.

tsalo: `get_masker` should preserve the type of any nilearn masker (e.g., `NiftiMasker`, `NiftiLabelsMasker`, `NiftiMapsMasker`), so you _should_ be able to feed in a NiftiLabelsMasker and it would then raise this exception.

JulioAPeraza: Great. Thanks!

tsalo (approved):
Everything looks good to me, thanks!

Issue #614, opened 2021-12-08
Convert GCLDAModel into an Annotator


tsalo:
## Summary

GCLDAModel currently accepts a count DataFrame and a coordinate DataFrame. If we convert it into an Annotator, it can take a Dataset directly and return a new one.

PR #742, opened 2022-08-17
Enhance LDA annotator

JulioAPeraza:
Closes None.

- Add the parallelization option and use it in the `LatentDirichletAllocation` model.
- Switch `learning_method` from `online` to the default `batch`.

### DIFF
--- a/nimare/annotate/lda.py
+++ b/nimare/annotate/lda.py
@@ -70,7 +70,7 @@ class LDAModel(NiMAREBase):
         self.model = LatentDirichletAllocation(
             n_components=n_topics,
             max_iter=max_iter,
-            learning_method="online",
+            learning_method="batch",
             doc_topic_prior=alpha,
             topic_word_prior=beta,
             n_jobs=n_cores,
--- a/nimare/annotate/lda.py
+++ b/nimare/annotate/lda.py
@@ -5,6 +5,7 @@ from sklearn.decomposition import LatentDirichletAllocation
 
 from nimare.annotate.text import generate_counts
 from nimare.base import NiMAREBase
+from nimare.utils import _check_ncores
 
 
 class LDAModel(NiMAREBase):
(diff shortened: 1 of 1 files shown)


JulioAPeraza:
@tsalo Do you think it would be Ok to add support for running LDA with `counts_df` as an input in this PR?

Also, I was wondering if it would be possible to initialize and fit both the LDA and GCLDA annotators more uniformly. For example, 
Init LDAModel with:
```
n_topics, max_iter=1000, alpha=None, beta=0.001, text_column="abstract", n_cores=1
```
Init GCLDAModel with
```
n_topics=100, n_iters=10000, loglikely_freq=10, n_regions=2, symmetric=True, alpha=0.1, beta=0.01, gamma=0.01, delta=1.0, dobs=25, roi_size=50.0, seed_init=1
```
And fitting the models with `fit(self, dset, counts_df=None) `.

tsalo:
>Do you think it would be Ok to add support for running LDA with counts_df as an input in this PR?

Sure, I think that's reasonable.

>Also, I was wondering if it would be possible to initialize and fit both the LDA and GCLDA annotators more uniformly.

Ultimately, we'll want to create an Annotator class with a standard interface (see #488 and #614), but the inits you propose look good. I haven't done a side-by-side comparison to see what you've changed, though.

>And fitting the models with fit(self, dset, counts_df=None)

That seems solid. I don't know how Annotators will look yet, so I don't have a strong opinion about this change, with the caveat that we may end up changing it once we have Annotators implemented.

BTW, the RTD CI failure is just a one-off, random thing (see #418). You can just request a rebuild and it will probably pass the next time- or at least get to a point where any _real_ bugs show up.

JulioAPeraza:
> That seems solid. I don't know how Annotators will look yet, so I don't have a strong opinion about this change, with the caveat that we may end up changing it once we have Annotators implemented.

Sounds great! In that case, I think we can leave those changes to the PR that adds the Annotator class.

tsalo (approved):
Sorry for the delay. LGTM!

PR #743, opened 2022-08-23
Fix import in download_neurosynth example

PTDZ:
Closes # .

- Line 77 error: "NameError: name 'download_abstracts' is not defined".
- Adding "extract" before "download_abstracts" solves the issue.


tsalo:
Thanks for catching this @PTDZ! Merging now.

PR #744, opened 2022-08-24
Optimize GCLDAModel with numba

JulioAPeraza:
Closes None.

Currently, the `GCLDA` model is slow, mainly when the vocabulary is large (for example for the 3228 terms from Neurosynth). `_update_word_topic_assignments` takes ~73 seconds per iteration and `_update_peak_assignments` takes ~52 seconds per iteration for a total of ~125 seconds of `_update()` per iteration. Compiling and running these functions with `numba` reduce these times to 2, 4, and 6 seconds per iteration respectively.

- Convert `_update_word_topic_assignments` and `_update_peak_assignments` to static methods, and add a decorator to compile and run them with `numba` in `nopython` mode.

### DIFF
--- a/nimare/annotate/gclda.py
+++ b/nimare/annotate/gclda.py
@@ -6,6 +6,7 @@ import nibabel as nib
 import numpy as np
 import pandas as pd
 from nilearn._utils import load_niimg
+from numba import jit
 from scipy.stats import multivariate_normal
 
 from nimare.base import NiMAREBase
@@ -408,11 +409,49 @@ class GCLDAModel(NiMAREBase):
 
         LGR.debug(f"Iter {self.iter:04d}: Sampling z")
         self.seed += 1
-        self._update_word_topic_assignments(self.seed)  # Update z-assignments
+        (
+            self.topics["wtoken_topic_idx"],
+            self.topics["n_word_tokens_word_by_topic"],
+            self.topics["total_n_word_tokens_by_topic"],
+            self.topics["n_word_tokens_doc_by_topic"],
+        ) = self._update_word_topic_assignments(
+            self.topics["wtoken_topic_idx"],
+            self.topics["n_word_tokens_word_by_topic"],
+            self.topics["total_n_word_tokens_by_topic"],
+            self.topics["n_word_tokens_doc_by_topic"],
+            self.topics["n_peak_tokens_doc_by_topic"],
+            np.array(self.data["wtoken_doc_idx"]),
+            np.array(self.data["wtoken_word_idx"]),
+            len(self.vocabulary),
+            self.params["beta"],
+            self.params["gamma"],
+            self.seed,
+        )  # Update z-assignments
 
         LGR.debug(f"Iter {self.iter:04d}: Sampling y|r")
         self.seed += 1
-        self._update_peak_assignments(self.seed)  # Update y-assignments
+        # Retrieve p(x|r,y) for all subregions
+        peak_probs = self._get_peak_probs(self)
+        (
+            self.topics["n_peak_tokens_region_by_topic"],
+            self.topics["n_peak_tokens_doc_by_topic"],
+            self.topics["peak_topic_idx"],
+            self.topics["peak_region_idx"],
+        ) = self._update_peak_assignments(
+            self.topics["n_peak_tokens_region_by_topic"],
+            self.topics["n_peak_tokens_doc_by_topic"],
+            self.topics["peak_topic_idx"],
+            self.topics["peak_region_idx"],
+            self.topics["n_word_tokens_doc_by_topic"],
+            peak_probs,
+            np.array(self.data["ptoken_doc_idx"]),
+            self.params["n_regions"],
+            self.params["n_topics"],
+            self.params["alpha"],
+            self.params["delta"],
+            self.params["gamma"],
+            self.seed,
+        )  # Update y-assignments
 
         LGR.debug(f"Iter {self.iter:04d}: Updating spatial params")
         self._update_regions()  # Update gaussian estimates for all subregions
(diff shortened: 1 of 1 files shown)


tsalo:
With the changes to the two update methods, it seems like they could be converted to functions instead. Is there any benefit to using `@staticmethod` over making them functions?

JulioAPeraza:
You're right. The two update methods can be converted to functions. I don't see any benefit in using `staticmethod` in this case.

nimare/tests/test_annotate_gclda.py
```diff
@@ -7,12 +7,12 @@
...
 from nimare import annotate, decode
 
+# Disable numba execution to allow pytest-cov to detect coverage in functions with @jit
+config.DISABLE_JIT = True
+
 
 def test_gclda_symmetric(testdata_laird):
     """A smoke test for GCLDA with symmetric regions."""
-    # Disable numba execution to allow pytest-cov to detect coverage in functions with @jit
-    config.DISABLE_JIT = True
```
tsalo: Can you do something like `with config.DISABLE_JIT`?

JulioAPeraza: I tried running the GCLDA test with `with config.DISABLE_JIT`, and it didn't work.
It looks like disabling jit for GCLDA and enabling it again for the other tests that use sparse is working.

nimare/annotate/gclda.py
```diff
@@ -14,6 +15,247 @@
...
+        # [R x T] array containing the proportional probability of all y/r combinations
+        probs_pdf = p_x_subregions * p_region_g_doc * p_peak_g_topic
+
+        # Convert from a [R x T] matrix into a [R*T x 1] array we can sample from
+        probs_pdf = np.reshape(probs_pdf, (n_regions * n_topics))
+
+        # Normalize the sampling distribution
+        probs_pdf = probs_pdf / np.sum(probs_pdf)
+        # Float precision issue in Numba: https://github.com/numba/numba/issues/3426
+        probs_pdf = np.trunc(probs_pdf * (10**12)) / (10**12)
```
JulioAPeraza, line 228: This line will need to be removed in the future. It is ensuring the sum never goes over 1.0 so that we don't get this error in numba, but this truncation is causing the maps on the [example](https://nimare--744.org.readthedocs.build/en/744/auto_examples/03_annotation/04_plot_gclda.html#look-at-topics) to look different to the ones from previous versions.

nimare/tests/test_annotate_gclda.py
JulioAPeraza: The Codecov report still does not cover functions with `@jit`. 
The issue is that disabling numba globally produces an error in the functions that use sparse arrays.

nimare/annotate/gclda.py
JulioAPeraza: I was running some tests with Neurosynth and Neuroquery, and got the error again:

```
File "<...>/lib/python3.9/site-packages/nimare/annotate/gclda.py", line 622, in fit
    self._update(loglikely_freq=loglikely_freq)
  File "<...>/lib/python3.9/site-packages/nimare/annotate/gclda.py", line 681, in _update
    ) = _update_peak_assignments(
  File "<...>/lib/python3.9/site-packages/numba/cpython/randomimpl.py", line 833, in binomial_impl
    raise ValueError("binomial(): p outside of [0, 1]")
ValueError: binomial(): p outside of [0, 1]
```

JulioAPeraza:
Given these limitations (precision issue with sum, and codecov of jitted functions), do you think we should try Cython for these two functions?

tsalo:
I think @jdkent or @adelavega would be better able to answer that.

nimare/annotate/gclda.py
jdkent: would np.clip help here? 
during normalization, you could do this:
`probs_pdf = (probs_pdf / np.sum(probs_pdf)).clip(0, 1)`

JulioAPeraza: I tried using `clip` but it still raised the same error. I think it has to do with the way `numba` has the sum implemented.

nimare/annotate/gclda.py
```diff
@@ -14,6 +15,247 @@
 LGR = logging.getLogger(__name__)
 
 
+@jit(nopython=True, cache=True)
```
jdkent, line 18: would codecov track this function if `nopython=False`?

JulioAPeraza: I'm not sure. Let me try this out.

nimare/annotate/gclda.py
```diff
@@ -14,6 +15,247 @@
...
+        # Normalize the sampling distribution
+        probs_pdf = probs_pdf / np.sum(probs_pdf)
+        # Float precision issue in Numba: https://github.com/numba/numba/issues/3426
+        probs_pdf = np.trunc(probs_pdf * (10**12)) / (10**12)
+
+        # Sample a single element (corresponding to a y_i and c_i assignment for the ptoken)
+        # from the sampling distribution
+        # Returns a binary [1 x R*T] vector with a '1' in location that was sampled
+        # and zeros everywhere else
+        assignment_vec = np.random.multinomial(1, probs_pdf)
```
tsalo, line 234: The documentation says to "use the `multinomial` method of a `default_rng()` instance instead". If we're lucky, that might also be faster?

JulioAPeraza: I have tried different combinations with:
```
rng = np.random.default_rng()
assignment_vec = rng.multinomial(1, probs_pdf)
```
and everything has failed in numba.

Using the correction proposed in this issue `https://github.com/numba/numba/issues/3426`, the current `assignment_vec = np.random.multinomial(1, probs_pdf)` is working fine without truncating `probs_pdf`.

adelavega: Ah that's good to hear. I'd love to hear what the speed up is using numba on a large-ish dataset.

JulioAPeraza:
@adelavega Running times per iteration for 200 topics:

| Function | Branch | Neurosynth (14369 studies, 3228 words) | NeuroQuery (13459 studies, 6145 words) |
| -------- | :--------: | :-----------: | :-----------: |
| `_update_word_topic_assignments` | master | ~96s | ~1342s |
| `_update_word_topic_assignments` | gclda-numba | ~10s | ~149s |
| `_update_peak_assignments` | master | ~112s | ~96s |
| `_update_peak_assignments` | gclda-numba | ~55s | ~50s |

Typically we have 10,000 iterations, so to train GCLDA on Neurosynth with `master` would take ~25 days, which can be reduced to ~7.5 days with `gclda-numba`.

JulioAPeraza:
The latest changes showed improvement in `_update_peak_assignments`.

Creating chunks (`n_chunks = n_cores`) of the data and running each chunk in separate processors reduced the overhead and the number of `delayed()` calls by the embarrassingly parallel routine. Right now we have 2 nested for loops, one iterates over chunks and it's embarrassingly parallelized, and the other iterates within chunks extending the duration of the task that is submited to Parallel. These changes, besides reducing the overhead, reduce the time that it takes to dispatch so many calls (in our case 500,000) of very fast tasks to workers.

However, this is only scalable to a small number of CPUs. I got small linear scalability for only <= 4 CPUs in a local machine and <= 8 CPUs in an HPC. 

What do you all think? Should we:
1. keep numba support for `_update_word_topic_assignments` and wait for numba to fix the issue with the multinomial generator to add support for `_update_peak_assignments`, or
2. keep both numba support for `_update_word_topic_assignments` and Parallel for `_update_peak_assignments`, with a warning to users that the algorithm is only scalable to a small number of CPUs (4-8)?

I think number 1 is good, since using numba for `_update_word_topic_assignments` already makes GCLDA 10x faster.

adelavega:
@JulioAPeraza awesome! q: why is it only scalable to a small number of CPUs? I don't quite understand that.

JulioAPeraza:
@adelavega 
I'm not sure. I guess the scalability problem comes from the cost of communication added by having all the input arrays written to the disk (or copied to memory) in parallel. Although the increase in the number of CPUs reduces the calculation time, it may also increase the communication making the algorithm not scalable. 
I just realized that by default `Parallel` will dump into a file any array larger than 1 megabyte. I ran a couple of tests with `max_nbytes=None` and now the algorithm is at least weakly scalable to all the cores available in my laptop (8 CPUs). I still need to test this on the HPC with a bigger number of CPUs. With 8 CPUs the calculation time per iteration of _update goes down from ~112s to ~15s.

adelavega:
Ah, disk i/o would certainly explain it. I wonder using "threads" would also help: https://joblib.readthedocs.io/en/latest/parallel.html#thread-based-parallelism-vs-process-based-parallelism

JulioAPeraza:
I tested the scalability of the function with more CPUs in the HPC and it looks good. I also ran some tests using `threads` but it was slower than the default `loky`.

This PR is ready for review. Thanks!

adelavega:
Do you know if explicitly setting `parallel=True` would help the numba optimized `_update_peak_assignments`?

https://numba.readthedocs.io/en/stable/user/parallel.html

It looks like the main concern would be the same matrices are being access by the document loop, but it looks like it would be different parts of the arrays?

adelavega:
Otherwise, I took a closer look at the GCLDA algorithm, and this all looks reasonable to me. I may have some more nit-picky comments later, but perhaps @jdkent and @tsalo can help with those as well.

JulioAPeraza:
Thanks, @adelavega.

> Do you know if explicitly setting `parallel=True` would help the numba optimized `_update_peak_assignments`?

I tried `parallel=True` before in `_update_word_topic_assignments` and also in `_update_peak_assignments` running on our fork of numba, and I got a `Segmentation Fault` error. I will look into it early tomorrow.

nimare/annotate/gclda.py
```diff
@@ -405,14 +635,85 @@ def _update(self, loglikely_freq=1):
...
+        # Array of seeds for multi-core reproducibility
+        seeds = np.random.randint(1e8, size=n_ptoken)
+        chunks = [
+            [
+                self.data["ptoken_doc_idx"][chunk_i : chunk_i + chunk_size],
+                old_regions[chunk_i : chunk_i + chunk_size],
+                old_topics[chunk_i : chunk_i + chunk_size],
+                peak_probs[chunk_i : chunk_i + chunk_size, :, :],
+                seeds[chunk_i : chunk_i + chunk_size],
+            ]
```
jdkent, line 679: can this be a dictionary so that the arguments being sent to `_update_peak_assignments` are explicit?

nimare/annotate/gclda.py
```diff
@@ -405,14 +635,85 @@ def _update(self, loglikely_freq=1):
...
+            for chunk_i in range(0, n_ptoken, chunk_size)
+        ]
+
+        results = Parallel(n_jobs=self.n_cores, max_nbytes=None)(
+            delayed(_update_peak_assignments)(
+                *chunk,
```
jdkent, line 685: with above, this would now be `**chunk`

nimare/annotate/gclda.py
```diff
@@ -405,14 +635,85 @@ def _update(self, loglikely_freq=1):
...
+                self.topics["n_peak_tokens_region_by_topic"],
+                self.topics["n_peak_tokens_doc_by_topic"],
+                self.topics["n_word_tokens_doc_by_topic"],
+                self.params["n_regions"],
+                self.params["n_topics"],
+                self.params["alpha"],
+                self.params["delta"],
+                self.params["gamma"],
```
jdkent, line 693: a bit redundant, but now the arguments here should be named too (e.g., `gamma = self.params["gamma"]

jdkent (changes requested):
Looks good, just a minor suggestion for readability for me.

nimare/annotate/gclda.py
```diff
@@ -1,24 +1,247 @@
...
+        Updated topic assignment for word tokens.
+    nz_word_topic : (W x T) :obj:`numpy.ndarray`
+        Updated number of word-tokens assigned to each topic (T) per word-type (W).
+    nz_sum_topic : (1 x T) :obj:`numpy.ndarray`
+        Updated total number of word-tokens assigned to each topic (T) (across all docs).
+    nz_doc_topic : (D x T) :obj:`numpy.ndarray`
+        Updated number of word-tokens assigned to each topic (T) per document (D).
+    """
+    # --- Seed random number generator
+    np.random.seed(randseed)
```
adelavega, line 74: It's likely jit does something slightly different here, since the results change if I comment out `@jit` above.

nimare/annotate/gclda.py
```diff
@@ -405,14 +635,85 @@ def _update(self, loglikely_freq=1):
...
+            self.topics["n_word_tokens_word_by_topic"],
+            self.topics["total_n_word_tokens_by_topic"],
+            self.topics["n_word_tokens_doc_by_topic"],
+            self.topics["n_peak_tokens_doc_by_topic"],
+            np.array(self.data["wtoken_doc_idx"]),
+            np.array(self.data["wtoken_word_idx"]),
+            len(self.vocabulary),
+            self.params["beta"],
+            self.params["gamma"],
+            self.seed,
```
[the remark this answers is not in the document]
JulioAPeraza: Probably that explains the difference you saw because we are testing with a small number of iterations. I think `self.seed = self.params["seed_init"]` is used to initialize the `peak_topic` and the `peak_subregion` assignments, whereas `self.seed` is used to perform the sampling in `_update()`.

jdkent:
this was completed in #982, thank you @JulioAPeraza for your work on this that inspired the solutions in the other pull request.

PR #745, opened 2022-09-21
Optimize compute_kda_ma

liuzhenqi77:

- Optimize `compute_kda_ma`

I was trying to speedup `MKDAChi2()` and made some changes to the matrix operations in `compute_kda_ma`, which could half the running time. Please feel free to make any revisions, as I could only test under my specific case.

- `nonzero_idx` is not sorted after the change, but I guess it's not an issue.
- I did not fully test against `sum_overlap=True`, in which `counts` could be a list.

I also found that parallelize this loop with joblib could speedup the process by another small percent. So, I was wondering if there is a way to pre-calculate `ma_maps` for all the files in a dataset and reuse them. I saw that many functions support using saved MA maps internally through `self.inputs_["ma_maps"]`, but I did not find a way to patch into the input.

Thank you.

### DIFF
--- a/nimare/meta/utils.py
+++ b/nimare/meta/utils.py
@@ -116,13 +116,10 @@ def compute_kda_ma(
         all_spheres = _convolve_sphere(kernel, peaks)
 
         if sum_overlap:
-            # if sum_overlap, counts=list
             all_spheres, counts = unique_rows(all_spheres, return_counts=True)
             counts = counts * value
         else:
-            # if not sum_overlap, counts=value
             all_spheres = unique_rows(all_spheres)
-            counts = value
 
         # Mask coordinates beyond space
         idx = np.all(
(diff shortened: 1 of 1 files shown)


liuzhenqi77:
Tried to fix the `sum_overlap=True` case. Unittests passed for me locally.

nimare/meta/utils.py
```diff
@@ -130,19 +130,22 @@ def _convolve_sphere(kernel, peaks):
...
-        ma_values[~mask_data] = 0
+        sphere_idx_inside_mask = np.where(mask_data[tuple(all_spheres.T)])[0]
+        sphere_idx_filtered = all_spheres[sphere_idx_inside_mask, :].T
+        nonzero_idx = tuple(sphere_idx_filtered)
 
-        nonzero_idx = np.where(ma_values > 0)
+        if sum_overlap:
+            nonzero_to_append = counts[sphere_idx_inside_mask]
+        else:
+            nonzero_to_append = np.ones((len(sphere_idx_inside_mask),)) * counts
```
JulioAPeraza, line 144:
```suggestion
            nonzero_to_append = np.ones((len(sphere_idx_inside_mask),)) * value
```
With these changes, I think here we could use `value` and remove `counts = value` from line 125 so that counts will always be a list.

nimare/meta/utils.py
```diff
@@ -130,19 +130,22 @@ def _convolve_sphere(kernel, peaks):
         )
 
         all_spheres = all_spheres[idx, :]
+
         if sum_overlap:
             counts = counts[idx]
```
JulioAPeraza, line 135: I think this can be merged with (or removed in favor of) the `if sum_overlap:` you have below.

JulioAPeraza (approved):
Thank you for your contribution! These changes look good to me. I just have two small suggestions. I was able to reproduce the profiling results on different datasets.

> I also found that parallelize this loop with joblib could speedup the process by another small percent.

That's great. We are currently profiling `MKDAChi2()` to see how we better handle the memory while computing multiple ma_maps in parallel.
Thanks!

PR #749, opened 2022-09-24
fix a bug in conversion from z statistics to p values

### LINKED ISSUE #748: Fix a bug in conversion from z statistics to p values
## Summary

In nimare.transform.z-to-p function which converts z statistics to p-values, I found it should be ```p=stats.norm.sf(z) ```
instead of ```p = stats.norm.sf(abs(z))``` for one-tailed test(line 654). Otherwise, in hypothesis testing with null hypothesis is $H_0: \mu=\mu_0$ and alternative hypothesis $H_1: \mu>\mu_0$, if $\hat{\mu}\ll\mu_0$, $z=(\hat{\mu}-\mu_0)/se(\hat{\mu})$ is highly negative, but after taking the absolute value, the p-value becomes significant @tsalo @jdkent 

## Additional details
Negative z statistics values are not covered in testing for one-tailed test. For example,  (-1.959963, "one", 0.975) should be included in the testing, but this function gives a wrong output (-1.959963, "one", 0.025)
```
@pytest.mark.parametrize(
    "z,tail,expected_p",
    [
        (0.0, "two", 1.0),
        (0.0, "one", 0.5),
        (1.959963, "two", 0.05),
        (1.959963, "one", 0.025),
        (-1.959963, "two", 0.05),
        ([0.0, 1.959963, -1.959963], "two", [1.0, 0.05, 0.05]),
    ],
)
def test_z_to_p(z, tail, expected_p):
    """Test z to p conversion."""
    p = transforms.z_to_p(z, tail)

    assert np.all(np.isclose(p, expected_p))
```
It also suggests that p-values of voxel in deactivation area are believed to be significant (in one-tailed test) with the existence of this bug. In fact, when I did comparison of p-value maps between ALE and CBMR algorithm on real datasets, the dice similarity coefficient was approx 99% with two-tailed test, but dropped to approx 95% with one-tailed test, which might suggest some activated voxels (roughly 4%?) detected by ALE (one-tail test) is actually deactivation.

yifan0330:
Closes #748.

- one-sided right-tail test should take absolute value of z statistics.


jdkent (changes requested):
Thanks @yifan0330! could you add a test to the `test_z_to_p` function demonstrating the fix?
add something like `(-1.959963, "one", 0.975)`, 

https://github.com/neurostuff/NiMARE/blob/13d7b196e95f481f2c6b9cda7411ced942075935/nimare/tests/test_transforms.py#L249-L264

yifan0330:
> Thanks @yifan0330! could you add a test to the `test_z_to_p` function demonstrating the fix? add something like `(-1.959963, "one", 0.975)`,
> 
> https://github.com/neurostuff/NiMARE/blob/13d7b196e95f481f2c6b9cda7411ced942075935/nimare/tests/test_transforms.py#L249-L264

Sorry I just saw your comments! Thanks for the suggestion @jdkent, I've added this test

jdkent (approved):
LGTM, thanks for the fix!

PR #750, opened 2022-10-04
Fix MACM analysis example

JulioAPeraza:
Closes None.

- I think `atlas["maps"]` is already a `nibabel.nifti1.Nifti1Image` object. Using `nib.load` on `atlas["maps"]` will raise the error:
```
TypeError: stat: path should be string, bytes, os.PathLike or integer, not Nifti1Image
```


jdkent (approved):
Looks correct, checked the oldest version of nilearn that NiMARE supports, and it looks like it should be a nifti image and not a filename:
https://github.com/nilearn/nilearn/blob/ad04ee8ce91fde193fe4627f666ddeb52271a383/nilearn/datasets/atlas.py#L271-L429

PR #752, opened 2022-11-11
Remove "dataset" `return_type` option from kernel transformers

### LINKED ISSUE #630: Using pre-made MA maps in a meta-analysis is more memory-intensive and slower than making them in `fit`
## Summary

We have an argument, `return_type`, in `KernelTransformer.transform` to determine how the modeled activation (MA) maps should be returned. The options are `array`, `image`, and `dataset`. With `dataset`, the transformer creates each MA map and saves it to a file, and then essentially treats the MA map as an "image" to be loaded and masked as needed, instead of creating the MA maps from scratch.

I have an ongoing job that looks at duration and maximum memory usage of the MKDAKernel transformation step and the MKDADensity fit step with different configurations and Dataset sizes, drawing from Neurosynth.

As can be seen in the bottom subplot of the figure below, the red and brown lines, corresponding to MKDADensity models run on a Dataset with pre-generated MA maps, are using _much_ more memory than the green and purple lines, which are running the MKDAKernel to generate MA maps _within_ the meta-analysis, rather than beforehand.

Essentially, I think that loading and masking images takes more time and uses more memory than creating those images from scratch, which seems problematic for large-scale meta-analyses. I will update this as the job increases the Dataset sizes more.

Also, relating to #629, the 500MB memory limit (i.e., `MKDADensity(memory_limit="500mb")`) only has an effect on the meta-analyses using the pre-generated MA maps, and only then when max memory usage is around 13GB (mostly definitely _not_ 500MB).

profiling_results_v0_0_11

## Additional details

- NiMARE version: 0.0.11

Here's a gist with the script I'm using: https://gist.github.com/tsalo/2f3e28b1608c53e9f171d008f545b447

JulioAPeraza:
Closes #630.

- Remove "dataset" `return_type` option from kernel transformers.
- I'm keeping `KernelTransformer._infer_names()`, just in case someone wants to generate image names and saves them in the `Dataset.images` dataframe for `CorrelationDistributionDecoder`, as shown in the test `test_CorrelationDistributionDecoder_smoke()`.

### DIFF
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -207,7 +207,6 @@ class CBMAEstimator(Estimator):
             Return a 4D sparse array of shape
             (n_studies, mask.shape) with MA maps.
         """
-        # if maps_key in self.inputs_.keys():
         LGR.debug(f"Generating MA maps from coordinates ({coords_key}).")
 
         ma_maps = self.kernel_transformer.transform(
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -1,7 +1,6 @@
 """CBMA methods from the ALE and MKDA families."""
 import logging
 from abc import abstractmethod
-from hashlib import md5
 
 import nibabel as nib
 import numpy as np
@@ -92,8 +91,7 @@ class CBMAEstimator(Estimator):
         ----------
         dataset : :obj:`~nimare.dataset.Dataset`
             In this method, the Dataset is used to (1) select the appropriate mask image,
-            (2) identify any pre-generated MA maps stored in its images attribute,
-            and (3) extract sample size metadata and place it into the coordinates input.
+            and (2) extract sample size metadata and place it into the coordinates input.
 
         Attributes
         ----------
@@ -111,17 +109,6 @@ class CBMAEstimator(Estimator):
 
         for name, (type_, _) in self._required_inputs.items():
             if type_ == "coordinates":
-                # Try to load existing MA maps
-                if hasattr(self, "kernel_transformer"):
-                    self.kernel_transformer._infer_names(affine=md5(mask_img.affine).hexdigest())
-                    if self.kernel_transformer.image_type in dataset.images.columns:
-                        files = dataset.get_images(
-                            ids=self.inputs_["id"],
-                            imtype=self.kernel_transformer.image_type,
-                        )
-                        if all(f is not None for f in files):
-                            self.inputs_["ma_maps"] = files
-
                 # Calculate IJK matrix indices for target mask
                 # Mask space is assumed to be the same as the Dataset's space
                 # These indices are used directly by any KernelTransformer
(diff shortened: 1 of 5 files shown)


nimare/meta/cbma/base.py
```diff
@@ -220,37 +207,14 @@ def _collect_ma_maps(self, coords_key="coordinates", maps_key="ma_maps"):
...
-            n_studies = len(self.inputs_[maps_key])
-            shape = img_data.shape
-            kernel_shape = (n_studies,) + shape
-
-            exp = np.hstack(all_exp)
-            coords = np.vstack((exp.flatten(), np.hstack(all_coords)))
-            data = np.hstack(all_data).flatten()
-
-            ma_maps = sparse.COO(coords, data, shape=kernel_shape)
+        # if maps_key in self.inputs_.keys():
```
tsalo, line 210: Can you remove this commented-out code or, if it's meant as a comment and not commented-out code, convert it to a comment?

nimare/meta/kernel.py
```diff
@@ -7,24 +7,25 @@
...
 
 LGR = logging.getLogger(__name__)
 
 
 class KernelTransformer(NiMAREBase):
     """Base class for modeled activation-generating methods in :mod:`~nimare.meta.kernel`.
 
+    .. versionchanged:: 0.0.13
+
+            - Remove "dataset" `return_type` option.
```
tsalo, line 27: Can you add this to the child classes' docstrings as well?

nimare/meta/kernel.py
```diff
@@ -82,8 +83,8 @@ def transform(self, dataset, masker=None, return_type="image"):
             If None (and ``dataset`` is a Dataset), the Dataset's masker attribute will be used.
             Default is None.
         return_type : {'sparse', 'array', 'image', 'dataset'}, optional
```
tsalo, line 85:
```suggestion
        return_type : {'sparse', 'array', 'image'}, optional
```

tsalo (changes requested):
This definitely simplifies the code. Thanks! I've only requested a few minor changes.

JulioAPeraza:
It looks like python 3.6 is no longer available in ubuntu-latest. 
@tsalo, @adelavega, @jdkent. Should we maintain supporting python 3.6?

tsalo:
I would have preferred to have a sunsetting timeline in place ahead of time, with deprecation warnings to users and the like for a few releases, but if we can't test on 3.6 then I think we should drop it. Probably worth jumping to 0.1.0 in that case though.

JulioAPeraza:
I think we can test 3.6 with ubuntu-20.04. What about adding the deprecation warnings for 0.0.13, and then dropping 3.6 for 0.1.0

tsalo:
That would be nice.

adelavega:
Python 3.6 is end-of-life already and 3.7 will be very soon (so I would add that to the deprecation cycle). 

This is true for many tools also, so I think users on 3.6 will have to upgrade to get the latest versions of most of the related scientific software regardless. 

But I'm not opposed to your plan if you want to go ahead with it!

jdkent (approved):
LGTM!

PR #761, opened 2023-01-23
Add CBMA workflow

### LINKED ISSUE #760: Run an arbitrary CBMA workflow
## Summary

create a workflow that can run an arbitrary CBMA estimator, with corrector and diagnostics.

JulioAPeraza:
Closes #760.

- Add CBMA workflow: estimator, with corrector and diagnostics.
- Fix a typo in `MetaResult.save_tables()`.

### DIFF
--- a/nimare/workflows/cbma.py
+++ b/nimare/workflows/cbma.py
@@ -1,0 +1,84 @@
+"""Workflow for running an coordinates-based meta-analysis from a NiMARE database."""
+import logging
+
+from nilearn import reporting
+
+from nimare.correct import FWECorrector
+from nimare.diagnostics import FocusCounter
+from nimare.meta import ALE
+
+LGR = logging.getLogger(__name__)
+
+
+def cbma_workflow(
+    dataset,
+    meta_estimator=ALE(),
+    corrector=FWECorrector(),
+    diagnostics=(FocusCounter(),),
+    output_dir=None,
+):
+    """Compose a coordinate-based meta-analysis workflow.
+
+    .. versionadded:: 0.0.14
+
+    This workflow performs a coordinate-based meta-analysis, multiple comparison correction,
+    and diagnosis analysis on corrected meta-analytic images.
+
+    Parameters
+    ----------
+    dataset : :obj:`~nimare.dataset.Dataset`
+        Dataset for which to run meta-analyses to generate maps.
+    meta_estimator : :class:`~nimare.base.CBMAEstimator`, optional
+        Meta-analysis estimator. Default is :class:`~nimare.meta.cbma.ale.ALE`.
+    meta_corrector : :class:`~nimare.correct.Corrector`, optional
+        Meta-analysis corrector. Default is :class:`~nimare.correct.FWECorrector`.
+    diagnostics : :obj:`tuple`, optional
+        Tuple of meta-analysis diagnostic classes.
+        Default is :class:`~nimare.diagnostics.FocusCounter`.
+    output_dir : :obj:`str`, optional
+        Output directory in which to save results. If the directory doesn't
+        exist, it will be created. Default is None (the results are not saved).
+
+    Returns
+    -------
+    :obj:`~nimare.results.MetaResult`
+        Results of Estimator and Corrector fitting with cluster and diagnostic tables.
+    """
+    LGR.info("Performing meta-analysis...")
+    results = meta_estimator.fit(dataset)
+
+    LGR.info("Performing correction on meta-analysis...")
+    corr_results = corrector.transform(results)
+
+    for img_key in corr_results.maps.keys():
+        # Save cluster table only for z maps
+        if img_key.startswith("z_"):
+            cbma_img = corr_results.get_map(img_key)
+
+            LGR.info("Generating cluster tables...")
+            cluster_df = reporting.get_clusters_table(cbma_img, stat_threshold=0)
+            # Remove prefix from name
+            img_name = "_".join(img_key.split("_")[1:])
+
+            if not cluster_df.empty:
+                clust_tbl_name = f"cluster_{img_name}"
+                corr_results.tables[clust_tbl_name] = cluster_df
+
+            # Run diagnostic on corrected z maps
+            if "_corr-" in img_key:
+                for diagnostic in diagnostics:
+                    diagnostic.target_image = img_key
+
+                    LGR.info("Performing diagnostic on corrected meta-analysis...")
+                    count_df, _ = diagnostic.transform(corr_results)
+
+                    if not count_df.empty:
+                        diag_tbl_name = f"{diagnostic.__class__.__name__}_{img_name}"
+                        corr_results.tables[diag_tbl_name] = count_df
+
+    if output_dir is not None:
+        LGR.info(f"Saving meta-result object to {output_dir}...")
+        corr_results.save_maps(output_dir=output_dir)
+        corr_results.save_tables(output_dir=output_dir)
+
+    return corr_results
--- a/nimare/results.py
+++ b/nimare/results.py
@@ -148,7 +148,7 @@ class MetaResult(object):
         tables = {k: self.tables[k] for k in names}
 
         for tabletype, table in tables.items():
-            filename = prefix + tables + ".tsv"
+            filename = prefix + tabletype + ".tsv"
             outpath = os.path.join(output_dir, filename)
             table.to_csv(outpath, sep="\t", index=False)
 
--- a/nimare/tests/test_workflows.py
+++ b/nimare/tests/test_workflows.py
@@ -1,7 +1,11 @@
 """Test nimare.workflows."""
 import os.path as op
 
+import nimare
 from nimare import cli, workflows
+from nimare.correct import FWECorrector
+from nimare.diagnostics import FocusCounter
+from nimare.meta.cbma.ale import ALE
 from nimare.tests.utils import get_test_data_path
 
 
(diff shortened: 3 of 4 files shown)


nimare/workflows/cbma.py
```diff
@@ -0,0 +1,84 @@
...
+
+    LGR.info("Performing correction on meta-analysis...")
+    corr_results = corrector.transform(results)
+
+    for img_key in corr_results.maps.keys():
+        # Save cluster table only for z maps
+        if img_key.startswith("z_"):
+            cbma_img = corr_results.get_map(img_key)
+
+            LGR.info("Generating cluster tables...")
```
jdkent, line 58: This can either be moved outside the "for loop" or you can include the image name in the message,
I think I prefer to say "Generating cluster tables and performing diagnostics on corrected meta-analyses" outside the "for loop"

nimare/workflows/cbma.py
```diff
@@ -0,0 +1,84 @@
...
+
+            # Run diagnostic on corrected z maps
+            if "_corr-" in img_key:
+                for diagnostic in diagnostics:
+                    diagnostic.target_image = img_key
+
+                    LGR.info("Performing diagnostic on corrected meta-analysis...")
+                    count_df, _ = diagnostic.transform(corr_results)
+
+                    if not count_df.empty:
```
jdkent, line 75: could use a LGR.warn statement here if the `count_df` is empty. Just so users have a record in case they expected a table and did not get it.

nimare/workflows/cbma.py
```diff
@@ -0,0 +1,91 @@
...
+
+    LGR.info("Performing correction on meta-analysis...")
+    corr_results = corrector.transform(results)
+
+    LGR.info("Generating cluster tables and performing diagnostics on corrected meta-analyses...")
+    for img_key in corr_results.maps.keys():
+        # Save cluster table only for z maps
+        if img_key.startswith("z_"):
+            cbma_img = corr_results.get_map(img_key)
+            cluster_df = reporting.get_clusters_table(cbma_img, stat_threshold=0)
```
JulioAPeraza, line 58: Should we set the cluster forming threshold to something like `3.1`?

tsalo: I think it depends on the Corrector settings and the map. For voxel-level corrected and uncorrected maps, I'd say no cluster-forming threshold should be applied. For the cluster-level corrected maps, then the threshold should match the one used for the Corrector.

JulioAPeraza: I'm not sure how to access the `ss_thresh` parameter from the Corrector directly. 
Wouldn't that threshold (`stat_threshold=ss_thresh`) produce the same cluster table as `stat_threshold=0` for the cluster-level corrected maps, given that those maps were already thresholded/corrected?

tsalo: You're totally right... in that case, I think I need to break it down by image:
- z (uncorrected): p < 0.001 might be good here
- cFWE corrected z: The threshold would remove the clusters, so p < 0.05 is a good default here.
- vFWE corrected z: Since it's corrected, we probably want p < 0.05, rather than p < 0.001

jdkent:
I think a notebook example could be helpful for this pull request, I started one here: https://github.com/jdkent/NiMARE/blob/cbma-workflow/examples/02_meta-analyses/10_plot_cbma_workflow.py

jdkent:
just got the boilerplate pull request merged in, so the items for this pull request are:
1. [ ] support string inputs ("ale", "mkda", "fwe", "fdr", "jackknife", etc.) for novice users so they do not need to import (include the options in the docstring of the class)
2. [ ] change the default corrector to FWE montecarlo
3. [ ] refactor diagnostics to have a base class and add a "return_clusters_table_only" flag (perhaps with another name) that only returns the cluster table to account for the case where the estimator is not compatible with the diagnostics.
4. [ ] look at the ALE workflow to see how that workflow incorporates the boilerplate and do the same here
5. [ ] create an example (in the gallery) demonstrating the usage of the workflow.

JulioAPeraza:
> change the default corrector to FWE montecarlo

One potential issue with setting the default corrector to FWE montecarlo is that the example with a single line `result = cbma_workflow(dset)` will run `10000` iterations which may impact the running time of the example gallery. WDYT?

jdkent:
lets bring this up in the meeting, perhaps FDR could be a better default.

JulioAPeraza:
I think this is almost ready for review. It will need #776 merged for the tests to pass.

nimare/workflows/cbma.py
JulioAPeraza: thanks! and for FDR and FWE bonferroni which default thresholds would you suggest?

jdkent: for bonferroni-corrected p-values, I would say 0.05 as well. 

For the fdr corrected p-values, I would use whatever alpha is set to in the class:
https://github.com/neurostuff/NiMARE/blob/ad09e4fe7fedaaf6a64949c4c14164d1a02a9cf8/nimare/correct.py#L340

nimare/workflows/cbma.py
```diff
@@ -0,0 +1,177 @@
...
+        if estimator is None
+        else _check_input(estimator, CBMAEstimator, estm_options, n_cores=n_cores)
+    )
+    corrector = (
+        FWECorrector(method="montecarlo", n_cores=n_cores)
+        if corrector is None
+        else _check_input(corrector, Corrector, corr_options, n_cores=n_cores)
+    )
+
+    pval = 0.05 if corrector.method == "montecarlo" else 0.001
```
jdkent, line 120: Let's have 0.05 and 0.001 be defined as global variables defined at the top of the file that tells what they mean.

0.05 could be `CORRECTED_P_VALUE_THRESHOLD`
0.001 could be `UNCORRECTED_P_VALUE_THRESHOLD`

nimare/workflows/cbma.py
```diff
@@ -0,0 +1,177 @@
...
+        clust_tbl_name = f"{img_key}_clust"
+        count_tbl_name = f"{img_key}_{diag_name}"
+        if not clusters_table.empty:
+            corr_results.tables[clust_tbl_name] = clusters_table
+            corr_results.tables[count_tbl_name] = contribution_table
+        else:
+            LGR.warning(
+                f"Key {count_tbl_name} and {clust_tbl_name} will not be stored in "
+                "MetaResult.tables dictionary."
+            )
```
jdkent, line 161: I feel like we had a discussion about this, where I agreed that not saving empty tables was good behavior.

But now I feel like I'm flip flopping on that decision. If the output is valid (i.e., the data and code are correct, it's just the data wasn't signifiicant) then I think I should have the cluster table and contribution table saved, even if they are empty.

nimare/workflows/cbma.py
```diff
@@ -0,0 +1,177 @@
...
+
+    if output_dir is not None:
+        LGR.info(f"Saving meta-analytic maps, tables and boilerplate to {output_dir}...")
+        corr_results.save_maps(output_dir=output_dir)
+        corr_results.save_tables(output_dir=output_dir)
+
+        boilerplate = corr_results.description_
```
jdkent, line 168: This comment will be for another pull request, but I think I would like a combined description of everything that happened (estimator/corrector/diagnostics, etc. to reside in the metaResult object)

jdkent (changes requested):
This is looking good!

Just a few minor comments from me.

nimare/workflows/cbma.py
JulioAPeraza: Sound good!
I'm not really sure how the alpha (`q`) is used here: 
https://github.com/neurostuff/PyMARE/blob/604bb4fc33ff74703f0408ccd2031d3802113702/pymare/stats.py#L178

nimare/workflows/cbma.py
JulioAPeraza: Looking at the documentation, I think the descriptions are appended. I see the combined boilerplate for the estimator/corrector.

nimare/workflows/cbma.py
JulioAPeraza: The `None` for the contribution table may rise an error when saving the `.tsv`. What do you think we should do in that case?

JulioAPeraza: Raise a warning when the table is None.

JulioAPeraza:
> for FDR/Bonferonni can we use nilean instead?

@jdkent @tsalo I was trying to implement this, but I'm a little confused about a couple of things.

1. If we adopt `threshold_stats_img` for `FDR` and FWE `Bonferroni`, should we also return thresholded maps from FWE `Montecarlo`? The Montecarlo method removes clusters that are not significant, but it doesn't actually threshold the maps. We can still see non-significant voxels in significant clusters.
2. Also, by using a `threshold_stats_img` for `FDR` and `Bonferroni` we won't be able to return a corrected p-values map as we currently support (since `threshold_stats_img` only work on z maps), and also the `correct_fdr_negcorr` method won't be supported.
3. Another difference is in the z values itself. With the current behavior, the Corrector return corrected z values, but with `threshold_stats_img` the z values will be thresholded but not corrected. I was wondering if this would be Ok.

An alternative approach would be to add an alpha parameter to the `cbma_workflow` to use in the diagnostics. That parameter will be used only for visualization and determining the clusters for the contribution/clusters table. Alpha will be independent of the cluster-forming threshold that is used in the FWE Montecarlo.
WDYT?

jdkent:
oof, I mean I still like the idea of off-loading this functionality to nilearn, but perhaps in the interim we do something similar to how we are handling cluster tables? Copy their code and modify it so it does not return thresholded maps?

I can see the argument from a data sharing perspective that we don't want to arbitrarily "zero-out" voxels. that's useful information generally.

I was viewing it through a result presentation perspective, where if you correct the results at a particular threshold, then that is the threshold that should be applied to the data.

jdkent:
but perhaps the nilearn implementation is too different for our use case.

jdkent:
everything looks good to me, great work @JulioAPeraza!

jdkent:
For now the nilearn method of thresholding is a different concept of how we are doing correction.

jdkent (approved):
LGTM!

PR #762, opened 2023-01-25
Support nibabel 5.0.0

### LINKED ISSUE #759: support nibabel 5.0.0+
## Summary

related to #758, after merging #757. we do not support nibabel 5.0.0 because of test failures.
## Additional details

supporting nibabel 5.0.0
## Next steps

identify places where we pass an int array into a nifti object

> Passing an int64 array to Nifti1Image without a header or dtype argument will raise a ValueError

JulioAPeraza:
Closes #759.

- Add header to `Nifti1Image` when passing an int64 array.

### DIFF
--- a/examples/03_annotation/04_plot_gclda.py
+++ b/examples/03_annotation/04_plot_gclda.py
@@ -113,9 +113,11 @@ decoded_df.sort_values(by="Weight", ascending=False).head(10)
 
 ###############################################################################
 # First we'll make an ROI
+hdr = nib.Nifti1Header()  # Passing an int64 array to Nifti1Image requires a header
+hdr.set_data_dtype(np.int64)
 arr = np.zeros(dset.masker.mask_img.shape, int)
 arr[65:75, 50:60, 50:60] = 1
-mask_img = nib.Nifti1Image(arr, dset.masker.mask_img.affine)
+mask_img = nib.Nifti1Image(arr, dset.masker.mask_img.affine, header=hdr)
 plotting.plot_roi(mask_img, draw_cross=False)
 
 ###############################################################################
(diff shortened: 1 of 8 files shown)


examples/03_annotation/04_plot_gclda.py
```diff
@@ -113,9 +113,11 @@
 
 ###############################################################################
 # First we'll make an ROI
+hdr = nib.Nifti1Header()  # Passing an int64 array to Nifti1Image requires a header
+hdr.set_data_dtype(np.int64)
 arr = np.zeros(dset.masker.mask_img.shape, int)
```
tsalo, line 118: Alternative, you could use `np.int32` as the dtype here. Then you wouldn't need to create and pass in the header.

nimare/meta/kernel.py
```diff
@@ -169,8 +169,10 @@ def transform(self, dataset, masker=None, return_type="image"):
                 img = kernel_data[mask_data]
                 imgs.append(img)
             elif return_type == "image":
+                hdr = nib.Nifti1Header()  # Passing an int64 array to Nifti1Image requires a header
+                hdr.set_data_dtype(np.int64)
                 kernel_data *= mask_data
-                img = nib.Nifti1Image(kernel_data, mask.affine)
+                img = nib.Nifti1Image(kernel_data, mask.affine, header=hdr)
```
tsalo, line 175: Since this is the general KernelTransformer, and there are a kernels that won't produce int data (e.g., the ALE kernel), I think it might be better to coerce the data in associated kernel transformers (KDA, MKDA) to int32.

tsalo (commented):
I think it might be better to change the numpy arrays' data type to int32 (when applicable) instead of creating int64 headers. I don't think AFNI plays well with int64 images anyway.

jdkent (approved):
LGTM

PR #763, opened 2023-01-26
[ENH] rudimentary support for nimads

jdkent:
taking over #628 to just implement reading the studysets without running workflows.
references #218 .

- this is really just a minimal representation to work with the upload process
- will add more functionality with more pull requests.


jdkent:
note to self: generate an example of how I want annotations to work.

examples/01_datasets/05_plot_nimads.py
```diff
@@ -0,0 +1,60 @@
...
+
+from nimare.io import convert_nimads_to_dataset
+from nimare.nimads import Studyset
+
+###############################################################################
+# Download Data from NeuroStore
+# -----------------------------------------------------------------------------
+
+
+def download_file(url):
```
adelavega, line 22: Minor comment: I'm not sure this function is that useful since you only use it twice (below). 

Maybe you could make this more useful and use it throughout

examples/01_datasets/05_plot_nimads.py
```diff
@@ -0,0 +1,60 @@
...
+# Then you can run NiMARE analyses on the Dataset object.
+
+nimare_dset = studyset.to_dataset()
+nimare_dset.coordinates.head()
+
+###############################################################################
+# Directly to NiMARE Dataset
+# -----------------------------------------------------------------------------
+# Alternatively, you can convert the NiMADS json files directly to a NiMARE Dataset object
+# if you wish to skip using the nimads studyset object directly.
```
adelavega, line 57: When does it make sense to use the Studyset object directly?

nimare/nimads.py
```diff
@@ -0,0 +1,415 @@
...
+        """Convert the Image to a dictionary."""
+        return {
+            "url": self.url,
+            "filename": self.filename,
+            "space": self.space,
+            "value_type": self.value_type,
+        }
+
+
+class Point:
```
adelavega, line 393: My only concern here is that this is sort of "hard coded". Should be fine for now but ideally this structure could be autogenerated from somewhere else?

I wonder if there's any way to auto generate an object given a JSON Schema

adelavega (approved):
Looks good! I was mostly thinking out loud in my comments, with respect to minimizing maintenance burdens in the future, but this looks functional.

nimare/nimads.py
jdkent: yep: https://docs.pydantic.dev/datamodel_code_generator/.

this can be an issue for a refactor.

examples/01_datasets/05_plot_nimads.py
jdkent: yeah, my threshold for wanting to write a function was pretty low, it's possible other examples could be downloaded down the line using this function.

PR #765, opened 2023-02-01
Support clusters table in Diagnostics

JulioAPeraza:
Closes None. With this PR we intend to remove `reporting.get_clusters_table` from the CBMA workflow (#761). 

- `Jackknife` and `FocusCounter` now return the clusters table in addition to the contribution table.

For a future PR it would be nice to refactor both Jackknife and FocusCounter to reduce some of the duplicated code.


tsalo:
Since this changes the numbers of outputs for the two methods, I'd label this PR as a breaking change. Alternatively, you could transpose the diagnostic tables and just tack on the clusters table as additional columns.

nimare/diagnostics.py
```diff
@@ -139,7 +150,6 @@ def transform(self, result):
         # Use study IDs in inputs_ instead of dataset, because we don't want to try fitting the
         # estimator to a study that might have been filtered out by the estimator's criteria.
         meta_ids = estimator.inputs_["id"]
-        rows = ["Center of Mass"] + list(meta_ids)
 
         # Let's label the clusters in the thresholded map so we can use it as a NiftiLabelsMasker
```
JulioAPeraza, line 154: Currently `get_clusters_table` only return the clusters table, so we need to re-run `ndimage.label` (reproducing the code in `get_clusters_table`). There is a current PR in nilearn to return the label map, but that won't be available until the next release.

jdkent: could you link this pull request in another issue so we can revisit this when nilearn makes another release?

jdkent (approved):
LGTM!

tsalo (changes requested):
I think this is really close. Can you just double-check that the `ndimage.generate_binary_structure` and `ndimage.label` calls are the same as nilearn? I think if they are, then the cluster labels will always match up between the clusters table and the contributions table.

JulioAPeraza:
Thanks, @tsalo! 
I think `ndimage.generate_binary_structure` runs with the same parameters:
nilearn:
```python
# Define array for 6-connectivity, aka NN1 or "faces"
bin_struct = generate_binary_structure(rank=3, connectivity=1)
```
nimare:
```python
conn = ndimage.generate_binary_structure(rank=3, connectivity=1)
```
The only difference is in `ndimage.label`, where nilearn labels positive and negative maps separately and runs on binarized arrays, whereas nimare doesn't make a distinction between positive and negative maps explicitly and runs on the thresholded but non-binarized arrays. 
nilearn:
```python
label_map = label(binarized, bin_struct)[0]
```
nimare:
```python
labeled_cluster_arr, n_clusters = ndimage.label(thresh_arr, conn)
```

examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.py
```diff
@@ -104,7 +104,7 @@
     target_image="z_desc-size_level-cluster_corr-FWE_method-montecarlo",
     voxel_thresh=None,
 )
-knowledge_count_table, _ = counter.transform(knowledge_corrected_results)
+knowledge_count_table, _, _ = counter.transform(knowledge_corrected_results)
```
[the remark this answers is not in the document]
JulioAPeraza: Should we show the clusters table for both the FocusCounter and Jackknife in this example?

jdkent:
My recommendation is to copy the code from the nilearn pull request and incorporate into nimare until the next nilearn release. Then we can copy how nilearn combines positive and negative clusters into a single table and replicate the order in the contribution tables.

JulioAPeraza:
Sounds good!
I was looking at [reporting.get_clusters_table](https://github.com/nilearn/nilearn/blob/main/nilearn/reporting/_get_clusters_table.py#L208) and noticed that for `two_sided=True`, it lists the clusters with positive values first and then the clusters with negative values. However, the cluster ID is reset, so we will find the first positive and negative clusters sharing the same "Cluster ID"=1.
In the contribution table, we use the cluster ID as the column identifier. Probably, we will need to use a new name for the columns. What do you think?

JulioAPeraza:
Another option would be to split the clusters table into two DFs: one for positive clusters and one for negative clusters. Then, the diagnostic will return three tuples:
```python
return contribution_tables, clusters_tables, labeled_cluster_imgs
```
where:
```python
contribution_tables = (pos_contribution_table, neg_contribution_table)
clusters_tables = (pos_clusters_table, neg_clusters_table)
labeled_cluster_imgs = (pos_labeled_cluster_img, neg_labeled_cluster_img)
```
We can check if there are negative values in `target_img` and run `reporting.get_clusters_table` with `two_sided=True`, or add `two_sided` to the list of parameters in `Jackknife` and `FocusCounter`.

JulioAPeraza:
@tsalo @jdkent
I was trying to use `get_clusters_tables` to get both the clusters table and label maps. We then use the label maps in `_transform` to get the contribution table.
However, I'm noticing a mismatch in the `Cluster ID` and the labels in the `label_map`. This is because `clust_ids` is [sorted](https://github.com/nilearn/nilearn/blob/main/nilearn/reporting/_get_clusters_table.py#L345) based on the `peak_vals`, but I don't think the label maps are relabeled too: https://github.com/nilearn/nilearn/blob/main/nilearn/reporting/_get_clusters_table.py#L340.

Do you think this is the desired behavior or it is a bug in `get_clusters_tables`?

tsalo:
I would guess it's a bug. Are you sure that https://github.com/nilearn/nilearn/pull/3477 didn't address the problem?

JulioAPeraza:
I do not think that PR solved the problem. The `label_maps` wasn't part of the output before, so there was no need to relabel the maps.

If that's the case, perhaps we can be solved the issue in our copy of `get_clusters_tables` by either changing the id (`c_id + 1`, with `c_val`) [here](https://github.com/nilearn/nilearn/blob/main/nilearn/reporting/_get_clusters_table.py#L375) and [here](https://github.com/nilearn/nilearn/blob/main/nilearn/reporting/_get_clusters_table.py#L386). Or by relabeling the label_map:

```python
re_label_map = np.zeros_like(label_map)
for c_id, c_val in enumerate(clust_ids):
    cluster_mask = label_map == c_val
    re_label_map[cluster_mask] = c_id + 1
```
What do you think?

JulioAPeraza:
@tsalo @jdkent, I think this is ready for review. thanks!

nimare/diagnostics.py
```diff
@@ -139,61 +138,59 @@ def transform(self, result):
...
+            contribution_table = pd.DataFrame(columns=["id"], data=rows)
+            return contribution_table, clusters_table, label_maps
+
+        contribution_tables = []
+        signs = [1, -1] if len(label_maps) == 2 else [1]
+        for sign, label_map in zip(signs, label_maps):
+            cluster_ids = sorted(list(np.unique(label_map.get_fdata())[1:]))
+
+            # Create contribution table
+            col_name = "PosTail" if sign == 1 else "NegTail"
```
jdkent, line 161: I don't have a great suggestion for what the name should be, maybe it should spell out `PositiveTail`, because it was not immediately clear in the example notebook. 

Regardless, If the fact "PosTail represents positive clusters and NegTail represents negative clusters" is added into the documentation of contribution_table then I would feel good about using whatever names you choose.

nimare/utils.py
```diff
@@ -1028,6 +1039,161 @@ def unique_rows(ar, return_counts=False):
         return ar[unique_row_indices]
 
 
+def _local_max(data, affine, min_distance):
```
jdkent, line 1042: I like to include the permalink of the code I copied:
https://github.com/nilearn/nilearn/blob/8b16d36bbf0f1b88b9ccaea2da1f2867024e16d5/nilearn/reporting/_get_clusters_table.py#L26-L116

or in your case it may be difficult since you have an open pull request to nilearn so the permalink from nilearn does not represent the fix you made for the labels map, so disregard this comment when appropriate

jdkent (approved):
LGTM, after our meeting, I think we've settled on decent names/structure of the tables with documentation.

PR #768, opened 2023-02-02
Link to NeuroStars software support category instead of neuro questions

tsalo:
Closes None.

- Link to NeuroStars software support category instead of the general category (which defaults new posts to "Neuro Questions").
- Add badges to link to the GitHub repo, so it's easier to go from the RTD site to the code.
- Fix Nilearn intersphinx link.


jdkent (approved):
LGTM, questions related to NiMARE are more likely to follow under software support than neuro questions, if we are being comprehensive, there are some questions that fall under using NiMARE syntactically correctly, but doing something conceptually wrong, but in those cases it's usually both syntax and concepts meaning it still fits under software support.

PR #769, opened 2023-02-04
Revert "Do not zero out one-tailed z-statistics for p-values > 0.5"

JulioAPeraza:
Reverts neurostuff/NiMARE#693


tsalo (approved):
Thanks for fixing this.

Issue #772, opened 2023-02-14
Refactor diagnostics module


JulioAPeraza:
## Summary

Following the changes in #765, where we adopted `get_clusters_table` from nilearn, we plan to get the label maps from that function as well. That output should be available in the next nilearn release. 

## Additional details

This will potentially reduce the computation time and the number of lines of code, as `ndimage.label` won't be re-run. 

## Next steps

For `nilearn>0.10.0`, add `return_label_maps` parameter:
```python
clusters_table, labeled_cluster_arr = reporting.get_clusters_table(
    target_img, 
    stat_threshold, 
    return_label_maps=True,
)
```

PR #776, opened 2023-03-15
Major refactoring of Diagnostics module

JulioAPeraza:
This is part of the refactoring issue #772, to reduce some of the duplicated code as was pointed out here #765.

- Create a new base class `Diagnostics`, to reduce duplicated code.  
- In addition, we plan to change the AttributeError to a Warning when using pairwise Estimators and return only the clusters table as suggested here #761 by @jdkent.


JulioAPeraza:
@jdkent To avoid the breaking change, do you think it would be a good idea if instead, we return an empty contribution table when the estimator is not supported, similar to what we do when the number of clusters is 0:
https://github.com/neurostuff/NiMARE/blob/08cbfad56e010f872bb858fca43e54f8bdc43bbe/nimare/diagnostics.py#L167-L170
In that case, we won't need to have the extra flag `return_clusters_table_only`.

JulioAPeraza:
Thanks! 
I've added the suggestions and documented the new changes. I also added a new test to improve coverage.

jdkent (approved):
Looked it over, changes look good and you test it with one of the pairwise estimators, LGTM.

JulioAPeraza:
thanks!

PR #777, opened 2023-03-23
Add `cluster_threshold` option to Diagnostics

JulioAPeraza:
Closes None.
This parameter will allow users to remove small clusters from the diagnostics.

- Add `cluster_threshold` option to Diagnostics
- Small refactoring.


jdkent (approved):
LGTM, I think adding a cluster_threshold argument makes the most sense for right now.

PR #780, opened 2023-03-31
Drop support for Python 3.6 and 3.7

### LINKED ISSUE #779: drop python 3.6, 3.7 from CI, and add python 3.11
## Summary

Keep up support of python versions that are supported in the python release cycle: https://devguide.python.org/versions/
Make a note in the documentation that this is the philosophy that we are adopting

## Additional details

keeping maintenance costs of the package to a reasonable level.

JulioAPeraza:
Closes #779.

- Drop Python 3.6 and 3.7.
- Support Python 3.11.

### DIFF
--- a/nimare/__init__.py
+++ b/nimare/__init__.py
@@ -1,6 +1,5 @@
 """NiMARE: Neuroimaging Meta-Analysis Research Environment."""
 import logging
-import sys
 import warnings
 
 from ._version import get_versions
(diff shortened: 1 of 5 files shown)


JulioAPeraza:
@jdkent It looks like some tools (e.g., `numba`) don't support Python 3.11 yet.
```
RuntimeError: Cannot install on Python version 3.11.2; only versions >=3.7,<3.11 are supported.
```

nimare/__init__.py
```diff
@@ -41,34 +40,3 @@
     ]
 
 del get_versions
-
-
-def _py367_deprecation_warning():
```
jdkent, line 46: This section of code still seems useful, but maybe we add back in later...

otherwise, I could see us tracking the minimum python version that we support and when python is ~3 months away from dropping support from that particular version, this code would become active again. This would also require changes to the developer notes for release.

This could be another issue/PR that is done later.

jdkent (approved):
LGTM, this solves the immediate issue, but we can brainstorm ideas on how we will deprecate and shift support to python releases in a semi-automatic way in the future.

PR #785, opened 2023-04-11
[DOC] initial add of jupyterlite

### LINKED ISSUE #784: add jupyterlite links to the examples gallery
## Summary

Add a Jupyterlite button to the generated examples for users

## Additional details

users could interact with the documentation code

## Next steps

https://sphinx-gallery.github.io/stable/configuration.html#jupyterlite

jdkent:
Closes pyodide/pyodide#784 .

- add jupyterlite buttons to every notebook
-


jdkent:
need to think about how to add dependencies to the python package:
https://jupyterlite.readthedocs.io/en/latest/howto/configure/kernels.html

JulioAPeraza:
I was trying to create an environment to load the pyodide kernel in Jupytelite, but we ran out of memory in RTD.

JulioAPeraza:
We also rely on nilearn for running the examples, which is not part of the pyodide kernel. Installing nilearn and nimare for running the notebook will take a lot of time.

jdkent:
I'm curious if we can build our own pyodide distribution in a github action and load the built/packaged pyodide in readthedocs?

JulioAPeraza:
`https://github.com/pyodide/pyodide/discussions/2719`

JulioAPeraza:
Installing the env using mamba seems to solve the memory issue in RTD but it looks like `cognitiveatlas` and `pymare=0.0.4rc2` are not part of the conda-forge channel.

JulioAPeraza:
Jupyterlite is now properly loading the pyodide kernel.

JulioAPeraza:
I have been trying to install a nimare wheel in pyodide, but some of the dependencies don't have a wheel available. For the nimare wheel available in Pypi I got this error:

```
ValueError: Can't find a pure Python 3 wheel for: 'numba', 'numpy<1.24,>=1.18', 'cognitiveatlas', 'indexed-gzip>=1.4.0', 'numba>=0.49'
```

And for the wheel created by this PR using a github action, I got this error:

```
ERROR: Could not find a version that satisfies the requirement cognitiveatlas (from nimare) (from versions: none)
ERROR: No matching distribution found for cognitiveatlas
```

The dependency with `indexed-gzip` was already addressed in a previous PR, and I think we can bump the versions for `numba` and `numpy`; but I'm not sure what to do about `cognitiveatlas`.

JulioAPeraza:
We will also need to find a server to host the pyodide wheel because they cannot be uploaded to PyPi: https://pyodide.org/en/stable/development/building-and-testing-packages.html#notes

JulioAPeraza:
Alternatively, we can leave the `build-pyodide` action just for testing nimare in the pyodide kernel, and make sure the next release of nimare is a pure Python package and thus installable via `micropip.install("nimare")` in Jupyterlite.

jdkent:
I am in support of trying to make nimare pure python if the only non-pure dependency is `indexed_gzip`, then that sounds like a very reasonable change.

JulioAPeraza:
Sounds good. 
We will also need to add Python 3.11 support before the next release. That will make nimare usable in the pyodide kernel, given that their latest version (0.23.1) run on Python 3.11.3. 
Python 3.11 support is pending the numba 0.57.0 release.

JulioAPeraza:
At least now we don't get the error with cognitiveatlas. I will be able to test the build again once https://github.com/neurostuff/NiMARE/pull/796 gets merged.

jdkent:
I think the numba issue is going to be non-trivial
follow the conversation here:
pyodide/pyodide#621
numba/numba#3284
numba/llvmlite#836

Unfortunately, I would consider this to be blocking since we depend on numba for sparse and sparse is essential for optimizing several core workflows for running meta-analyses.

jdkent:
I subscribed to the relevant issues to keep track on updates to this, but since this cannot be worked on for the forseeable future, I'm closing this issue.

anutosh491:
> I think the numba issue is going to be non-trivial

Hi, I just solved this problem. See my comment here : https://github.com/numba/numba/issues/3284#issuecomment-5071137306

Was just seeing who all might benefit from our work and I landed here @jdkent !

PR #790, opened 2023-04-20
Remove RC versions from Changelog

JulioAPeraza:
Closes None.

-
-

### DIFF
--- a/.github/workflows/update-changelog.yml
+++ b/.github/workflows/update-changelog.yml
@@ -2,7 +2,7 @@ name: "Update Changelog"
 
 on:
   release:
-    types: [published]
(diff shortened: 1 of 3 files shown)


.github/workflows/update-changelog.yml
```diff
@@ -2,7 +2,7 @@ name: "Update Changelog"
 
 on:
   release:
-    types: [published]
```
jdkent, line 5: here is relevant bit of documentation: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

makes sense.

jdkent (approved):
LGTM

PR #791, opened 2023-04-21
[MAINT] make indexed_gzip install optional

jdkent:
Closes # .

removing indexed_gzip makes NiMARE a pure python package (and therefore usable by pyodide/jupyterlite, etc)

- make indexed_gzip installable via `[gzip]`
-


tsalo:
If indexed_gzip isn't necessary, it would be great to remove it entirely. It might be good to see if the original issue that led to us requiring indexed_gzip (#399) has been resolved (i.e., hopefully nibabel has dealt with this issue already).

jdkent:
@tsalo do you know if there was a test written that covered the particular issue you referenced?

PR #792, opened 2023-04-29
handle case of no sample size being reported

jdkent:
Closes # .

- fix the conversion of studyset to dataset when sample size is not reported
-

### DIFF
--- a/nimare/nimads.py
+++ b/nimare/nimads.py
@@ -204,7 +204,7 @@ class Study:
         self.name = source["name"] or ""
         self.authors = source["authors"] or ""
         self.publication = source["publication"] or ""
-        self.metadata = source.get("metadata", {})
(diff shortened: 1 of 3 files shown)


nimare/nimads.py
```diff
@@ -204,7 +204,7 @@ def __init__(self, source):
         self.name = source["name"] or ""
         self.authors = source["authors"] or ""
         self.publication = source["publication"] or ""
-        self.metadata = source.get("metadata", {})
+        self.metadata = source.get("metadata", {}) or {}
```
tsalo, line 207: Shouldn't `source.get("metadata", {})` default to `{}` anyway?

tsalo: ohhh.... that makes sense. Thanks!

PR #797, opened 2023-05-03
[MAINT] fix readthedocs

### LINKED ISSUE #798: readthedocs build failing
looks like this is being fixed on the readthedocs side: readthedocs/readthedocs.org#10287, so not something we need to likely fix.

_Originally posted by @jdkent in https://github.com/neurostuff/NiMARE/issues/797#issuecomment-1533363534_

Keeping track of this as an issue instead of a pull request since we likely do not need to make any changes.

jdkent:
Closes #798 .

-
-

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,4 +16,3 @@ python:
       path: .
       extra_requirements:
         - doc
-  system_packages: true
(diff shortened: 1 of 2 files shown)


jdkent:
looks like this is being fixed on the readthedocs side: readthedocs/readthedocs.org#10287, so not something we need to likely fix.

PR #802, opened 2023-05-12
Add `reports` module

### LINKED ISSUE #773: Add Reports module
## Summary

A reports module could take diagnostics and information about the inputs/meta-analysis and create an html report

## Additional details

the report could be used for easily sharing results of a meta-analysis, would make writing a paper easier

## Next steps

- include either FocusCenter/Jackknife in report
- [SDM jove](https://www.jove.com/t/59841/meta-analysis-of-voxel-based-neuroimaging-studies-using-seed-based-d-mapping-with-permutation-of-subject-images-(sdm-psi))
- [MKDA toolbox](https://canlab.github.io/_pages/canlab_meta_analysis_walkthrough1.m/canlab_meta_analysis_walkthrough1.html)
- Number of foci per analysis
- Original space of coordinates (“UNKNOWN”)
- Check if foci are within the mask and report number of foci outside if any (Focusfilter)
- generate summaries of the estimator/corrector/diagnostics used.

JulioAPeraza:
Closes #773.

- Create HTML files with a summary and figures from a MetaResult object.


JulioAPeraza:
The report is now included in the documentation in the CBMA workflow example. It is still a work in progress.
https://nimare--802.org.readthedocs.build/en/802/auto_examples/02_meta-analyses/10_plot_cbma_workflow.html#report

JulioAPeraza:
It will be difficult to combine `view_connectome` and the heatmap in the same `report.html` file. Both plot use `plotly`, but `view_connectome` use a fixed [plotly.js](https://github.com/nilearn/nilearn/blob/70450b5ceb03157a249b0b4e36590ef5a674ccb9/nilearn/plotting/data/js/plotly-gl3d-latest.min.js) that is quite old. When I give priority to the latest version `view_connectome` doesn't render properly, and when I give priority to the old version `view_connectome` looks fine but the heatmap looks totally wrong (it showed a scatter plot instead of the heatmap).
One solution could be to embed the `view_connectome` plot to `report.html` using iframe. Or add a button to open the `view_connectome` in a separate window.

jdkent:
what does the actual html file look like? I'm curious if different versions of the javascript library can be loaded at different sections of the html file.

JulioAPeraza:
I think this is now ready for review.

jdkent (approved):
These changes look good to me! Great work, default.yml looks way more maintainable, we will tackle any incompatibilities when we come around to supporting some pairwise estimators and image based meta-analyses.

PR #803, opened 2023-05-16
Set `n_iters` and `n_cores` defaults only for estimators with `null_method="montecarlo"`

### LINKED ISSUE #801: Set `n_iters=10000` default only for `null_method="montecarlo"`
## Summary

It is a little confusing to set `n_iters=10000` as the default even when a  `null_method="approximate"` is used.

## Additional details

- NiMARE version: `main`

This is also related to the reports module #773 when I tried to show the parameters used to fit the estimator. 

Shows `n_iters=None` when `null_method="approximate"` is used.

### What actually happened?
Shows `n_iters=10000` when `null_method="approximate"` is used.

## Fixing the bug

```
self.n_iters = None if null_method == "approximate" else n_iters or 10000
```

JulioAPeraza:
Closes #801 .

- Set `n_iters` and `n_cores` defaults only for estimators with `null_method="montecarlo"`.

### DIFF
--- a/nimare/meta/cbma/ale.py
+++ b/nimare/meta/cbma/ale.py
@@ -125,8 +125,8 @@ class ALE(CBMAEstimator):
         self,
         kernel_transformer=ALEKernel,
         null_method="approximate",
-        n_iters=10000,
-        n_cores=1,
+        n_iters=None,
+        n_cores=None,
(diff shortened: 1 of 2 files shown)


jdkent (approved):
LGTM!

PR #804, opened 2023-05-17
Add Corrector and Diagnostics attributes to MetaResult object

### LINKED ISSUE #800: Add Corrector and Diagnostics as attributes to the MetaResults object
## Summary

While working on #773, I found that it would be nice to access the parameters that were used by the Estimator, Corrector, and Diagnostics and show them in the report.
If the report is built from a MetaResult object, in addition to the Estimator, I think we would also need the Corrector and Diagnostics as additional attributes. For example
```
MetaResult(estimator, corrector, diagnostics, mask, maps, tables, description)
```

## Additional details

Currently, the Diagnostics don't return a MetaResult object, but the outputs are tables and maps, which can be included in a MetaResult object.

## Next steps

Add Corrector and Diagnostics as attributes to the MetaResults object.

JulioAPeraza:
Closes #800 .

- Add Corrector and Diagnostics attributes to MetaResult object.
- Diagnostics.transform now returns a MetaResult object.
- Maps and tables from diagnostics are now added to `maps` and `tables` dictionaries.

### DIFF
--- a/nimare/tests/test_diagnostics.py
+++ b/nimare/tests/test_diagnostics.py
@@ -45,9 +45,11 @@ def test_jackknife_smoke(
     jackknife = diagnostics.Jackknife(target_image=target_image, voxel_thresh=1.65)
     results = jackknife.transform(res)
 
-    contribution_table = results.tables[f"{target_image}_diag-Jackknife_counts"]
-    clusters_table = results.tables[f"{target_image}_clust"]
-    label_maps = results.maps[f"{target_image}_PositiveTail"]
+    image_name = "_".join(target_image.split("_")[1:])
+    image_name = "_" + image_name if image_name else image_name
+    contribution_table = results.tables[f"{target_image}_diag-Jackknife_tab-counts"]
+    clusters_table = results.tables[f"{target_image}_tab-clust"]
+    label_maps = results.maps[f"label{image_name}_tail-positive"]
     if n_samples == "twosample":
         assert contribution_table is None
         assert not clusters_table.empty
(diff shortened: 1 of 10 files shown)


nimare/tests/test_diagnostics.py
```diff
@@ -60,8 +64,11 @@ def test_jackknife_with_zero_clusters(testdata_cbma_full):
     res = meta.fit(testdata_cbma_full)
 
     jackknife = diagnostics.Jackknife(target_image="z", voxel_thresh=10)
-    contribution_table, clusters_table, label_maps = jackknife.transform(res)
+    results = jackknife.transform(res)
 
+    contribution_table = results.tables["z_diag-Jackknife_counts"]
+    clusters_table = results.tables["z_clust"]
+    label_maps = results.maps["z_PositiveTail"]
```
jdkent, line 71: maybe a little unintuitive what PositiveTail means in this context, may need to make it more explicit that this is a label map and not a statistical map.

jdkent: and is the label map represented as a vector or a 3D array? I think the other maps are represented as 1D arrays.

JulioAPeraza: > maybe a little unintuitive what PositiveTail means in this context, may need to make it more explicit that this is a label map and not a statistical map.

Agreed. I've added a couple of new modalities to our bids-like format. See here: https://nimare--804.org.readthedocs.build/en/804/outputs.html#file-names

> and is the label map represented as a vector or a 3D array? I think the other maps are represented as vectors.

Good catch! I was missing the np.squeeze.

Thanks!

jdkent (approved):
LGTM, thanks @JulioAPeraza!

PR #807, opened 2023-05-23
Add NiMAREBase features to the Corrector base class

JulioAPeraza:
Closes None.

It would be nice to have some of the NiMAREBase features available in Corrector. For example, with `get_params()` we could get the parameters used to fit the corrector and thus include them in the report (#802). 

- Add NiMAREBase features to the Corrector base class


jdkent (approved):
As long as tests pass, this looks good to me.

PR #809, opened 2023-06-02
Add Workflow and CBMAWorkflow classes. Support pairwise CBMA workflows

JulioAPeraza:
Closes None.

- Convert cbma_workflow function into a class (CBMAWorkflow).
- Add Workflow base class.
- Support pairwise estimators in Diagnostics.
- Support pairwise estimators in Workflow.
- Support pairwise estimators in Report.


JulioAPeraza:
@jdkent We are getting an error while building the documentation. It seems related to the nimads source file:

```
WARNING: /home/docs/checkouts/readthedocs.org/user_builds/nimare/checkouts/809/examples/01_datasets/05_plot_nimads.py failed to execute correctly: Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/nimare/checkouts/809/examples/01_datasets/05_plot_nimads.py", line 28, in <module>
    nimads_studyset = download_file("https://neurostore.org/api/studysets/Cv2LLUqG76W9?nested=true")
  File "/home/docs/checkouts/readthedocs.org/user_builds/nimare/checkouts/809/examples/01_datasets/05_plot_nimads.py", line 25, in download_file
    return response.json()
  File "/home/docs/checkouts/readthedocs.org/user_builds/nimare/envs/809/lib/python3.8/site-packages/requests/models.py", line 975, in json
    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

jdkent:
Thanks for the heads up, I should have fixed the issue now.

nimare/diagnostics.py
```diff
@@ -288,15 +291,21 @@ def _transform(self, expid, label_map, result):
             target_value_map = "est"
         elif "stat" in result.maps:
             target_value_map = "stat"
+        elif "z_desc-consistency" in result.maps:
```
jdkent, line 294: I think the specificity map should be the preference (i.e., `z_desc-specificity`) since the result is what is compared to the base rate of activations across the reference dataset.

nimare/diagnostics.py
```diff
@@ -204,8 +202,14 @@ def transform(self, result):
 
             contribution_tables.append(contribution_table.reset_index())
 
-        # Concat PositiveTail and NegativeTail tables
-        contribution_table = pd.concat(contribution_tables, ignore_index=True, sort=False)
+        if pairwaise_estimators or len(label_maps) == 1:
```
jdkent, line 205: For MKDAChi2, the z-values can be positive or negative (you can check the z-map results from running the test in this pull request: https://github.com/neurostuff/NiMARE/pull/811).

I think we would want to display both Positive and Negative Tail cluster with Pairwise estimators.

nimare/workflows/__init__.py
```diff
@@ -1,7 +1,14 @@
...
 from .ale import ale_sleuth_workflow
-from .cbma import cbma_workflow
+from .base import Workflow
+from .cbma import CBMAWorkflow, PairwiseCBMAWorkflow
 from .macm import macm_workflow
 
-__all__ = ["ale_sleuth_workflow", "cbma_workflow", "macm_workflow"]
+__all__ = [
+    "ale_sleuth_workflow",
+    "Workflow",
```
jdkent, line 10: I would remove the  `Workflow` baseclass from `__all__`, this is not a public facing class (e.g., do not want users importing  and trying to use it by mistake.

jdkent (commented):
Looking good, just have a couple comments so far.

Could you add a report that showcases the output of a pairwise estimator?
you can use this existing example: [08_plot_cbma_subtraction_conjunction.py](https://github.com/neurostuff/NiMARE/blame/6c60e91c211ce7de1e7fa35d18f5e10ae1a72eb6/examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.py)

nimare/diagnostics.py
JulioAPeraza: Also in the report, we only show the summary for `id1` and `coordinates1` from dataset1. Should we include another section for the second group?

jdkent: This is a little more questionable, overall, I would say yes, we do want to display `id2`  for the coordinates. for the contribution table, this is a bit debatable. With ALESubtraction, yes, I think it is useful to know how both groups contributed to the existing clusters.

For the MKDAChi2 I think it is less useful to see a contribution table (and with jacknife it would be computationally prohibitive if one was using the entire neurosynth dataset).

But the notion that ALESubtraction is used for more similarly sized groups and MKDAChi2 is used for comparing a relatively smaller dataset to a huge dataset is only reflective of how we are currently the algorithms, there's nothing stopping you from using ALESubtraction with neurosynth.

So I think the answer would be creating an option/flag for the user to decide whether they want to show the second group or not. could be called `display_second_group` or something, with a default value of `False`.

nimare/diagnostics.py
```diff
@@ -288,15 +312,27 @@ def _transform(self, expid, label_map, result):
             target_value_map = "est"
         elif "stat" in result.maps:
             target_value_map = "stat"
+        elif "stat_desc-group1MinusGroup2" in result.maps:
+            target_value_map = "stat_desc-group1MinusGroup2"
+        elif "z_desc-specificity" in result.maps:
+            target_value_map = "z_desc-specificity"
         else:
             target_value_map = "z"
```
jdkent, line 320: this could be replaced by taking the union of two sets? one set being ("est", "stat", "stat_desc-group1MinusGroup2", "z_desc-specificity") and the other being the result.maps, then if the union of the sets is empty, have the target_value_map be "z".

Or taking a step back, the purpose of this is to find the uncorrected values, correct? (if the target_image was the corrected values, so one could theoretically manipulate the target_image name to create the target_value_map name, then it would just be a parameter instead of a list of if statements.

jdkent (changes requested):
Thanks for all your work on this!, I have a couple comments/questions for idea on how to restructure small sections of code.

nimare/diagnostics.py
```diff
@@ -288,15 +312,27 @@ def _transform(self, expid, label_map, result):
...
 
         # Fit Estimator to all studies except the target study
         other_ids = [id_ for id_ in all_ids if id_ != expid]
-        temp_dset = dset.slice(other_ids)
-        temp_result = estimator.fit(temp_dset)
+        if self._is_pairwaise_estimator:
+            temp_dset = (
+                estimator.dataset1.slice(other_ids)
+                if sign == "PositiveTail"
+                else estimator.dataset2.slice(other_ids)
```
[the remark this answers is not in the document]
JulioAPeraza: Good catch! I forgot to cover these cases.

jdkent (changes requested):
I think "PositiveTail" and "NegativeTail" should be defined as variables and described since they are written as strings a few too many times, so it could be hard to get the context of what the string means and more difficult to change in the future if we have to change every string instead of a variable.

JulioAPeraza:
Great point.

examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.py
```diff
@@ -129,7 +131,7 @@
 )
 related_corrected_results = jackknife.transform(related_corrected_results)
```
jdkent, line 132: wasn't from this pull request, but I think this variable should have a different name, like `related_diagnostic_results`, so that `related_corrected_results` is not overwritten. follows the convention of going from regular results to corrected results.

jdkent (changes requested):
sorry, just another minor comment, I think after that it looks good to me.

jdkent (approved):
LGTM! Thanks @JulioAPeraza

PR #815, opened 2023-06-13
Fix the aspect ratio and size of the heatmap in Reports

### LINKED ISSUE #814: plotly graph for reports has large text when report viewed locally
## Summary

if you view this generated report locally, the text for the diagnostic heatmap is too large
[report.zip](https://github.com/neurostuff/NiMARE/files/11730569/report.zip)

## Additional details

- NiMARE version: 0.1.1

### What actually happened?
Screenshot from 2023-06-12 23-24-07

## Reproducing the bug

@JulioAPeraza was this a problem you were having earlier when trying to load the different javascript library in an iframe? Or do you think this is more of a case of just needing to give more parameters to the heatmap for it to display the text smaller and the actual heatmap larger?

JulioAPeraza:
Closes #814.

- Use iframes for Connectomes instead, given that its aspect ratio is fixed. 
- Keep the size of the heatmap fixed by including the str in the HTML directly.
- Set the size of the heatmap proportional to the number of rows (number of studies) and columns (number of clusters).

### DIFF
--- a/nimare/reports/base.py
+++ b/nimare/reports/base.py
@@ -60,8 +60,9 @@ SVG_SNIPPET = [
 ]
 
 IFRAME_SNIPPET = """\
-<iframe id="igraph" scrolling="no" style="border:none;" seamless="seamless" \
-src="./{0}" height="800" width="100%"></iframe>
+<div class="igraph-container">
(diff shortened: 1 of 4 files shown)


jdkent (approved):
LGTM, thanks!

PR #817, opened 2023-06-23
Add IBMAWorkflow

JulioAPeraza:
Closes None.

- Support IBMA estimators in the Workflow module.
- Generate HTML report for IBMAWorkflow.


examples/02_meta-analyses/12_plot_ibma_workflow.py
```diff
@@ -0,0 +1,114 @@
...
+dset_file = os.path.join(get_resource_path(), "nidm_pain_dset.json")
+dset = Dataset(dset_file)
+dset.update_path(dset_dir)
+
+# Resample images
+imgs = dset.get({"z_maps": ("image", "z")})["z_maps"]
+mask_img = dset.masker.mask_img
+for img in imgs:
+    res_img = resample_to_img(nib.load(img), mask_img, clip=True, interpolation="linear")
+    res_img.to_filename(img)
```
jdkent, line 47: Is resampling not being handled here?

https://github.com/neurostuff/NiMARE/blob/e833c4bcb4a921ddea7b7f8c7c1aab28fec7dfa8/nimare/meta/ibma.py#L76-L88

jdkent (commented):
Looks good, just one question in the example notebook

examples/02_meta-analyses/12_plot_ibma_workflow.py
JulioAPeraza: I think resampling is performed if `resample=True` in the IBMA estimator. Since we are initializing the estimator with the default parameters, where `resample=False`, that step is skipped.

jdkent: I think we should initialize the estimator with resample=true, it means we can't use simple strings and have to import the estimator, but I think that's better than what we are currently doing.

JulioAPeraza:
It looks like an epsilon that small does not solve the issue. I think we would need a higher threshold so that the coverage maps are not 1 in all the voxels.

examples/02_meta-analyses/12_plot_ibma_workflow.py
```diff
@@ -0,0 +1,107 @@
...
+from nimare.dataset import Dataset
+from nimare.transforms import ImageTransformer
+from nimare.utils import get_resource_path
+
+dset_file = os.path.join(get_resource_path(), "nidm_pain_dset.json")
+dset = Dataset(dset_file)
+dset.update_path(dset_dir)
+
+# Calculate missing images
+xformer = ImageTransformer(target=["varcope", "z"])
```
jdkent, line 43: it would be nice if we tried to handle the transformations for the user based on the algorithm they selected.

jdkent (commented):
Looks good to me, would the workflow to try to handle transformations, and raise an error if there isn't the right data to do the transformations.

examples/02_meta-analyses/12_plot_ibma_workflow.py
JulioAPeraza: that's a great idea!

jdkent (approved):
LGTM! Thanks for all your work on this!

PR #820, opened 2023-07-20
A bug with CBMRInference when input moderators is None

yifan0330:
Closes # .

- We found a bug with CBMRInference when the parameter moderators is None
- It happens when calling the function `create_regular_expressions` for parsing contrast names


jdkent:
closed in favor of #821

PR #840, opened 2023-09-20
fix an index error in diagnostics.py

yifan0330:
Closes # .

- When checking if a foci coordinate falls within brain mask, index starts from 0 while coordinate starts from 1, so index should be [coord[0]-1, coord[1]-1, coord[2]-1].
-

### DIFF
--- a/nimare/diagnostics.py
+++ b/nimare/diagnostics.py
@@ -478,7 +478,8 @@ class FocusFilter(NiMAREBase):
         # Only retain coordinates inside the brain mask
         def check_coord(coord):
             try:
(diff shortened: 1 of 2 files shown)


jdkent:
This should be fixed by #839, the error wasn't an off-by-one, there are actually coordinates that are outside of the data array because of coordinate extraction errors from Neurosynth. The previous code didn't raise an error when coordinates were outside the data array.

yifan0330:
Thanks for your comments @jdkent ! Yes you're right, I thought coords voxel space start from 1 but it seems to start from 0

PR #845, opened 2023-11-07
Implement caching for Estimators and Transformers

### LINKED ISSUE #844: Add caching for nimare.meta algorithms
## Summary
Hi! Similar to #826 it would also be nice if the `nimare.meta` algorithms would also be cacheable :)

## Additional details

I for example use `nimare.meta.cbma.mkda.MKDAChi2()` which can take quite some time. While writing my analysis script this step always takes some time to finish which is a little bit tedious when trying to prototype an analysis script.

## Next steps
Use joblib just as in nilearn or scikit-learn?

JulioAPeraza:
Closes #844.

We are implementing caching at three levels of a meta-analysis: 

- High-level (or estimator-level): caches `_fit()`
- Mid-level (or transformer-level): caches `_transform()`
- Low-level (or modeled activation (ma)-level): caches `_get_ma_map()`

The most typical use case will be at lower levels, given that it is very common to recompute the same MA maps when working with the same database (e.g., Neurosynth). 

Transformer, at the mid-level, will benefit as well from caching.

At the estimator level, caching was implemented just for the sake of completeness. We currently recommend saving the MetaResult object to a pickle file and loading it again if will be reused. This reduces the large overhead that comes from hashing a whole NiMARE database object.


JulioAPeraza:
This shows the performance at memory level 1 (caching only `_fit()`). With memory level 2 (caching `_fit()` and `_transform()`) the performance was the same. However, at memory level 3 (caching `_fit()`, `_transform()`, and `_get_ma_map()`) I saw a minimal decrease in the performance but still better than no caching.

output

JulioAPeraza:
I tested the hierarchy inverted:

Memory level 1: caching `_get_ma_map()`.
Memory level 2: caching `_get_ma_map()`, `_transform()`.
Memory level 3: caching `_get_ma_map()`, `_transform()`, and `_fit()`.

However, the performance at level 1 with caching was worse than no caching, probably because the computation in `_get_ma_map()` is not that expensive.

jdkent (approved):
LGTM!

PR #848, opened 2023-12-01
Support liberal mask in IBMA estimators

JulioAPeraza:
Closes None. Related to https://github.com/neurostuff/PyMARE/issues/44

- Add _apply_liberal_mask to separate the data in bags of voxels in the same studies. Some parts of the proposed function require nested for loops, which was inefficient, so I have adapted some of the procedures to use numba with nopython=True.
- Currently, we loop over the bags and create a Pymare dataset. In the future, we can parallelize that loop to improve performance.


JulioAPeraza:
Some preliminary results using Stouffers:
image

tsalo:
That's so awesome! Do you have a DOF map among the outputs?

JulioAPeraza:
We don't have them it, but that will be a great addition.

JulioAPeraza:
Thanks @tsalo! Would that look like the figure below?

image

tsalo:
I can't access the linked figure, unfortunately.

jdkent (changes requested):
just a couple changes I requested, but otherwise LGTM, will think of possible refactor of some of the class specific bag handling to be placed in the base class.

jdkent (approved):
There is some potential repetitive code, but that can be refactored later, LGTM!

PR #852, opened 2023-12-13
Add similarity correlation matrix figure to IBMA reports

### LINKED ISSUE #847: Add similarity-sorted correlation matrix figure to IBMA reports
## Summary

One issue with images on NeuroVault is that the contrast is not always clear or correctly labeled.
I was thinking that something that could help diagnose inverted contrasts would be a similarity-sorted correlation matrix figure, like the one in the [NARPS paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7771346/) that helped identify which processing pipelines gave inverted results to the others? It might be useful for flagging images with the wrong contrast (see below).

I don't necessarily think this would need clusters.

It could alternatively be a diagnostic of some kind.

image

JulioAPeraza:
Closes #847.

- Adding a similarity-sorted correlation matrix figure to the summary section of the IBMA reports.

### DIFF
--- a/nimare/reports/base.py
+++ b/nimare/reports/base.py
@@ -321,7 +321,7 @@ def _gen_figures(results, img_key, diag_name, threshold, fig_dir):
 
         # Plot heatmaps
         [
-            plot_heatmap(contribution_table, fig_dir / heatmap_name, reorder="single", zmin=0)
+            plot_heatmap(contribution_table, fig_dir / heatmap_name, zmin=0)
             for heatmap_name, contribution_table in zip(heatmap_names, contribution_tables)
         ]
 
--- a/nimare/reports/base.py
+++ b/nimare/reports/base.py
@@ -27,6 +27,7 @@ from glob import glob
 from pathlib import Path
 
 import jinja2
+import pandas as pd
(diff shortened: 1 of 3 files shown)


JulioAPeraza:
Find an example here: https://nimare--852.org.readthedocs.build/en/852/auto_examples/02_meta-analyses/12_plot_ibma_workflow.html#report

tsalo:
The figure looks great, although I'm not sure about the map sorting. What sorting method do you use?

JulioAPeraza:
I'm using [scipy.cluster.hierarchy.linkage](https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html) with the method `average` for this matrix. I found that the diagnostic table looks better with `single`. Would you suggest a different method for the correlation matrix?

Here is a figure using `complete`:

image

JulioAPeraza:
Using `single`:

image

tsalo:
It looks like they're the same either way. "single" sounds good. Thanks for trying it both ways!

jdkent (approved):
LGTM!

PR #856, opened 2023-12-21
Disable computation of probabilities by default for MKDAChi2

### LINKED ISSUE #855: Disable MKDAChi2 posterior probability outputs by default
## Summary

This stems from a discussion on Slack. @nicholst and @adelavega are concerned that users might misuse posterior probability estimates produced by `MKDAChi2`, especially if they're relying on the default uniform prior (0.5).

## Next steps

1. Replace default `prior` value of 0.5 with `None`.
2. Disable posterior probability calculation if `prior` is `None`.
3. Add info about change to documentation.

adelavega:
Closes #855 

- Makes `prior` argument in `MKDAChi2` optional, and only computes posterior probabilities based on prior if given
- Renames outputs to match Neurosynth.org (consistency -> uniformity; specificity -> association)


adelavega:
Not sure why doc build is failing

tsalo:
Can you update your branch from `main`? Maybe that'll fix the docs build.

nimare/meta/cbma/mkda.py
```diff
@@ -535,9 +546,7 @@ def _fit(self, dataset1, dataset2):
             "p_desc-consistency": pAgF_p_vals,
             "p_desc-specificity": pFgA_p_vals,
         }
-        description = self._generate_description()
-
-        return maps, {}, description
+        return maps, {}
```
jdkent, line 549:
```suggestion
        description = self._generate_description()
        return maps, {}, description
```

jdkent: this should fix a number of those failing tests.

examples/misc-notebooks/nidm_pain_meta-analyses.ipynb
jdkent, line 1: Can't view this diff, did you purposely change something in this example notebook?

adelavega: yes, there was a reference to the ld naming style for chi2

nimare/meta/cbma/mkda.py
```diff
@@ -523,20 +528,22 @@ def _fit(self, dataset1, dataset2):
...
-            "z_desc-consistency": pAgF_z,
-            "z_desc-specificity": pFgA_z,
-            "chi2_desc-consistency": pAgF_chi2_vals,
-            "chi2_desc-specificity": pFgA_chi2_vals,
-            "p_desc-consistency": pAgF_p_vals,
-            "p_desc-specificity": pFgA_p_vals,
         }
-        description = self._generate_description()
 
+        if self.prior is not None:
```
jdkent, line 542:
```suggestion
        if self.prior:
```
with the current code, if `self.prior` is set to false it will still run the prior.
Taking anything that is truthy or false-y is also a bit weird, but a little more intuitive imo

examples/misc-notebooks/nidm_pain_meta-analyses.ipynb
adelavega: Restoring old notebook, as it needs updates that are out of the scope of this PR

nimare/meta/cbma/mkda.py
```diff
@@ -476,26 +481,26 @@ def _fit(self, dataset1, dataset2):
...
         pAgF = n_selected_active_voxels / n_selected
         pAgU = n_unselected_active_voxels / n_unselected
         pFgA = pAgF * pF / pA
 
         del pF
 
-        # Recompute conditionals with uniform prior
-        pAgF_prior = self.prior * pAgF + (1 - self.prior) * pAgU
-        pFgA_prior = pAgF * self.prior / pAgF_prior
+        if self.prior is not None:
```
jdkent, line 490:
```suggestion
        if self.prior:
```
oh I guess the if statement is there twice

PR #857, opened 2024-01-05
Optimize compute_kda_ma for memory and speed

adelavega:
- Implement "summary_array" return type for `MKDAKernel`, which convolves kernels to coordinates in a 3D dense volume, summing counts in place, saving substantial memory and compute. 
- Using numba to speed up sphere kernel convolution
- @jdkent set types to int to reduce memory usage
- Minor improvements throughout

For large-scale `MKDAChi2` (i.e. using Neurosynth dataset), memory footprint is reduced ~18-20x (25GB to 1.2GB), and computation is sped up ~3.2x.

### DIFF
--- a/nimare/meta/cbma/base.py
+++ b/nimare/meta/cbma/base.py
@@ -205,7 +205,7 @@ class CBMAEstimator(Estimator):
         """
         return None
 
-    def _collect_ma_maps(self, coords_key="coordinates", maps_key="ma_maps"):
+    def _collect_ma_maps(self, coords_key="coordinates", maps_key="ma_maps", return_type="sparse"):
         """Collect modeled activation maps from Estimator inputs.
 
         Parameters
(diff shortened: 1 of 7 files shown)


adelavega:
@jdkent these improvements make a small difference to run speed, but ultimately it's a pretty hard thing to optimize. 

ultimately, it's just slow and fairly expensive process

adelavega:
Even more speed increases by indexing the ijks array within numba

jdkent (approved):
LGTM! it's added and tested. @JulioAPeraza will be adding results for additional confirmation the results looks the same using the updated method.

adelavega:
Awesome. Let's merge once everything passes, and cut a release soon.

JulioAPeraza:
@adelavega, the new changes are fantastic! In the past, I could not train the decoder with more than 16 cores. 

I used 40 cores this time, and it took 11 minutes to train an LDA-based decoder with 200 topics, and it only required 1.7 GB of memory.

@jdkent, I compared the decoders' results, and the output maps' values are exactly the same.

I also tested the new changes with different dataset sizes based on Neurosynth. See below:

MKDAChi2_comparison

adelavega:
Julio, thank for your running this comparison. This is fantastic :)
This should make it much easier for people to replicate the results of your new paper.

PR #858, opened 2024-01-06
Kernalize all

adelavega:
Attempt to kernalize all at once in numpy

@jdkent can you test the memory consuption of this branch?

it might be more efficient because it convolves all coordiantes at once. no need to merge arrays.


adelavega:
It's also *slightly* faster. 51 seconds for pain vs neurosynth, vs 55 with the other branch.

jdkent:
ended up making memory consumption more and time complexity more

PR #859, opened 2024-01-09
Add `sum_across_studies` to kda

adelavega:
`sum_across_studies` returns an array where study counts are sequentially added to a single dense matrix, rather than compiling an exhaustive matrix containg all study-level count data. 

This is much more effecient memory and compute wise, enabling MKDA Chi squared with Neurosynth as reference, using <1GB of memory.

### DIFF
--- a/nimare/meta/cbma/mkda.py
+++ b/nimare/meta/cbma/mkda.py
@@ -571,22 +571,17 @@ class MKDAChi2(PairwiseCBMAEstimator):
         iter_df2[["x", "y", "z"]] = iter_xyz2
 
         # Generate MA maps and calculate count variables for first dataset
-        temp_ma_maps1 = self.kernel_transformer.transform(
+        n_selected_active_voxels = self.kernel_transformer.transform(
             iter_df1, self.masker, return_type="sparse"
         )
-        n_selected = temp_ma_maps1.shape[0]
-        n_selected_active_voxels = self._load_ma_maps(temp_ma_maps1)
 
-        del temp_ma_maps1
+        n_selected = self.dataset1.coordinates["id"].unique().shape[0]
 
         # Generate MA maps and calculate count variables for second dataset
-        temp_ma_maps2 = self.kernel_transformer.transform(
+        n_unselected_active_voxels = self.kernel_transformer.transform(
             iter_df2, self.masker, return_type="sparse"
         )
-        n_unselected = temp_ma_maps2.shape[0]
-        n_unselected_active_voxels = self._load_ma_maps(temp_ma_maps2)
-
-        del temp_ma_maps2
+        n_unselected = self.dataset2.coordinates["id"].unique().shape[0]
 
         # Currently unused conditional probabilities
         # pAgF = n_selected_active_voxels / n_selected
--- a/nimare/meta/cbma/mkda.py
+++ b/nimare/meta/cbma/mkda.py
@@ -388,7 +388,7 @@ class MKDAChi2(PairwiseCBMAEstimator):
 
     def __init__(
         self,
-        kernel_transformer=MKDAKernel,
+        kernel_transformer=MKDAKernel(sum_across_studies=True),
         prior=0.5,
         memory=Memory(location=None, verbose=0),
         memory_level=0,
--- a/nimare/meta/cbma/mkda.py
+++ b/nimare/meta/cbma/mkda.py
@@ -431,28 +431,6 @@ class MKDAChi2(PairwiseCBMAEstimator):
 
         return description
 
-    def _load_ma_maps(self, ma_maps, chunk_size=4000):
-        """Load the MA maps for analysis."""
-        n_maps = ma_maps.shape[0]
-
-        n_chunks = (n_maps + chunk_size - 1) // chunk_size
-
-        n_active_voxels = sparse.COO(np.zeros(ma_maps.shape[1:]))
-
-        for i in range(n_chunks):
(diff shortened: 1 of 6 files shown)


jdkent:
oh shoot, do we actually not need the individual matrices for fwe correction? Looking over the code it seems like we don't, and the first thing done for mkdachi2 is summing across all the studies.

adelavega:
Allright guys @JulioAPeraza @jdkent @tsalo, this is realy exciting.

For MKDAChi2 we don't actually need study level modeled activiation. This realization means we can save massive amounts of memory and substantial time (2-3x speed up over an already optimized PR I was working on). 

When looping over studies, instead of saving all study level (M)KDA modeled activation (MA) maps, we can sum them in-place into a single MNI sized dense volume. 

This means no need to run `np.unique`, which is slow, and that the data is O(1) with number of studies. 

I initally enable this for all MKDAKernels, but it seems to result in incorrect statisics for `MKDADensity`. We could potentially implement this there as well but need to look at stats more careful. We might be able to save intermediary stats to get correct results.

adelavega:
On a measly Macbook Air with 8GB of RAM, I was able to run `MKDAChi2` comparing the example pain dataset to all of Neurosynth in 32 seconds. Only need a little over 1 GB of RAM.

Screen Shot 2024-01-08 at 10 49 29 PM

Someone check my logic in case I'm missing something!

nimare/meta/utils.py
```diff
@@ -119,33 +123,69 @@ def compute_kda_ma(
...
+            curr_exp_idx = exp_idx == i_exp
+            peaks = ijks[curr_exp_idx]
+
+            sphere_coords = _convolve_sphere(kernel, peaks, np.array(shape))
+
+            # Go from sparse to dense
+            study_values = np.zeros(shape, dtype=np.int32)
+
+            if sum_overlap:
+                sphere_coords = unique_rows(sphere_coords)
```
jdkent, line 141:
```suggestion
                study_values[sphere_coords[:, 0], sphere_coords[:, 1], sphere_coords[:, 2]] += value
```

would this be the equivalent for the sum_overlap option?

adelavega: Ah yes, good catch. Although no tests for it since there aren't any kernels that yet need both sum across studies and sum overlap.

adelavega:
I was able to simplify the code further by just returning a dense matrix. Saves ~2 more seconds. 

I'm open to other stylistic suggestions though.

nimare/meta/utils.py
```diff
@@ -119,33 +123,69 @@ def compute_kda_ma(
...
+            else:
+                study_values[sphere_coords[:, 0], sphere_coords[:, 1], sphere_coords[:, 2]] = value
+
+            # Sum across studies
+            all_values += study_values
+
+        # Set voxel outside the mask to zero.
+        all_values[~mask_data] = 0
```
jdkent: and this step is functionally replacing this

```
sphere_idx_inside_mask = np.where(mask_data[tuple(all_spheres.T)])[0]
all_spheres = all_spheres[sphere_idx_inside_mask, :]
```

nimare/meta/utils.py
```diff
@@ -119,33 +123,69 @@ def compute_kda_ma(
...
 
-        # Convolve with sphere
-        all_spheres = _convolve_sphere(kernel, peaks, np.array(shape))
+            # Convolve with sphere
+            all_spheres = _convolve_sphere(kernel, peaks, np.array(shape))
 
-        if not sum_overlap:
-            all_spheres = unique_rows(all_spheres)
+            if not sum_overlap:
+                all_spheres = unique_rows(all_spheres)
```
jdkent, line 165: still need this bit of code here for the original process, no?

```suggestion
        sphere_idx_inside_mask = np.where(mask_data[tuple(all_spheres.T)])[0]
        all_spheres = all_spheres[sphere_idx_inside_mask, :]
```

nimare/meta/utils.py
adelavega: Actually, I'm nots ure this work. 

I suggest we don't support `sum_overlap` since it's not needed by anything

nimare/meta/cbma/mkda.py
```diff
@@ -431,53 +431,26 @@ def _generate_description(self):
...
-        ma_maps1 = self._collect_ma_maps(
+        n_selected_active_voxels = self._collect_ma_maps(
             maps_key="ma_maps1",
             coords_key="coordinates1",
         )
-        n_selected = ma_maps1.shape[0]
-        n_selected_active_voxels = self._load_ma_maps(ma_maps1)
 
-        del ma_maps1
+        n_selected = self.dataset1.coordinates["id"].unique().shape[0]
```
jdkent, line 446: probably should have `_collect_ma_maps` return what the `n_selected` value is, all the coordinates for a particular experiment could exist outside the mask and the experiment would not be included.

jdkent (commented):
this line: 

https://github.com/neurostuff/NiMARE/pull/859/files#diff-953de5154c5b38a0a853444e8525f6d8edc0eef147cb69c54130eb389fc92cf5L187


need to add dtype to image creation:
```
img = nib.Nifti1Image(kernel_data, mask.affine, dtype=kernel_data.dtype)
```

this should fix some of the failing tests

adelavega:
Your changes look good to me, James.

jdkent:
fantastic work! merging it in.

PR #863, opened 2024-01-16
Fix ridgeline plot in IBMA report

JulioAPeraza:
Closes None.

- Replace the custom code based on seaborn, which was producing strange results for specific data.
- Ridgeplot is more straightforward and interactive given that it is based on Plotly.
- Another improvement is the ability to set the color gradient along the horizontal axis, which helps compare the different distributions from the images.

seaborn-ridgeplot

### DIFF
--- a/nimare/reports/figures.py
+++ b/nimare/reports/figures.py
@@ -407,7 +407,6 @@ def _plot_ridgeplot(maps_arr, ids_, x_label, out_filename):
 
     .. versionadded:: 0.2.0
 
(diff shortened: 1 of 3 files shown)


nimare/reports/figures.py
```diff
@@ -34,6 +34,10 @@
 ]
 
 
+PXS_PER_STD = 30  # Number of pixels per study, control the size (height) of Plotly figures
```
jdkent, line 37: yay! thank you for the informative variable names!

jdkent (approved):
This all looks reasonable to me, thanks!

PR #865, opened 2024-01-18
Clarify FWECorrector arguments

adelavega:
The `FWECorrector` was written generally, such that methods beyond bonferoni and montecarlo could be specified if an `Estimator` implements one.

However, as of yet no such alternate method is implemented, and this generic organization obfuscates the API for the `FWECorrector`, as its not clear what possible options are implmeneted by estimators for monte carlo correction. 

Here, I make this API more explicit by:

* Makes montecarlo and bonferroni only possible options for `FWECorrector`, instead of any possible  yet to defined method
* Sets `n_iters` to 1000 across the board
* Makes other `montecarlo` arguments directly available to `FWECorrector`

### DIFF
--- a/nimare/annotate/gclda.py
+++ b/nimare/annotate/gclda.py
@@ -353,7 +353,7 @@ class GCLDAModel(NiMAREBase):
             self.topics["total_n_word_tokens_by_topic"][0, topic] += 1
             self.topics["n_word_tokens_doc_by_topic"][doc, topic] += 1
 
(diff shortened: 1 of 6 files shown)


jdkent (commented):
could n_iters be set in the base_class?

https://github.com/neurostuff/NiMARE/pull/865/files#diff-4cafb2c711a65e5b19450202efbd9250dc086c814fb97588c1fd88afcc1add0cL73-L80

could save having to keep track of the default value in multiple locations

PR #882, opened 2024-04-19
Add correction for multiple contrasts within a study in Stouffer's IBMA

JulioAPeraza:
Closes None.

- Add the correction for multiple contrasts within a study in Stouffer's IBMA, implemented in https://github.com/neurostuff/PyMARE/pull/117

### DIFF
--- a/nimare/meta/ibma.py
+++ b/nimare/meta/ibma.py
@@ -281,6 +281,10 @@ class Stouffers(IBMAEstimator):
 
     This method is described in :footcite:t:`stouffer1949american`.
 
+    .. versionchanged:: 0.2.3
+
+        * Add correction for multiple contrasts within a study.
+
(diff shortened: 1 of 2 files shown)


jdkent:
if you merge in main, that should fix the codecov errors you are seeing!

jdkent (commented):
LGTM, np.tile may blow up the memory usage, looking into an alternative. 

EDIT: it does not appear to have a large impact, but we can revisit changing the API.

JulioAPeraza:
That's a good point. I tried to pass a 1D array since the groups and weights are the same across voxels, but PyMARE didn't allow that.
In the end, we only use the information from one voxel: https://github.com/neurostuff/PyMARE/blob/master/pymare/estimators/combination.py#L145-L146

PR #884, opened 2024-05-17
Add support for the `concordant` mode test in Fisher's and Stouffer's estimators

JulioAPeraza:
Closes None.

- Currently, we only support `directed` tests in Fisher's and Stouffer's estimators. The new parameter `two_sided` (following Nilearn's naming convention) adds support for the `concordant` (two-sided) test.
- That parameter will be used in the Diagnostics module to determine if the diagnostic should be run for z < 0.


JulioAPeraza:
It looks like something changed in Nilearn's `plot_stat_map`. The color bars for maps with only positive values in the example gallery are not symmetric by default: https://nimare--884.org.readthedocs.build/en/884/auto_examples/02_meta-analyses/01_plot_cbma.html#sphx-glr-auto-examples-02-meta-analyses-01-plot-cbma-py

JulioAPeraza:
This could also be addressed using sequential colormaps instead. I open an issue [here](https://github.com/nilearn/nilearn/issues/4421) to see if there is a better way to approach this.

JulioAPeraza:
When I checked the gallery the other day, it looked like the RTD was not properly built. Setting `symmetric_cbar` to `True` in `plot_stat_map` actually solved the issue.

nimare/resources/references.bib
```diff
@@ -529,3 +529,16 @@ @article{geoffroy2001poisson
   year={2001},
   publisher={Taylor \& Francis}
 }
+
+@article{winkler2016non,
```
jdkent, line 533: is this referenced in the permutedOLS class?

nimare/meta/ibma.py
```diff
@@ -337,12 +358,15 @@ class Stouffers(IBMAEstimator):
 
     _required_inputs = {"z_maps": ("image", "z")}
 
-    def __init__(self, use_sample_size=False, **kwargs):
+    def __init__(self, use_sample_size=False, two_sided=True, **kwargs):
```
jdkent, line 361: this changes the default to concordant right? we should make note that this is different from previous behavior (but I do agree with changing this default behavior)

nimare/resources/references.bib
JulioAPeraza: Good catch! I added that when I defined the `mode` parameter, but later, I removed it in favor of `two_sided` and forgot to remove the citation.

JulioAPeraza:
@jdkent Thanks for the review! I added a note for the change in the `versionchanged` field of the docstring. We could also include a sentence in the release note of 0.2.3.

jdkent (approved):
LGTM, thanks @JulioAPeraza!

PR #887, opened 2024-05-23
Add support for publication size weighting. General refactoring of IBMA estimators

JulioAPeraza:
Closes None.

- Add for publication size weighting with a new parameter `use_group_size`.
- In addition, we perform a general refactoring of IBMA estimators to remove duplicate code and reduce the memory in the estimator.


nimare/meta/ibma.py
```diff
@@ -320,6 +320,9 @@ class Stouffers(IBMAEstimator):
         Whether to use sample sizes for weights (i.e., "weighted Stouffer's") or not,
         as described in :footcite:t:`zaykin2011optimally`.
         Default is False.
+    use_group_size : :obj:`bool`, optional
```
jdkent, line 323: I'm having second thoughts about `use_group_size`, group may be more ambiguous than what we are using the variable for. group could mean:

1) patient/control group
2) a group of papers about shizophrenia versus depression

whereas we are using group to mean specifically the number of analyses (contrasts) within a study.

`use_group_size` could be renamed to `normalize_contrast_weights`

and `group_size` to `num_contrasts`

and `group` to `contrast_names`

I think that is more aligned with what the variables are doing.

unless the labels serve more functions than identifying contrasts.

jdkent (changes requested):
LGTM!

I'm trying to think if there would be a case when a contrast wouldn't be included in the actual meta-analysis (e.g., didn't have the correct images for that particular algorithm, or was otherwise filtered from the analysis), but the re-weighing would correct the values as though that contrast was included.

nimare/meta/ibma.py
JulioAPeraza: I 100% agree! I will make the changes in a bit.

JulioAPeraza: > I'm trying to think if there would be a case when a contrast wouldn't be included in the actual meta-analysis (e.g., didn't have the correct images for that particular algorithm, or was otherwise filtered from the analysis), but the re-weighing would correct the values as though that contrast was included.

In the new `_preprocess_input` in Stouffers, the number of contrasts is estimated from `self.inputs_["id"]`, so any excluded images will also be excluded from the count.

jdkent (approved):
LGTM, thanks @JulioAPeraza!

PR #889, opened 2024-06-01
Compute correlation matrix in signal voxels

### LINKED ISSUE #888: The correlation matrix should be calculated in signal voxels
## Summary

I now notice that we calculate the correlation matrix in Stouffers in all the voxels, even the ones with 0 or Nan. I think we can apply the aggressive mask to all images and then calculate the correlation matrix. We can also save that matrix in the estimator because we will use it in the report module.

## Additional details

Saving the matrix will decrease the computation time in the Workflows but at the cost of extra memory in the estimator class. 

## Next steps

Apply the aggressive mask and calculate the correlation matrix in _preprocess_input. This will only affect Stouffers estimator.

JulioAPeraza:
Closes #888.

As documented in the issue, the correlation matrix for calculating the inflation reduction term in Stouffers, and the same matrix for showing the similarity between IBMA studies in the report, should be calculated in signal voxels.

- Here, we calculate that matrix in the Stouffers._preprocess_input method and use the same in the report if it exists.

### DIFF
--- a/nimare/meta/ibma.py
+++ b/nimare/meta/ibma.py
@@ -130,22 +130,21 @@ class IBMAEstimator(Estimator):
                 # in the stimator if self.aggressive_mask is True.
                 self.inputs_[name] = temp_arr
 
-                # Regardless of the masking strategy, we need to determine the good voxels here
-                # to mask the bad ones later when calculating the correlation matrix.
-                nonzero_voxels_bool = np.all(temp_arr != 0, axis=0)
-                nonnan_voxels_bool = np.all(~np.isnan(temp_arr), axis=0)
-                good_voxels_bool = np.logical_and(nonzero_voxels_bool, nonnan_voxels_bool)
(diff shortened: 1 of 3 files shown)


nimare/meta/ibma.py
```diff
@@ -123,11 +126,12 @@ def _preprocess_input(self, dataset):
                 # Mask required input images using either the dataset's mask or the estimator's.
                 temp_arr = masker.transform(img4d)
 
-                # To save memory, we only save the original image array and perform masking later.
+                # To save memory, we only save the original image array and perform masking later
+                # in the stimator if self.aggressive_mask is True.
```
jdkent, line 130:
```suggestion
                # in the estimator if self.aggressive_mask is True.
```

jdkent (commented):
Just a small spelling error, but otherwise looks good!

jdkent:
LGTM, thanks @JulioAPeraza!

PR #925, opened 2025-05-08
Fix the use correlation matrix from the results object when available for Reports

JulioAPeraza:
Closes # .

- Fix the use correlation matrix from the results object when available, when generating reports
-

## Summary by Sourcery

Bug Fixes:
- Fixed correlation matrix sourcing in report generation by using the results object's estimator inputs

### DIFF
--- a/nimare/reports/base.py
+++ b/nimare/reports/base.py
@@ -540,7 +540,7 @@ class Report:
                             rowvar=True,
                         )
                 else:
-                    corr = self.inputs_["corr_matrix"]
+                    corr = self.results.estimator.inputs_["corr_matrix"]
 
                 similarity_table = pd.DataFrame(
                     index=ids_,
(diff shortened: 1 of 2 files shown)


JulioAPeraza:
@jdkent, do you know why we now get this `Too Many Requests` error when using `fetch_neuroquery`?

jdkent:
My slightly unhelpful answer is that github is rate-limiting the number of requests we can make. A solution would be to cache the request load so that the test does not actually download the blob from github each time. (but this does mean the blob will be stored within our repo. Another solution is to mock the response, depending on what the test is doing with the downloaded data.

jdkent (approved):
I fixed the 429 error in #926, so this LGTM!

Thanks!

JulioAPeraza:
thank you, James!

PR #967, opened 2026-01-28
[ENH] neurosynth loading and MKDACHI2 speedup

jdkent:
Closes # .

- support pandas 3.0
- loading time of neurosynth dataset is greatly improved
- time with MKDA algotihms show modest improvment

## Summary by Sourcery

Improve Neurosynth dataset conversion robustness and optimize MKDA Chi-square meta-analytic computations for better performance.

Enhancements:
- Handle missing metadata fields and absent coordinates more robustly when converting Neurosynth data to NiMARE dictionaries.
- Switch sparse feature matrix loading to use dense NumPy arrays compatible with downstream processing.
- Introduce a JIT-compiled spherical convolution helper that writes into a boolean occupancy mask to speed up MKDA-based computations and reduce memory overhead.
- Optimize non-overlap handling in MKDA coordinate generation by using an occupancy mask instead of explicit coordinate deduplication.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -317,10 +317,6 @@ For more information about fetching data from the internet, see :ref:`fetching t
    utils.mni2tal
    utils.unique_rows
 
-.. warning::
-   ``utils.unique_rows`` is deprecated and will be removed in a future release.
-   Use ``numpy.unique(ar, axis=0)`` instead.
-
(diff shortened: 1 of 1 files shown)


jdkent (commented):
just a small issue with the warning still being in the docs.

PR #1067, opened 2026-05-02
[DOCS] add qualitative analysis example

### LINKED ISSUE #1068: document qualitative descriptions of data
Conversations with some users suggest they just want to be able to look at the data without drawing inferences from it, in other words, do a more qualitative analysis, versus an inferential analysis.

In the long tail of research questions where there are not enough studies for a properly powered meta-analysis, should we dutifully keeps our eyes closed and make a call to action for people to do more studies until we hit the pre-agreed number of studies? Maybe. But due to the incentives to find/explore/show _something_, people are going to turn the knob on that threshold down until they reach "significance". 

Am I to say that across 8-15 studies that reported coordinates for a coordinate based meta-analysis that there is no information gained from looking at them until they reach a golden threshold? 

I think there is a path where we can describe results qualitatively without resorting to inferential statistics, and I can write an example demonstrating how to descriptively look at your data without carrying the baggage of inference. I believe the information is valuable, and perhaps it could be the call to arms for directing more research in that area.

In principle, this is data peeking/optional stopping, which would require subsequent inferential statistical analyses to be all the more scrupulous by choosing a stricter alpha. (in most cases today, getting that extra data point is another grant costing a million dollars, so willy nilly adding data points is not as much of a concern).

This may be a problem in the future, as more people share data, adding more open data to your hypothesis until you reach significance is a distinct possibility.

whelp, this may need to be a blog post at some point... but for now, let's give the people what they want.

jdkent:
Closes #1068  .

- is a rose with a p value of 0.051 odorless?
- add a descriptive analysis example

## Summary by Sourcery

Documentation:
- Document a step-by-step workflow for descriptively inspecting ALE stat maps, plotting coordinates, and performing jackknife sensitivity analysis without claiming statistical significance.

### DIFF
--- a/examples/02_meta-analyses/13_plot_qualitative_ale_stat.py
+++ b/examples/02_meta-analyses/13_plot_qualitative_ale_stat.py
@@ -1,0 +1,290 @@
+# flake8: noqa: D205,D400
+"""
+
+.. _metas_qualitative_ale_stat:
+
+=================================================
+Qualitative interpretation of ALE statistic maps
+=================================================
+
+This example shows a conservative way to inspect an ALE result when a
+coordinate-based meta-analysis does not have enough eligible studies for robust
+voxelwise inference.
+
+The key choice is to inspect the raw ``stat`` map descriptively, instead of
+interpreting the ``z`` or ``p`` maps as evidence. For ALE, the ``stat`` map is
+the observed ALE convergence value. Larger values indicate stronger descriptive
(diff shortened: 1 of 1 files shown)


jdkent:
Screenshot from 2026-05-01 22-02-46

[aaeedea.mp3](https://github.com/user-attachments/files/27295777/aaeedea.mp3)

PR #1080, opened 2026-05-18
Add spatial CBMR estimator and tutorial

yifan0330:
Closes # .

-
-

## Summary by Sourcery

Extend CBMR with spatially varying and mixed moderator effects, robust inference, masking controls, and comprehensive tutorial coverage.

New Features:
- Add global, voxelwise, and mixed moderator-effect parameterizations to the CBMR estimator.
- Add voxelwise moderator-effect diagnostics with relative-intensity and intensity-difference maps and plotting support.
- Add robust sandwich and inverse-Fisher inference options across global, voxelwise, and mixed CBMR analyses.
- Add ROI and empirical-incidence masking controls for CBMR estimation and inference.

Bug Fixes:
- Improve CBMR inference stability and preserve fitted results while applying masks and covariance-method changes.
- Ensure coordinate filtering and experiment-level focus data remain aligned through sparse representations.

Enhancements:
- Refactor CBMR result-centered inference to support shared named contrasts and mode-specific outputs.
- Add full and approximate spatial CBMR fitting backends with spatial coefficient summaries and maps.

Build:
- Raise the minimum matplotlib version to 3.8.0.

Documentation:
- Replace the CBMR example with a tutorial covering global, voxelwise, and mixed moderator effects and their inference workflows.
- Add generated documentation for execution times.

Tests:
- Expand CBMR tests to cover spatial fitting, mixed-model inference, masking, sandwich covariance, diagnostics, sparse inputs, and public API validation.
- Mock NeuroVault downloads in I/O tests to make conversion tests self-contained.


jdkent (changes requested):
Should SpatialCBMR be its own class? I think this should be a variant/flag of the original CBMR class, but I'm open to discussion if there is some technical barrier making it difficult to incorporate into the original CBMR class. 

(How much code is reimplemented/borrowed from the original class/how much new code is being introduced?)

examples/02_meta-analyses/12_plot_spatial_cbmr.py
```diff
@@ -0,0 +1,261 @@
...
+from nimare.meta import SpatialCBMREstimator
+from nimare.transforms import StandardizeField
+
+###############################################################################
+# Load Studyset-compatible data
+# -----------------------------------------------------------------------------
+# We simulate a coordinate-based Studyset with the same structure used in the
+# CBMR tutorial: studies have reported foci, sample sizes, diagnosis labels,
+# drug-status labels, and continuous study-level moderators. Spatial CBMR can be
+# computationally heavier than standard CBMR because it keeps experiment-by-voxel
```
jdkent: Are the experiment-by-voxel matrices kept in sparse/dense format? I've found some big speedups/memory reduction when using Compressed Sparsed Row (CSR) matrices.

examples/02_meta-analyses/12_plot_spatial_cbmr.py
jdkent, line 1: The other CBMR example should just be added to, a whole new example is not needed.

jdkent (changes requested):
Okay, taking a deeper look, my opinion on the user consuming API hasn't changed, I still want a single CBMREstimator class. with the effect being specified as follows:

CBMREstimator(moderator_effect="voxelwise|global")

the moderator effect can be voxelwise or global.

However, my opinion on the underlying code has changed. There is a significant difference in the calculation of likelihood. There can be more separation between CBMR and SpatialCBMR in the code. I'm still unsure on how different `Result` and `Inference` classes should be, and if you're adding a whole new class with duplicate methods (but different implementations of those methods), then you should define a base class with those methods stated, and then inherit that base class for CBMR and SpatialCBMR classes.

yifan0330:
> Okay, taking a deeper look, my opinion on the user consuming API hasn't changed, I still want a single CBMREstimator class. with the effect being specified as follows:
> 
> CBMREstimator(moderator_effect="voxelwise|global")
> 
> the moderator effect can be voxelwise or global.
> 
> However, my opinion on the underlying code has changed. There is a significant difference in the calculation of likelihood. There can be more separation between CBMR and SpatialCBMR in the code. I'm still unsure on how different `Result` and `Inference` classes should be, and if you're adding a whole new class with duplicate methods (but different implementations of those methods), then you should define a base class with those methods stated, and then inherit that base class for CBMR and SpatialCBMR classes.

Thanks for the detailed feedback! I agree and have updated the implementation accordingly. Now, the user-facing API still uses a single CBMREstimator interface with:

CBMREstimator(moderator_effect="voxelwise|global")

nimare/meta/cbmr.py
```diff
@@ -2211,28 +2301,46 @@ def _compute_sandwich_covariance(
...
+            residuals = np.nan_to_num(y - mean, nan=0.0, posinf=0.0, neginf=0.0)
+            residuals, correction_factor = cls._apply_sandwich_correction(
+                correction,
+                bread_inverse,
+                moderators,
+                bases,
+                mean,
+                residuals,
+            )
+            meat_matrix = cls._sandwich_meat_matrix(moderators, bases, residuals, meat)
```
jdkent: when will the response be dense?

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
...
+CBMR is a generative framework for estimating smooth activation intensity
+functions from coordinate-based meta-analytic data. The same
+:class:`~nimare.meta.cbmr.CBMREstimator` can parameterize study-level moderator
+effects in two ways:
 
-This tutorial is intended to provide a brief description and example of the CBMR
-algorithm implemented in NiMARE.
+* ``moderator_effect="global"`` estimates one scalar coefficient per moderator.
+  This is the classic CBMR model and answers whether a study-level covariate has
+  an overall effect on the spatial intensity function.
```
jdkent:
```suggestion
* ``moderator_effect="global"`` estimates one scalar coefficient per moderator.  This assumes the effect of the moderator has a uniform impact across the entire brain.
```
Is this accurate to say? I want to make the description more applicable to the users of the algorithm, which will be focused on analyzing the brain.

jdkent: I also want to avoid using the word "study-level covariate", since the covariate/moderator could represent something about the study, or about the specific analysis within the study.

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
...
+* ``moderator_effect="voxelwise"`` estimates a smooth map for each moderator.
+  This spatially varying model asks where the study-level covariate effect is
+  stronger or weaker over voxels.
```
jdkent:
```suggestion
* ``moderator_effect="voxelwise"`` estimates a scalar coefficient _per voxel_ per moderator.
  This assumes the effect of the moderator differentially impacts voxels throughout the brain. This is likely a more accurate assumption, but requires a lot more data for estimation.
```

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
...
 
-For a more detailed introduction to the elements of a coordinate-based meta-regression,
-see the
-`online course <https://www.coursera.org/lecture/functional-mri-2/module-3-meta-analysis-Vd4zz>`_
-or a `brief overview <https://libguides.princeton.edu/neuroimaging_meta>`_.
```
jdkent, line 24: It looks like these links were tangential, so it looks good to remove those.

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
-"""
+r"""
```
jdkent, line 1: other documents do not use the raw text marker, unless you have a specific reason you need raw text and another solution will not work, I would remove this marker.

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
...
 )
-# set up group columns: diagnosis & drug_status
+
 annotations_df = studyset.annotations_df.copy()
 n_rows = annotations_df.shape[0]
 annotations_df["diagnosis"] = [
     "schizophrenia" if i % 2 == 0 else "depression" for i in range(n_rows)
 ]
 annotations_df["drug_status"] = ["Yes" if i % 2 == 0 else "No" for i in range(n_rows)]
-annotations_df["drug_status"] = (
```
jdkent, line 59: are the comments/text changed in this section for a particular reason? I agree that an overview comment about the intention of the code is an improvement over the line by line comments re-telling what the code is actually doing, but I would also like to maintain a top level description for the heading that's rendered as readable text for the user, and not just have comments in the code area.

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
...
-    title="Schizophrenia without drug treatment (FDR corrected)",
-    threshold=scipy.stats.norm.isf(0.05),
-    vmax=30,
+###############################################################################
+# Option 2: voxelwise moderator effects
+# -----------------------------------------------------------------------------
+# The same estimator exposes spatially varying CBMR through
+# ``moderator_effect="voxelwise"``. This option uses the same groups and the same
+# standardized moderators as above, but estimates a smooth map for each moderator
+# within each group. The approximate backend is used here for speed.
```
jdkent: describe the difference between approximate and full briefly here.

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
...
+    alpha=1e-3,
+    damping=1.0,
+    compute_nll=False,
+    device="cpu",  # the full backend also accepts "cuda" if a GPU is available
 )
+voxelwise_results = voxelwise_cbmr.fit(dataset=studyset)
+
+print(voxelwise_results.describe_inference_inputs())
+print(voxelwise_results.sv_moderator_names)
+print(voxelwise_results.describe_sv_effects())
```
jdkent: I want both the global and voxelwise result objects to have the same attributes, the only reason they would need to be differentiated is if we wanted to specify some moderators to be modelled globally and others with a voxelwise pattern. 

If that's something we'd like to support, then we can design around specific moderators being either global or voxelwise within the same model.

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
...
+###############################################################################
+# Plot voxelwise moderator-effect maps
+# -----------------------------------------------------------------------------
+# In the voxelwise model, each moderator has a fitted map in each group. This is
+# the key difference from global CBMR, where moderator inference is summarized in
+# scalar tables.
 
 plot_stat_map(
-    cres.get_map("z_group-DepressionYes_corr-FDR_method-indep"),
+    voxelwise_results.get_map("svModerator_standardized_sample_sizes_group-SchizophreniaYes"),
```
jdkent: If we are separating voxelwise and global maps, then I would like to do it in a BIDS-y way.
like `moderator-global` or `moderator-voxelwise`

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,427 +1,315 @@
...
-# of spatial intensity estimates across all four groups (finding consistent activation
-# regions). Note that only :math:`n-1` contrast vectors are necessary for testing the
-# equality of :math:`n` groups.
-
-contrast_result = results.infer(
-    group_contrasts=[[[1, -1, 0, 0], [1, 0, -1, 0], [0, 0, 1, -1]]],
-    moderator_contrasts=False,
-)
+# The same voxelwise inference helpers can use inverse Fisher information rather
+# than the sandwich estimator.
```
jdkent: why would a user want to do this? Explain what they gain, what different information do they get if they use inverse fisher information versus the sandwich estimator. Does the inverse Fisher information have more assumptions about the model? Is it more computationally efficient than the sandwich estimator.

nimare/meta/cbmr.py
```diff
@@ -81,6 +94,11 @@ def moderators(self):
         """Return fitted moderator names in display order."""
         return tuple(getattr(self.estimator, "moderators", ()) or ())
 
+    @property
+    def sv_moderator_names(self):
+        """Return spatially varying moderator map names."""
+        return tuple(name for name in self.maps if name.startswith("svModerator_"))
+
```
jdkent: instead of spatially varying, the consistent language used more throughout NiMARE is "voxelwise".

nimare/meta/cbmr.py
jdkent, line 1: for public methods (that will be consumed by people using the tool), I want to keep docstrings of the functions that desribe the parameters:

```
      Parameters
        ----------
        group_contrasts : bool, dict, list, tuple, str, or None, optional
            Group homogeneity or comparison specification. Use ``False`` to skip group inference.
        moderator_contrasts : bool, dict, list, tuple, str, or None, optional
            Moderator effect or comparison specification. Use ``False`` to skip moderator
            inference.
        device : str, optional
            Compute device to use for inference. Defaults to the device recorded on the fitted
            estimator.
        """
```

for internal methods that are only called by other methods within the class (and are not meant to be seen by users, add a leading underscore.

nimare/meta/cbmr.py
```diff
@@ -253,24 +293,91 @@ class CBMREstimator(Estimator):
 
     _required_inputs = {"coordinates": ("coordinates", None)}
     _group_column = "_cbmr_group"
+    _valid_moderator_effects = ("global", "voxelwise")
+    _valid_backends = ("full", "approximate")
```
jdkent: is the difference between full and approximate in the documentation/in the docstring?

nimare/meta/cbmr.py
```diff
@@ -671,15 +864,197 @@ def _fit(self, dataset):
...
+                group,
+                moderators,
+                coefficient,
+            )
+        return maps, tables, self._spatial_cbmr_description("approximate")
+
+    @staticmethod
+    def _get_spatial_cbmr_approximate_solver():
+        """Return the approximate solver used by the voxelwise backend."""
+        return fit_spatial_cbmr_approximate
```
jdkent: This should be a propery/attribute, not a method.

jdkent (changes requested):
Still working through the code, nice work!

I have some comments on structure to keep it in line with the philosophy of NiMARE.

docs/sg_execution_times.rst
jdkent, line 1: remove this from the commit

nimare/meta/cbmr.py
```diff
@@ -253,47 +478,169 @@ class CBMREstimator(Estimator):
...
+        return moderator_effect
 
     def __init__(
         self,
         group_categories=None,
         moderators=None,
+        global_moderators=None,
+        voxelwise_moderators=None,
+        global_moderator=None,
+        moderator_effect="voxelwise",
```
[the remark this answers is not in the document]
jdkent: no, this is an explicit backwards breaking change.

nimare/meta/models.py
```diff
@@ -40,6 +40,116 @@ def subset(self, groups):
...
+            linear_predictor = linear_predictor + moderators @ moderator_coef
+        if self.global_moderators_linear is not None and global_moderators is not None:
+            linear_predictor = linear_predictor + self.global_moderators_linear(global_moderators)
+        return linear_predictor
+
+    @staticmethod
+    def _poisson_nll(linear_predictor, foci):
+        """Return the log-Poisson negative log-likelihood."""
+        mean = torch.exp(linear_predictor)
+        return -(foci * linear_predictor - mean).mean()
```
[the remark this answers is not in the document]
jdkent: This should be an explicit choice, do we want to give small groups the weight as large groups?

jdkent:
I'm working on another API refactor, I want models to be flexibly expressed using patsy (R-style notation), I'm looking into it and I think it will bring more cohesion and expressiveness to cbmr.

yifan0330:
> I'm working on another API refactor, I want models to be flexibly expressed using patsy (R-style notation), I'm looking into it and I think it will bring more cohesion and expressiveness to cbmr.

Sounds good. I'll hold this PR and wait until your API refactor finishes? Could you share a link to the refactor so I can follow along? And yes I'm happy to use the patsy formula notation and align this with the new API.

jdkent:
I have another pull request making the Studyset interface more cohesive that will also impact cbmr, I'm finishing that one up, I'll merge that into the spatial sbmr refactor, and then I'll open a pull request to your spatial-cbmr branch.

jdkent:
here is my branch/PR: https://github.com/neurostuff/NiMARE/pull/1104
(It's already merged with your git history, so if you merge it into your branch, it will remove the conflicts; and it should incorporate the fixes you applied)

here's the tutorial (the results also somehow look real-ish relative to the current docs): 
my PR: https://nimare--1104.org.readthedocs.build/en/1104/auto_examples/02_meta-analyses/11_plot_cbmr.html

the current docs: https://nimare.readthedocs.io/en/stable/auto_examples/02_meta-analyses/11_plot_cbmr.html

regardless they are significantly different which is worth noting.

I like how classes are split between files and are given good purpose, feels like natural kinds, and I like the user facing API.

Claude added some matrix algebra that I didn't validate, so you can double check that/add any other ideas you want. but now CBMR should be very expressive.

yifan0330:
> here is my branch/PR: #1104 (It's already merged with your git history, so if you merge it into your branch, it will remove the conflicts; and it should incorporate the fixes you applied)
> 
> here's the tutorial (the results also somehow look real-ish relative to the current docs): my PR: https://nimare--1104.org.readthedocs.build/en/1104/auto_examples/02_meta-analyses/11_plot_cbmr.html
> 
> the current docs: https://nimare.readthedocs.io/en/stable/auto_examples/02_meta-analyses/11_plot_cbmr.html
> 
> regardless they are significantly different which is worth noting.
> 
> I like how classes are split between files and are given good purpose, feels like natural kinds, and I like the user facing API.
> 
> Claude added some matrix algebra that I didn't validate, so you can double check that/add any other ideas you want. but now CBMR should be very expressive.

Thanks so much! I've validated all of the matrix computations, and current implementations all make sense.

The previous doc version had some numerical stability issues around matrix inversion. In the current version, I’ve implemented a few numerical tricks to avoid numerical singularities when inverting the covariance matrices, which seems to have resolved those issues and explains why the highlighted regions now look different but much more reasonable.

I've merged #1104 into this branch to pick up the conflict resolution and fixes. I really like the file split and the user-facing API. Thanks again for the PR!

yifan0330:
@jdkent The latest commit has passed all checks. Could you take a look and let me know if it's ready to be integrated into main? Thanks very much :)

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,428 +1,385 @@
...
-        ("DepressionYes", "DepressionNo"),
-    ]
-)
+# ``s(diagnosis:drug_status)`` gives every cell a free map. The additive alternative says the two
+# factors shift one underlying map independently -- a stronger claim, and far cheaper.
+#
+# Writing it as ``s(diagnosis) + s(drug_status)`` does not work, and CBMR refuses it: each
+# cell-means spatial factor's columns sum to the constant, so their difference is exactly zero
+# and the design is rank deficient by a whole basis width whatever the data. ``sz()`` is the
+# identified form, after mgcv's ``bs="sz"`` basis. It constrains each factor's coefficients to
```
jdkent, line 308: I want to add a citation here to mgcv, but that is not blocking this pull request.

examples/02_meta-analyses/11_plot_cbmr.py
```diff
@@ -1,428 +1,385 @@
...
-    symmetric_cbar=True,
-    title="GLH_groups_0",
-    threshold=scipy.stats.norm.isf(0.4),
-)
-print("The contrast matrix of GLH_0 is {}".format(contrast_result.metadata["GLH_groups_0"]))
+print(overdispersed.tables["overdispersion"])
 
 ###############################################################################
-# GLH testing for study-level moderators
+# Migrating from the old interface
```
jdkent, line 358: I don't think we need this since people have not really used the "old interface" yet.
Again not, blocking, will do in a separate PR. What would be more useful is links to understand the syntax of model specification (to learn more about model notation in general, and the specific applications of `s()` and `sz()`

nimare/meta/cbmr/_helpers.py
jdkent, line 1: the convention for nimare is utils.py, but I honestly like the semi private design of these methods

nimare/meta/cbmr/contrasts.py
jdkent, line 1: reading this and covariance, I'm becoming convinced that foci should be an attribute on CBMRModel instead of being an argument passed around. the foci should never be different than what was fit with the estimator so there is one, and only one, correct value for the argument foci.

nimare/meta/cbmr/model.py
```diff
@@ -0,0 +1,262 @@
+"""Fitting and covariance for a term-based CBMR model.
+
+Holds one flat coefficient vector, sliced according to the design's parameter layout. That is
+not a detail of storage; it is what makes the covariance correct by construction.
```
jdkent:
```suggestion
Holds one flat coefficient vector, sliced according to the design's parameter layout.
```

jdkent: removing ai slop comment

nimare/meta/cbmr/model.py
```diff
@@ -0,0 +1,262 @@
...
+
+Standard errors come from the inverse of the Fisher information. The information over a flat
+vector is a single matrix, so the covariance of a term is a *block of the inverse*. Keeping
+coefficients in separate per-group and per-moderator containers, as the older code does, makes
+it natural to invert each block instead, which drops the cross terms -- and cross terms are not
+small here. A moderator coefficient pooled across groups is correlated with every group's
+spatial coefficients, and ignoring that understated its variance by about 6%, while a separate
+bug that computed the block from only the last group inflated it by around 50%. Both were
+possible because the layout was implicit. Here it is explicit and there is only one matrix to
+invert.
```
jdkent:
```suggestion
```

delete this paragraph, it is just exposition about how design choices were made, not relevant to the module.

nimare/tests/test_io.py
jdkent, line 1: unrelated to this pull request, but because neurovault is flaky (and the API is unlikely to change), this should be monkeypatched.

DEVELOPMENT_CONVENTIONS.md
jdkent, line 1: good details, but should either be in an AGENTS.md or the contributors/developers file.

jdkent (approved):
There are some changes I would ideally prefer, but those can be done in separate smaller pull requests, I'll merge this one in and open issues for the others, since they are more about aesthetics than correctness.

yifan0330:
> There are some changes I would ideally prefer, but those can be done in separate smaller pull requests, I'll merge this one in and open issues for the others, since they are more about aesthetics than correctness.

Thanks for the review and the approval! I will work on the changes in follow-up PRs this week :)

jdkent:
I set claude on a mission to reduce the CPU/memory consumption for CBMR; I'll be uploading the symbolic logic proofs later today for validation. I took a look, but I do not have the expertise to say it's a valid derivation; I did find numerical equivalency in the cases I tested.

jdkent:
got the pull request and validation repo up:
https://github.com/neurostuff/NiMARE/pull/1121
https://github.com/jdkent/cbmr-proofs

PR #1085, opened 2026-06-24
Raise an informative error when a fetch query matches no files

### LINKED ISSUE #640: Raise informative error when `fetch_neuroquery`/`fetch_neurosynth` query does not match available patterns
## Summary

This stems from #639. When users use an incompatible combination of queries, the fetching functions should raise informative errors, such as a list of available patterns or even a link to the pattern file.

lobennett:
`fetch_neurosynth`/`fetch_neuroquery` passed an unmatched query straight through `_fetch_database`, which returned an empty list and silently yielded no data — so a typo'd `source`/`vocab`/`type` looked like an empty result with no hint.

This raises an informative `ValueError` when a query matches no files in the bundled manifest, listing the available values per entity (version/vocab/source/type) for the requested data source. The match happens before any download, so the error is immediate. Adds a small pure helper (`_get_available_entities`) with unit tests, plus offline tests that the fetch functions raise.

Closes #640

### Before / after
```python
>>> nimare.extract.fetch_neurosynth(vocab="not_a_real_vocab", return_type="files")
```
Before:
```
[]        # silent empty result, no hint
```
After:
```
ValueError: No files matched the query {'vocab': 'not_a_real_vocab', 'data': 'neurosynth', 'version': '7'}.
Available options for data-neurosynth are: version: 3, 4, 5, 6, 7;
vocab: LDA100, LDA200, LDA400, LDA50, terms; source: abstract; type: tfidf, weight.
```

## Summary by Sourcery

Fail fast with actionable guidance when database fetch queries do not match any bundled files.

Bug Fixes:
- Raise an informative ValueError when a fetch query matches no files, including available entity values for the requested data source.

Enhancements:
- Add manifest-based entity discovery with source-scoped, sorted options and safe handling of unknown filename segments.

Tests:
- Add unit and offline fetch tests covering entity discovery, informative unmatched-query errors, and fallback messages.

### DIFF
--- a/nimare/extract/extract.py
+++ b/nimare/extract/extract.py
@@ -57,6 +57,42 @@ def _find_entities(filename, search_pairs, log=False):
     return False
 
 
+def _get_available_entities(database_file_manifest, data=None):
+    """Collect the searchable entity values present in the database manifest.
+
+    Parameters
+    ----------
+    database_file_manifest : :obj:`list` of :obj:`dict`
+        Parsed ``database_file_manifest.json``; each entry has a ``"features"``
+        list of dictionaries with a ``"features"`` filename.
+    data : :obj:`str` or None, optional
+        If provided, restrict results to feature files for this data source
+        (e.g. ``"neurosynth"``). Default is None (all sources).
+
+    Returns
+    -------
+    :obj:`dict`
+        Mapping of each searchable entity (``"version"``, ``"vocab"``,
+        ``"source"``, ``"type"``) to a sorted list of its distinct values
+        across the matching feature files. Entities with no values are omitted.
+    """
+    searchable = ["version", "vocab", "source", "type"]
+    values = {key: set() for key in searchable}
+    for database in database_file_manifest:
+        for feature_dict in database.get("features", []):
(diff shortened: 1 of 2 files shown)


lobennett:
Addressed both review points in 538fb9a:

- **`data-None` / empty options in `_fetch_database`:** the message now falls back to a generic scope ("the requested database") when the query has no `data` key, and emits a generic "No matching entries were found … see the database file manifest for valid patterns." message when no entity options are available. It still leads with "No files matched the query …".
- **`_get_available_entities` parser:** replaced the "any hyphenated segment is an entity" logic with a known-key allowlist (`data, version, vocab, source, type`). Unknown `key-value` segments are ignored, so a filename-format change can no longer silently mismap entities. I used an allowlist rather than a regex — it's stricter and avoids importing `re` for a simple `partition('-')`.

Added three tests (unknown-segment ignored, `data-None` omitted, generic fallback); `pytest nimare/tests/test_extract.py` passes (12) and flake8 is clean.

jdkent (approved):
LGTM, thanks @lobennett!

PR #1089, opened 2026-07-22
Implement machine learning workflow

lyraluoyu:
Closes #T010-T029 (Phase 3) .

- Produce usable sklearn datasets from a perfect Studyset, including numeric descriptors, scalar target, grouped train/test split, feature reduction, and estimator fitting. 
-

## Summary by Sourcery

Implement the machine-learning workflow for transforming NiMARE Studysets into grouped, reducible scikit-learn datasets.

New Features:
- Add a scikit-learn-compatible workflow for converting Studysets into modeled-activation feature datasets with targets, groups, descriptors, and provenance.
- Support truncated SVD and atlas-based dimensionality reduction for voxelwise modeled-activation features.
- Add executable examples comparing grouped machine-learning workflows on n-back and flanker task analyses.

Bug Fixes:
- Handle analyses with missing coordinates by either dropping them or retaining aligned zero-valued feature rows.

Enhancements:
- Add reusable map-feature caching and preprocessor construction for leakage-safe integration with scikit-learn pipelines.

Build:
- Package the bundled n-back versus flanker Studyset resources.

Documentation:
- Add documentation examples and README content for machine-learning feature extraction and reduction workflows.

Tests:
- Expand machine-learning coverage for dataset export, feature alignment, caching, reduction methods, missing coordinates, and grouped model selection.


nimare/resources/nback_vs_flanker_studyset_2026-07/analysis_examples.csv
jdkent, line 1: This file is not used in the repo, all this information should be replicated in the parquet files. I think this can be deleted.

examples/05_machine_learning/01_plot_ma_feature_dataset.py
```diff
@@ -5,27 +5,69 @@
...
+train_bunch, test_bunch = extractor.to_sklearn(
+    studyset,
+    map_reducer="truncated_svd",
+    map_reducer_params={"n_components": 50, "random_state": 13},
+)
+
+classifier = LogisticRegression(
+    max_iter=1000,
+    class_weight="balanced",
+    random_state=13,
```
jdkent: `random_state` should be a variable that's defined at the top of the file, like `RANDOM_SEED=13`. then reference that variable instead of calling the number 13 in multiple places.

Perhaps there's another change we can make as well, where the value for random state is re-used inside the `MAFeatureExtractor` class so it can be applied to the map reducer.

examples/05_machine_learning/01_plot_ma_feature_dataset.py
```diff
@@ -5,27 +5,69 @@
...
+classifier.fit(train_bunch.data, train_bunch.target)
+predicted = classifier.predict(test_bunch.data)
 
-print("Masked activation feature dataset public API scaffold:")
-print("\n".join(public_api))
+print(f"Training data shape: {train_bunch.data.shape}")
+print(f"Test data shape: {test_bunch.data.shape}")
+print(f"Training labels: {sorted(set(train_bunch.target))}")
+print(f"Test accuracy: {accuracy_score(test_bunch.target, predicted):.3f}")
+print("Test balanced accuracy: " f"{balanced_accuracy_score(test_bunch.target, predicted):.3f}")
```
jdkent: add the confusion matrix since that's a more informative metric than accuracy.

nimare/tests/test_ml.py
```diff
@@ -7,91 +7,195 @@
...
 
+from nimare.meta.kernel import MKDAKernel
 from nimare.ml import MAFeatureDataset, MAFeatureExtractor, make_map_reducer
 from nimare.nimads import Studyset
 from nimare.utils import get_masker, get_template
 
 
 def _build_studyset_ml_source():
-    """Build the Studyset fixture used by the ML tests.
+    """Build the shared perfect Studyset source used by the ML tests.
```
jdkent: we do not need to document the Studyset as "perfect", please remove the word perfect.

jdkent: remove the word "perfect" below too.

nimare/tests/test_ml.py
```diff
@@ -7,91 +7,195 @@
...
+                            "target_score": target_score,
+                        },
+                        "texts": {"abstract": f"Study {idx} abstract."},
+                        "points": [{"space": "MNI", "coordinates": coordinate}],
                         "images": [],
                     }
                 ],
-            },
-        ],
+            }
```
jdkent: so each study gets one analysis and that analysis has one coordinate? and it appears the coordinates are mirror flip of each other across one or more axes. This is better than just a single study (what was there before).

Please group the coordinates into alpha and beta groups explicitly so one can see the alpha coordinates and beta coordinates clearly.

Even better is to replace alpha and beta with "motor" and "visual" and choose a sample of coordinates that are significant for motor movement (https://neurosynth.org/analyses/terms/motor/) and visual perception (https://neurosynth.org/analyses/terms/visual/)

nimare/tests/test_ml.py
```diff
@@ -7,91 +7,195 @@
...
+        "id": "studyset_ml_source",
+        "name": "Studyset ML source",
+        "masker": get_masker(get_template(space="mni152_2mm", mask="brain")),
+        "studies": studies,
     }
 
 
 def build_studyset():
-    """Build a fresh Studyset from the studyset ML source dict."""
+    """Build a fresh shared perfect Studyset for ML tests."""
```
jdkent: remove perfect

nimare/tests/test_ml.py
```diff
@@ -7,91 +7,195 @@
...
     source = _build_studyset_ml_source()
     return Studyset(source, mask=source["masker"])
 
 
+def build_ma_feature_dataset():
```
jdkent: _build_studyset_ml_source should return a full studyset, not just a dict that looks like a studyset and is processed more in this function. Functions should do "one" thing.

I think there should be a `_build_ml_studyset` function and a `build_ma_feature_dataset` function

nimare/tests/test_ml.py
```diff
@@ -7,91 +7,195 @@
...
+    descriptor_rows = []
+    target = []
+
+    for study in source["studies"]:
+        analysis = study["analyses"][0]
+        coordinate = analysis["points"][0]["coordinates"]
+        ids.append(f"{study['id']}-{analysis['id']}")
+        study_ids.append(study["id"])
+        map_rows.append([coordinate[0], coordinate[1]])
+        descriptor_rows.append([analysis["annotations"]["alpha_label"]])
```
jdkent: instead of using map_rows and descriptor rows, you can use metadata/annotations that are attributes of a Studyset. then you can create some arbitrary numerical output as well, perhaps analyses associated with alpha/motor have on average higher values than analyses associated with beta/visual (or the value has some sort of relationship to the target score as well).

nimare/tests/test_ml.py
```diff
@@ -7,91 +7,195 @@
...
+from sklearn.linear_model import Ridge
 from sklearn.utils import Bunch
 
+from nimare.meta.kernel import MKDAKernel
 from nimare.ml import MAFeatureDataset, MAFeatureExtractor, make_map_reducer
 from nimare.nimads import Studyset
 from nimare.utils import get_masker, get_template
 
 
 def _build_studyset_ml_source():
```
jdkent: once this returns a "Studyset" object, make it a fixture so it can be referenced in other test functions. Here is an example of a fixture: https://github.com/neurostuff/NiMARE/blob/87ebdaad203b046adb2fc08913cf82d8fb5052b0/nimare/tests/conftest.py#L23-L51

the decorator @pytest.fixture makes a function a fixture which makes it accessible/reusable for the rest of the functions without having to recreate/rerun this data generating function.

nimare/tests/test_ml.py
```diff
@@ -7,91 +7,195 @@
...
+            for row_idx, study_id in enumerate(groups):
+                assert study_id in expected_values
+                assert column[row_idx] == expected_values[study_id]
+
+    if estimator is not None:
+        assert target is not None
+        estimator.fit(bunch.data, target)
 
 
 def test_ma_feature_dataset_initialization():
```
jdkent: use a studyset fixture, do not create another partial studyset object like you are here.

nimare/tests/test_ml.py
```diff
@@ -191,73 +334,307 @@ def __len__(self):
...
-        ds.apply_map_reducer(object())
-
-    with pytest.raises(NotImplementedError):
-        ds.copy()
+    assert bunch.data is ds.features
+
+
+def test_ma_feature_dataset_split():
+    """Dataset grouped split preserves row alignment and avoids study leakage."""
+    ds = build_ma_feature_dataset()
```
jdkent: `build_ma_feature_dataset`  should also be a fixture.

nimare/tests/test_ml.py
```diff
@@ -191,73 +334,307 @@ def __len__(self):
...
-    """Test that MAFeatureExtractor methods currently raise NotImplementedError."""
+def test_ma_feature_extractor_studyset_values_alignment():
+    """Studyset table extraction preserves row-aligned IDs and field values."""
+    source = _build_studyset_ml_source()
+    studyset = Studyset(source, mask=source["masker"])
     extractor = MAFeatureExtractor(kernel_transformer=object())
 
-    with pytest.raises(NotImplementedError):
-        extractor._get_studyset_tables(build_studyset())
+    tables = extractor._get_studyset_tables(studyset)
```
jdkent: I don't think we need a `_get_studyset_tables` function, these attributes are accessible from the `Studyset` object itself. if there's a dataview that `Studyset` needs to support that is not currently supported, then we can add functionality to Studyset.

nimare/tests/test_ml.py
```diff
@@ -191,73 +334,307 @@ def __len__(self):
...
+    target_by_study = {
+        study["id"]: study["analyses"][0]["annotations"][target_name]
+        for study in source["studies"]
+    }
+    extractor = MAFeatureExtractor(
+        kernel_transformer=MKDAKernel(r=4, value=1),
+        descriptor_fields=[{"source": "annotations", "field": descriptor_name}],
+        target_field={"source": "annotations", "field": target_name},
+        test_size=0.33,
+        random_state=13,
```
jdkent: in general numbers that are not attached to variables are called magic numbers and are in general, discouraged.

TEST_DATA_PERCENTAGE
RANDOM_SEED

those variables can be used instead.

jdkent (changes requested):
This is a partial review,  I have not gotten to ml.py yet.

nimare/ml.py
```diff
@@ -291,13 +437,89 @@ def transform(self, studyset: Any):
...
-            This public API is scaffolded only.
         """
-        raise NotImplementedError("MAFeatureExtractor.transform is not yet implemented.")
+        ids = np.asarray(studyset.ids, dtype=str)
+        study_ids = (
+            studyset.metadata.drop_duplicates("id")
+            .set_index("id")["study_id"]
+            .reindex(ids)
+            .to_numpy(dtype=str)
+        )
```
jdkent: there is studyset.study_ids you can use here instead.

nimare/ml.py
```diff
@@ -291,13 +437,89 @@ def transform(self, studyset: Any):
...
+        if isinstance(kernel_transformer, type):
+            kernel_transformer = kernel_transformer()
+
+        cache_key = (id(studyset), id(self.kernel_transformer))
+
+        if self.cache_maps and cache_key in self._map_cache:
+            map_features = self._map_cache[cache_key].copy()
+        else:
+            map_features = kernel_transformer.transform(studyset, return_type="sparse")
+            map_features = sparse.csr_matrix(map_features)
```
jdkent: this line is a no-op since `map_features` is already a csr matrix from the previous command.

nimare/ml.py
```diff
@@ -257,16 +381,38 @@ def __init__(
...
+
+    @staticmethod
+    def _get_selected_values(studyset: Any, selector: Any):
+        """Return selected Studyset values aligned to analysis IDs."""
+        if isinstance(selector, dict):
+            source = selector.get("source")
+            field = selector.get("field")
+        else:
+            source = getattr(selector, "source", None)
+            field = getattr(selector, "field", None)
```
jdkent: simplify, let's only accept dicts here.

nimare/ml.py
jdkent:
```suggestion
        target_field: Dict| None = None,
```

jdkent:
```suggestion
        descriptor_fields: list[Dict] | None = None,
```

nimare/ml.py
```diff
@@ -291,13 +437,89 @@ def transform(self, studyset: Any):
...
+
+        target = None
+        if self.target_field is not None:
+            values, _ = self._get_selected_values(studyset, self.target_field)
+            target_values = []
+
+            for value in values:
+                target_values.append(value)
+
+            target = np.asarray(target_values, dtype=object)
```
jdkent: if values should be a np array, then that could be the output of `_get_selected_values`, you do not need to do the other computations.

nimare/ml.py
```diff
@@ -257,16 +381,38 @@ def __init__(
...
+        }
+        table = source_to_table.get(source)
+        if table is None:
+            raise ValueError(f"Unsupported field selector source: {source}")
+
+        if field not in table.columns:
+            raise ValueError(f"Field '{field}' not found in Studyset {source}.")
+
+        ids = np.asarray(studyset.ids, dtype=str)
+        values = table.drop_duplicates("id").set_index("id")[field].reindex(ids)
```
jdkent: should not have to drop_duplicates/set_index/reindex, just get the column name from either metadata or annotations_df.

nimare/ml.py
```diff
@@ -44,18 +57,13 @@ def __init__(
...
-        try:
-            n_rows = int(features.shape[0])
-            n_cols = int(features.shape[1])
-        except Exception:
-            try:
-                n_rows = int(len(features))
-                n_cols = int(len(features[0])) if n_rows > 0 else 0
-            except Exception:
-                raise ValueError("Unable to determine number of rows or columns from features")
+        n_rows, n_cols = self._get_feature_shape(features)
```
jdkent:
```suggestion
        n_rows, n_cols = features.shape
```

nimare/ml.py
```diff
@@ -71,12 +79,80 @@ def __init__(
             if len(feature_names) != n_cols:
                 raise ValueError("feature_names length must match number of columns in features")
 
+        if map_features is None:
+            map_features = features
+
+        n_map_rows, _ = self._get_feature_shape(map_features)
```
jdkent:
```suggestion
        n_map_rows = map_features.shape[0]
```

nimare/ml.py
jdkent: We can remove `cv`, we will only want to use GroupShufleSplit

nimare/ml.py
```diff
@@ -71,12 +79,80 @@ def __init__(
...
+
+        if descriptor_features is not None:
+            n_descriptor_rows, _ = self._get_feature_shape(descriptor_features)
+            if n_descriptor_rows != n_rows:
+                raise ValueError(
+                    "descriptor_features row count must match number of rows in features"
+                )
+
         self.features = features
         self.ids = list(ids)
```
jdkent: `studyset.ids` could just be passed around, type coercion into a list is not necessary.

nimare/ml.py
```diff
@@ -321,13 +543,28 @@ def to_sklearn(
...
+                if map_reducer_params:
+                    reducer.set_params(**map_reducer_params)
+
+        if reducer is not None:
+            train_dataset = train_dataset.apply_map_reducer(reducer, fit=True)
+            if test_dataset is not None:
+                test_dataset = test_dataset.apply_map_reducer(reducer, fit=False)
+
+        train_bunch = train_dataset.to_sklearn()
+        test_bunch = None if test_dataset is None else test_dataset.to_sklearn()
```
jdkent: I do not want to call `to_sklearn` within `to_sklearn`, if we are running map_reducer, run it within the transform function, we don't need to keep the full voxelwise features if we are reducing the voxels to smaller components.

examples/05_machine_learning/01_plot_ma_feature_dataset.py
```diff
@@ -5,27 +5,74 @@
...
-public_api = (
-    MAFeatureDataset.__name__,
-    MAFeatureExtractor.__name__,
-    make_map_reducer.__name__,
+# ``comparison_task`` is a metadata field with labels ``"n-back"`` and
+# ``"flanker"``. The split is grouped by Studyset study ID, so analyses from
+# one study cannot appear in both partitions.
+extractor = MAFeatureExtractor(
+    kernel_transformer=MKDAKernel(r=10),
+    target_field={"source": "metadata", "field": "comparison_task"},
```
jdkent, line 47: how was "comparison_task" generated? was that something I made?

jdkent (changes requested):
good work! My high level feedback is to see if it's necessary to cast data from one type to another (e.g., list to numpy array to list), and be a little more strict on type annotations for inputs, if you use a particular data type in the examples/tests, stick to those types (use Any less). I would prefer you make a guess on what datatype(s) you think are valid for an argument, and then broaden it to support more datatypes as necessary as opposed to trying to support everything. This suggestion goes against what I've said before about public APIs being generous on what they accept (so people don't have change their preferred way of formatting data to work with our API), but I think for making the first pass, being stricter on inputs makes it easier to reason about the code.

simbot912:
This is a really clean ML workflow integration! Converting Studysets to sklearn Bunches with grouped splitting and map feature reduction is exactly what the neuroimaging meta-analysis community needs. A few thoughts:

1. The truncated-SVD map reducer is a smart choice for dimensionality reduction — voxelwise features are massively overcomplete. Have you considered comparing against UMAP or Hessian Eigenmaps for non-linear reduction?

2. The n-back vs flanker classification example is a great demo. Would be interesting to see how this performs on more heterogeneous task categories (e.g., emotion vs cognition).

3. The parquet-backed Studyset resource is a nice touch for reproducibility.

Solid work on the leakage-safe splitting too — that's a common pitfall in neuroimaging ML.

PR #1092, opened 2026-08-20
[ENH] Add d, g and g_var targets to ImageTransformer

jdkent:
Standardized effect sizes were reachable only inside FixedEffectsHedges, which derives them internally and is fixed-effect only. Exposing them as transform targets makes the pair a meta-regression takes -- g in the effect slot, g_var in the variance slot -- available to any estimator, which matters when studies report betas in pipeline-specific units that no varcope can reconcile.

All three resolve from t and the sample size, and t already resolves from z, so a studyset holding nothing but z maps can reach them. The conversion is the one-sample d = t / sqrt(N); more than one group sample size now warns, since a between-group contrast needs both sizes separately.

Closes # .

-
-

## Summary by Sourcery

Expose standardized effect sizes and their variance as image transformation targets for broader meta-analysis workflows.

New Features:
- Add ImageTransformer and transform_images support for Cohen’s d, Hedges’ g, and the sampling variance of g.
- Enable standardized effect-size targets to be derived from t or z maps and study sample sizes for use in meta-regression.

Enhancements:
- Warn when standardized effect sizes are derived from multiple group sample sizes using the one-sample conversion.

Documentation:
- Document the new standardized effect-size transformation targets and their intended use.

Tests:
- Add coverage for direct d, g, and g_var conversions, z-to-g resolution, missing sample sizes, and multiple-group warnings.

### DIFF
--- a/nimare/tests/test_transforms.py
+++ b/nimare/tests/test_transforms.py
@@ -1,5 +1,6 @@
 """Test nimare.transforms."""
 
+import logging
 import re
 
 import nibabel as nib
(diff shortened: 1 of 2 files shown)


jdkent (commented):
the transform is inefficient since the same code will need to run to form both g and g_var, one value is just being thrown away. only returning one type of map is consistent with the current transformer behavior, so I'll keep it for now, but will want to refactor to make this more efficient.

PR #1096, opened 2026-08-20
[FIX] Keep the combination tests' z maps finite when the combined p u…

jdkent:
…nderflows

PyMARE reports z as isf(p), and the combined p is exactly zero in double precision beyond about z = 38, so z came back infinite -- 332 voxels of the 21-study pain studyset for Fisher's. An infinite z reached the results map, where it breaks thresholding, plotting and any maximum over the map, while the p and logp maps were already clipped and so looked fine.

Bounded at the largest z isf returns for a representable p-value, with the sign preserved and a warning naming the voxel count. This caps the reported statistic rather than recovering the true one; computing z accurately past that point needs log-space evaluation (chi2.logsf and ndtri_exp) inside PyMARE.

Closes # .

-
-

## Summary by Sourcery

Keep combination-test z maps finite when extreme combined p-values underflow.

Bug Fixes:
- Prevent combination-test z maps from containing infinite values when combined p-values underflow to zero.
- Preserve the sign of underflowed z values and emit a warning identifying affected voxels.

Enhancements:
- Define the finite double-precision limit for normal inverse-survival z values and apply it consistently across combination estimators.

Tests:
- Add coverage for finite, sign-preserving z values, warning behavior, unaffected voxels, and the double-precision z limit.

### DIFF
--- a/nimare/meta/ibma.py
+++ b/nimare/meta/ibma.py
@@ -12,6 +12,7 @@ import pymare
 from joblib import Memory
 from nilearn.image import concat_imgs, resample_to_img
 from nilearn.maskers import NiftiMasker
+from scipy import stats as sp_stats
 
 try:
     # nilearn >= 0.13.0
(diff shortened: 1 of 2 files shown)


jdkent:
closing in favor of #1099

PR #1097, opened 2026-08-20
[ENH] Evaluate the p-to-z tail in log space

jdkent:
Three ceilings were bounding z, none of them a property of the float32 the maps are stored in:

- t_to_z floored its internal p-value at the machine epsilon of its dtype. Epsilon measures precision, not magnitude, so |z| saturated at about 8.13: a t of 1000 on 20 dof reported 8.13 rather than 14.63. Now evaluated through logsf and ndtri_exp, which agrees with the previous values to 1e-15 below the old bound and keeps going above it.

- p_to_z bottoms out at the smallest positive float32, i.e. a z of 14.1, because a p-value that small is all a float32 can hold. log_p_to_z takes the log p-value a caller already has from logsf and never materializes the p, so that bound does not apply.

- _clip_logp_values clipped -log10(p) at -log10(smallest positive p), 44.85, which is again a z of 14.1. Clipping a logarithm to the range of the unlogged quantity discards the headroom the logarithm buys; float32 holds a -log10(p) of 5000, describing a z of about 152. Only the non-finite ends are clipped now.

The p-value itself is still bounded, and inherently: 1.4e-45 in float32 is a z of 14.1, and 4.9e-324 in float64 is 38.5. So -log10(p), not p, is the map that can carry a deep tail.

Closes # .

-
-

## Summary by Sourcery

Preserve deep statistical tails by evaluating p-to-z transformations and t-to-z conversions in log space.

New Features:
- Add a log-space p-value to z-score conversion that preserves extremely small tail probabilities without materializing the p-value.

Bug Fixes:
- Remove artificial z-score saturation in t-to-z conversion caused by flooring tail probabilities at machine epsilon.
- Allow stored -log10(p) values to retain deep-tail results beyond the representable p-value range.

Enhancements:
- Evaluate t-distribution tails in log space while preserving prior results below the former ceiling and maintaining signed, antisymmetric z-scores.

Documentation:
- Document the log-space conversion and the expanded tail behavior of t-to-z and -log10(p) values.

Tests:
- Add coverage for conversion accuracy below the former ceiling, deep-tail behavior, tail validation, and preservation of large finite -log10(p) values.

### DIFF
--- a/nimare/tests/test_transforms.py
+++ b/nimare/tests/test_transforms.py
@@ -6,8 +6,10 @@ import re
 import nibabel as nib
 import numpy as np
 import pytest
+from scipy import stats
 
 from nimare import transforms
+from nimare.utils import DEFAULT_FLOAT_DTYPE
 
 
 def test_ImageTransformer(testdata_ibma):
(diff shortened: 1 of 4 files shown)


jdkent:
closing in favor of #1099

PR #1098, opened 2026-08-20
Correct off-by-one indexing in GCLDA log-likelihood calculation and replace LAPACK matrix inversion

tsalo:
Closes none, but addresses two minor issues identified by Claude when I was porting GCLDA to Rust.

- Replace mistaken "1-indexing" with actual 0-indexing in GCLDA log-likelihood estimation.
- Replace LAPACK matrix inversion with closed-form adjugate inverse.

Notes from Claude:

### log-likelihood off-by-one (correctness bug)

`compute_log_likelihood` contains, at `gclda.py:984`, `:1020`, and `:1022`:

```python
doc = self.data["ptoken_doc_idx"][i_ptoken] - 1   # "convert didx from 1-idx to 0-idx"
word_token = self.data["wtoken_word_idx"][i_wtoken] - 1
doc = self.data["wtoken_doc_idx"][i_wtoken] - 1
```

These indices are built by `docidx_mapper = {id_: i for (i, id_) in enumerate(ids)}` and are
**already 0-indexed**. Subtracting 1 maps document 0 to index -1, which wraps to the last
document. Every reported log-likelihood is computed against shifted indices.

This does not corrupt the fitted model — log-likelihood never feeds back into sampling — but
every logged, stored, and plotted likelihood value is wrong.

### LAPACK's 3x3 inverse is NOT reproducible in scalar arithmetic — CONFIRMED PROBLEM

A naive scalar LU with partial pivoting matched `np.linalg.inv` bit-for-bit in only **104 of
3000** trials (worst relative difference 4.0e-13); `slogdet` matched in 2717 of 3000. NumPy
dispatches to OpenBLAS, whose blocked and vectorized kernels depend on build configuration, CPU
features, and threading.

Because these precision matrices feed sampling probabilities directly, a single 1-ulp difference
can flip a sampled index, and the divergence then cascades through every subsequent iteration.

**Resolution — change both sides to a closed-form 3x3 adjugate inverse with a fully specified
operation order.** Peak coordinates are always `(x, y, z)`, so the covariance is always 3x3.
Measured against LAPACK over 3000 random regularized covariances:

| Property | Closed-form adjugate |
|---|---|
| Accuracy vs LAPACK | 4.4e-13 relative (same class as any LU) |
| Self-reproducibility | 3000 / 3000 identical bits |
| Symmetry of result | **Exactly symmetric** (LAPACK's result is not) |
| Speed, 40000 inverse+logdet | **0.064 s vs 0.158 s — 2.5x faster** |

This makes bit-exactness a designed property rather than a coincidence. It also removes 200
LAPACK calls per iteration from the Python hot path and produces the exactly-symmetric inverse
that a covariance inverse mathematically should be. It is a genuine improvement to shipped
NiMARE behavior, not merely a porting convenience.

## Summary by Sourcery

Correct GCLDA likelihood indexing and use deterministic closed-form covariance inversion for reproducible spatial calculations.

Bug Fixes:
- Correct GCLDA log-likelihood calculations to use the already zero-based document and token indices without shifting them.
- Preserve positive-definiteness validation while replacing LAPACK-based covariance inversion with a deterministic closed-form 3x3 inverse and log-determinant calculation.

Enhancements:
- Make GCLDA spatial parameter computation reproducible and ensure inverses of symmetric covariance matrices are exactly symmetric.

Tests:
- Add coverage validating the closed-form inverse against LAPACK, exact symmetry and determinism, and rejection of non-positive-definite matrices.

### DIFF
--- a/nimare/annotate/gclda.py
+++ b/nimare/annotate/gclda.py
@@ -34,6 +34,59 @@ def _sample_from_unnormalized(weights):
     return weights.shape[0] - 1
 
 
+def _inv3_logdet(sigma):
+    """Invert a 3x3 matrix in closed form and return its log-determinant.
+
+    Uses the adjugate formula with a fixed operation order. Unlike LAPACK,
+    this is bit-for-bit reproducible across platforms and BLAS builds, which
+    allows the Rust implementation to match Python exactly. It is also
+    exactly symmetric for symmetric input, and about 2.5x faster than the
+    ``inv`` + ``slogdet`` pair it replaces.
+
+    Parameters
+    ----------
+    sigma : (3, 3) :obj:`numpy.ndarray`
+        A symmetric positive-definite matrix.
+
+    Returns
+    -------
+    inv : (3, 3) :obj:`numpy.ndarray`
+        The matrix inverse.
+    logdet : :obj:`float`
+        The natural log of the determinant.
+    """
+    a00, a01, a02 = sigma[0, 0], sigma[0, 1], sigma[0, 2]
+    a10, a11, a12 = sigma[1, 0], sigma[1, 1], sigma[1, 2]
+    a20, a21, a22 = sigma[2, 0], sigma[2, 1], sigma[2, 2]
+
+    c00 = a11 * a22 - a12 * a21
+    c01 = a02 * a21 - a01 * a22
+    c02 = a01 * a12 - a02 * a11
+    c10 = a12 * a20 - a10 * a22
+    c11 = a00 * a22 - a02 * a20
+    c12 = a02 * a10 - a00 * a12
+    c20 = a10 * a21 - a11 * a20
+    c21 = a01 * a20 - a00 * a21
+    c22 = a00 * a11 - a01 * a10
+
+    det = a00 * c00 + a01 * c10 + a02 * c20
+    if not det > 0.0:
+        raise np.linalg.LinAlgError("Region covariance must be positive definite.")
+
+    inv = np.empty((3, 3), dtype=np.float64)
+    inv[0, 0] = c00 / det
+    inv[0, 1] = c01 / det
+    inv[0, 2] = c02 / det
+    inv[1, 0] = c10 / det
+    inv[1, 1] = c11 / det
+    inv[1, 2] = c12 / det
+    inv[2, 0] = c20 / det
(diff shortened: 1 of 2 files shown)


nimare/annotate/gclda.py
```diff
@@ -34,6 +34,59 @@ def _sample_from_unnormalized(weights):
...
 
 
+def _inv3_logdet(sigma):
+    """Invert a 3x3 matrix in closed form and return its log-determinant.
+
+    Uses the adjugate formula with a fixed operation order. Unlike LAPACK,
+    this is bit-for-bit reproducible across platforms and BLAS builds, which
+    allows the Rust implementation to match Python exactly. It is also
+    exactly symmetric for symmetric input, and about 2.5x faster than the
+    ``inv`` + ``slogdet`` pair it replaces.
```
jdkent:
```suggestion
    Uses the adjugate formula with a fixed operation order. Unlike LAPACK,
    this is bit-for-bit reproducible across platforms and BLAS builds.
```

nimare/tests/test_annotate_gclda.py
```diff
@@ -58,6 +58,95 @@ def test_gclda_symmetric(testdata_laird):
...
+def test_gclda_loglikelihood_uses_zero_indexed_tokens(testdata_laird):
+    """Log-likelihood must index documents/words directly, not offset by one.
+
+    Regression test: the indices produced by docidx_mapper are already
+    0-indexed, so subtracting 1 wrapped document 0 to the final document.
+    """
+    counts_df = annotate.text.generate_counts(
+        testdata_laird.texts, text_column="abstract", tfidf=False, min_df=1, max_df=1.0
+    )
+    model = annotate.gclda.GCLDAModel(
```
jdkent: Now that we're doing correct indexing, I don't think we need a test, the code is unlikely to drift in that direction again

jdkent (changes requested):
thanks! just a couple of edits to mitigate the overzealousness of claude.

nimare/tests/test_annotate_gclda.py
tsalo: I'll remove it. Thanks for taking a look!

jdkent (approved):
LGTM, thanks @tsalo!

PR #1108, opened 2026-08-22
[FIX] warn unsigned p to z

jdkent:
Closes # .

-
-

## Summary by Sourcery

Preserve available sign information during p-to-z conversion and clearly warn when converted z maps are unsigned.

Bug Fixes:
- Preserve the direction of z maps converted from p maps when a corresponding t map is available, even without sample-size information.
- Warn when converting p maps without directional information would produce unsigned z values, including the affected analysis identifier.

Documentation:
- Document the behavior and limitations of p-to-z conversions when directional information is unavailable.

Tests:
- Add coverage for signed and unsigned p-to-z conversions and analysis-specific warnings.

### DIFF
--- a/nimare/transforms.py
+++ b/nimare/transforms.py
@@ -273,7 +273,7 @@ def transform_images(images_df, target, masker, metadata_df=None, out_dir=None, 
                     available_data[k] = v
 
         # Get converted data
-        img = resolve_transforms(target, available_data, new_masker, id_=id_)
+        img = resolve_transforms(target, available_data, new_masker)
         if img is not None:
             if overwrite or not op.isfile(new_file):
                 img.to_filename(new_file)
@@ -291,14 +291,13 @@ def transform_images(images_df, target, masker, metadata_df=None, out_dir=None, 
     return new_images_df
 
 
-def resolve_transforms(target, available_data, masker, id_=None):
+def resolve_transforms(target, available_data, masker):
     """Determine and apply the appropriate transforms to a target image type from available data.
 
     .. versionchanged:: 0.21.0
 
         * [FIX] Take the sign from a t map when converting a p map without a sample size.
         * [ENH] Warn when ``z`` is converted from a p map with no sign available.
-        * [ENH] Accept ``id_``, to name the analysis in that warning.
 
     .. versionchanged:: 0.0.8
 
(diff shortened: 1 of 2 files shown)


nimare/transforms.py
```diff
@@ -333,13 +350,33 @@ def resolve_transforms(target, available_data, masker):
...
+                LGR.warning(
+                    f"{prefix}Deriving 'z' from a p map, with nothing in the analysis to "
+                    "give the direction. A p-value carries no sign, so the result is "
+                    "unsigned: every voxel is positive. In a signed test (Stouffers, "
+                    "Fishers) this analysis then contributes positive evidence only, "
+                    "whichever way its contrast ran. The p-values are read as two-tailed "
+                    "as well, so a one-tailed p map comes out with the wrong magnitude. "
+                    "Supply a z map, or a t map: with a sample size it sets the "
+                    "magnitude, without one it still sets the sign."
+                )
```
jdkent, line 364:
```suggestion
                LGR.warning(
                    f"{prefix}Deriving 'z' from a p map, with nothing in the analysis to "
                    "give the direction. A p-value carries no sign, so the result is "
                    "unsigned. This will be invalid input for signed tests (like Stouffers and Fishers)."
                )
```

jdkent (commented):
changed the wording of the warning, otherwise looks good

nimare/transforms.py
```diff
@@ -333,13 +350,28 @@ def resolve_transforms(target, available_data, masker):
...
+                # exactly 0 the voxel goes to 0: sign(0) is 0, and a t of 0 is a p of 1,
+                # whose z is 0 anyway, so the two agree. Non-finite t propagates.
+                t = masker.transform(available_data["t"])
+                z = np.sign(t) * z
+            else:
+                prefix = f"{id_}: " if id_ is not None else ""
+                LGR.warning(
+                    f"{prefix}Deriving 'z' from a p map, with nothing in the analysis to "
+                    "give the direction. A p-value carries no sign, so the result is "
+                    "unsigned. This will be invalid input for signed tests (like Stouffers and Fishers)."
```
jdkent:
```suggestion
                    "unsigned. "
                    "This will be invalid input for signed tests (like Stouffers)."
```

jdkent (commented):
LGTM!

PR #1109, opened 2026-08-22
[FIX] Drop None-valued Corrector kwargs and validate the rest against the m…

jdkent:
…ethod

FWECorrector and FDRCorrector collect every unrecognised keyword into ``self.parameters`` and forward it verbatim to whichever correction method ``transform`` ends up calling. Two things go wrong there.

An unset parameter arrives as ``None``, which is what the neurosynth-compose frontend sends for a field the user left blank. Storing it means the caller overrides the correction method's own default with nothing: a ``voxel_thresh`` of ``None`` crashes PermutedOLS, whose ``correct_fwe_montecarlo`` has no such argument, and quietly displaces ALE's 0.001 until it reaches ``_p_to_summarystat(None)``. Between them that was the only FWE path either estimator has. ``n_iters`` already had this rule; apply it to every keyword, in ``Corrector.__init__`` so FDR gets it too.

The keywords that remain were still only checked when Python bound them to the correction method, deep inside a run that may have taken hours to reach that point -- and not checked at all when the method is one of the Corrector's own, which take no keywords and so dropped them silently. The estimator is not known at construction, but it is by ``transform``, so validate there against the signature of the method about to be called and name it in the error. Methods declaring ``**kwargs`` accept anything and are left alone.

Closes # .

-
-

## Summary by Sourcery

Make correction parameter handling defer unset values, forward configured estimator parameters, and reject unsupported keywords before correction runs.

Bug Fixes:
- Drop None-valued correction parameters so estimator defaults remain effective.
- Validate correction keywords during transform against the selected correction method and report unsupported arguments early.
- Forward FDR alpha values to estimator-implemented correction methods.

Enhancements:
- Support correction methods with **kwargs while preserving method-specific keyword validation for other callables.
- Add coverage for omitted parameters, estimator and native methods, permissive methods, and validation errors.

Documentation:
- Document None-parameter handling and transform-time validation for correction methods.

### DIFF
--- a/nimare/correct.py
+++ b/nimare/correct.py
@@ -12,6 +12,7 @@ from nimare.stats import nlogp_bonferroni, nlogp_fdr
 from nimare.transforms import nlogp_to_z
 from nimare.utils import (
     DEFAULT_FLOAT_DTYPE,
+    _acceptable_kwargs,
     _clip_p_values,
     _minimum_positive_float,
     _nlogp_to_logp_values,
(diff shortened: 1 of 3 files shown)


jdkent (commented):
LGTM

jdkent (commented):
LGTM!

PR #1110, opened 2026-08-23
[FIX] Drop a dependence group that cancels instead of failing the run

jdkent:
One study's uploads could end a whole image-based meta-analysis. When a group's images cancel -- their block of the estimated null correlation sums to zero -- PyMARE refuses to aggregate it and raises, naming the internal integer code and pointing the reader at "the correlation matrix supplied for this group", which the caller never supplied: NiMARE estimated it. On real staging data this fired on 3 of 16 studysets, in two shapes: an exact mirror pair (the same contrast uploaded both ways round, correlation -1), and three complementary network maps from one cohort whose pairwise correlations were unremarkable but whose 3x3 block still summed to nothing.

Screen each model before handing it to PyMARE, and exclude a group that cancels from that model rather than letting it end the run. The whole group, not one member: with three or more there is no principled choice of which to remove, and a group whose contrasts carry no joint signal has nothing to contribute either way.

Per liberal-mask bag, not per meta-analysis. A bag holds only the images valid over its voxels, so whether a group cancels is a property of the bag -- a two-image group can only cancel where both images are valid, and is a lone contrast everywhere else. Dropping globally would discard images over voxels where nothing is wrong with them.

The drop lives in _fit_over_bags, which already holds study_mask and the bag's arrays and already skips a bag left without two independent groups. Filtering study_mask alongside the data means everything derived from it -- the weights passed into PyMARE, the dof map written back -- keeps describing the fit that actually happened, which is also why this does not belong in PyMARE: PyMARE would have to return a record of what it dropped for NiMARE to re-derive both. Its guard stays as a backstop that should now never fire.

Two triggers, on two different correlations. Cancellation is judged on the estimated null correlation, because that is the matrix PyMARE aggregates the group with. A pair of images that are exact mirrors but do not cancel -- other members keep the block sum off zero -- is judged on the empirical correlation and only warned about, loudly, with the data kept: the null correlation cannot see it, since a real mirror pair sharing signal estimates near -0.94 while correlating exactly -1. Anything short of -1 that does not cancel stays silent; where to draw a line short of exact cancellation is a statistical policy only the caller can set.

Applies to Fishers as well as Stouffers, which is a judgement call worth naming: Fishers never raised here, because Brown's method rescales a cancelled block rather than dividing by it. It rescaled it as though it carried information, though, so the same exclusion is at worst neutral there and at best removes an inflated contribution -- and it keeps the verdict on a group a property of the data rather than of which combination test was chosen.

Closes # .

-
-

## Summary by Sourcery

Exclude dependence groups whose estimated correlations cancel before combination testing so affected image-based meta-analyses continue with accurate per-bag results and diagnostics.

Bug Fixes:
- Prevent cancelling dependence groups from terminating image-based meta-analyses by excluding the affected group from each model or voxel bag before combination testing.
- Apply cancellation handling to both Stouffer and Fisher combination estimators while preserving valid data outside affected bags.

Enhancements:
- Add per-group screening based on estimated null-correlation variance and warnings for exact empirical sign-flipped image pairs that are retained.
- Track filtered groups when calculating fitted weights and degrees of freedom, and report affected studies, images, models, and voxels.

Documentation:
- Document that cancelling dependence groups are excluded locally with a warning rather than ending the meta-analysis.

Tests:
- Add coverage for per-bag cancellation, whole-group three-way cancellation, retained mirror-image warnings, near-cancellation silence, and alignment with PyMARE's cancellation threshold.

### DIFF
--- a/nimare/meta/_dependence.py
+++ b/nimare/meta/_dependence.py
@@ -106,6 +106,24 @@ class DependenceModel:
         """Number of independent units backing the inference."""
         return int(self.group_order.size)
 
+    def group_rows(self):
+        """Yield ``(label, rows)`` for every group, in :attr:`group_order`.
+
+        The enumeration PyMARE's combination tests run over the labels they are handed, so a
+        per-group quantity computed here describes the block PyMARE will aggregate.
+
+        Yields
+        ------
+        label : :obj:`object`
+            The block label.
+        rows : :obj:`numpy.ndarray`
+            Row indices of that block's images, into this grouping rather than the full
+            dataset; :attr:`image_indices` translates them back.
+        """
+        codes, labels = self._encoded_blocks
+        for index, label in enumerate(labels):
+            yield label, np.flatnonzero(codes == index)
+
     @property
     def supports_inference(self):
         """Whether there are at least two independent units to compare.
--- a/nimare/meta/ibma.py
+++ b/nimare/meta/ibma.py
@@ -113,6 +113,41 @@ _REMOVED_PARAMETERS = {
 }
 
 
+#: Variance at or below which a dependence group's aggregated z statistic carries no
+#: information, because its images cancel. Matches the floor PyMARE's combination tests
+#: refuse to aggregate below, so a group is excluded here before PyMARE raises on it. Not a
+#: statistical threshold: it screens out exact cancellation and nothing else.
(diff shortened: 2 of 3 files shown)


jdkent (commented):
LGTM, this was a common case with julio's work

PR #1111, opened 2026-08-23
Let MetaResult.save leave the input images out of the file

jdkent:
For an image-based meta-analysis, estimator.inputs_ is most of a pickled MetaResult: on twenty maps over the 2mm MNI mask it is 29.3 MB of a 36.5 MB pickle (80%), and 88% of the gzipped file, because the input maps are stored once as z_maps and again as the per-bag copies _preprocess_input derives from them. All of it is a second copy of NIfTIs that are still on disk.

It cannot come out by default, and it cannot come out via __getstate__. Loading a result and reading its inputs back is something three supported operations do: run_reports plots the input maps and their correlation matrix, PermutedOLS.correct_fwe_montecarlo refits from inputs_["beta_maps"], and the diagnostics enumerate studies from inputs_["id"]. And pickling is how NiMARE deep-copies an estimator -- both MetaResult.__init__ and MetaResult.copy go through copy.deepcopy -- so an estimator __getstate__ that dropped the inputs would break the first Corrector.transform, not just a later load.

So this is an opt-in on save, and it drops as little as it can. The keys emptied are the ones the estimator declares as images in _required_inputs plus the data_bags derived from them; the ids, sample sizes, contrast labels, correlation matrix and aggressive mask stay, since they are small and the jackknife needs the ids to enumerate the studies it refits from disk. A result saved this way still serves get_map, save_maps, save_tables, copy, Bonferroni and FDR correction, and Jackknife and FocusCounter, which re-read the images themselves. It is 17% of the full file on the pain test set and 12% at 2mm.

What it no longer serves is reporting and Monte Carlo FWE correction. Rather than let those fail with a KeyError several frames in, the dropped values are replaced by a DroppedInput placeholder that raises on any read and names the key and the flag that removed it. A coordinate-based estimator declares no image inputs, so nothing is dropped from one.

Closes # .

-
-

## Summary by Sourcery

Allow MetaResult files to be saved without image inputs while clearly limiting operations that require those images.

New Features:
- Add an opt-in `with_inputs=False` save mode that produces smaller MetaResult files by omitting image-based estimator inputs while preserving other metadata and outputs.

Bug Fixes:
- Replace omitted inputs with explicit placeholders that provide actionable errors for reporting and Monte Carlo FWE correction instead of delayed missing-key failures.

Enhancements:
- Centralize image-input discovery across estimators, preprocessing, workflows, and result serialization.
- Preserve supported map, table, correction, copying, and diagnostic operations for results saved without input images.

Documentation:
- Document the new save option, retained capabilities, and operations unavailable when image inputs are omitted.

Tests:
- Add coverage for full and reduced save round trips, supported operations, informative failures, and unchanged coordinate-based results.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -42,6 +42,7 @@ For more information about the components of coordinate-based meta-analysis in N
    :template: class.rst
 
    results.MetaResult
+   results.DroppedInput
 
 
 .. _api_correct_ref:
--- a/nimare/base.py
+++ b/nimare/base.py
@@ -162,6 +162,18 @@ class NiMAREBase(CacheMixin, metaclass=ABCMeta):
     #: written into ``inputs_``; values are ``(kind, field)`` pairs.
     _required_inputs = {}
 
+    def _image_input_fields(self):
+        """Return the ``_required_inputs`` entries declared as images.
+
+        Returns
+        -------
+        :obj:`dict`
+            ``{inputs_ key: image field}``, in ``_required_inputs`` order.
+        """
+        return {
+            name: field for name, (kind, field) in self._required_inputs.items() if kind == "image"
+        }
+
     def _collect_inputs(self, dataset, drop_invalid=True):
         """Search for, and validate, required inputs as necessary.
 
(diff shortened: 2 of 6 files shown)


jdkent (commented):
LGTM

PR #1112, opened 2026-08-23
Refuse an image-based meta-analysis of fewer than two analyses

jdkent:
Every IBMA estimator needs at least two independent units, so a fit over a single analysis returns maps that are NaN at every voxel. Nothing said so. `StudysetView.resolve` guards zero rows -- "The collection has no data for 'z_maps'" -- but nothing guarded one, and the two masking strategies both produce the NaN quietly: `aggressive_mask=True` fits one model that `DependenceModel.supports_inference` rejects, and `aggressive_mask=False` has `_liberal_mask_bags` drop every bag for holding fewer than two studies, which leaves no skipped model for `_fit_over_bags` to warn about. The caller gets a MetaResult, writes it out, and finds out later.

This came out of a Neurostore studyset of 12 analyses that lost 11 of them to missing required map types. The one that survived fit, and the answer the caller needed was "your studyset lost 11 of 12 analyses", not "the maps are NaN" -- so the message carries both counts, and names the required inputs when analyses were dropped for want of them.

The floor is on the IBMA path only. A single-experiment CBMA produces a real map rather than NaN -- ALE, MKDADensity and KDA all return finite z over one study's foci -- so there is no evidence for the same floor in `Estimator._collect_inputs`, and it would be a stricter change.

Checked before raising, so no image is loaded first: too few analyses is a property of the collection, not of what the files contain.

Closes # .

-
-

## Summary by Sourcery

Require sufficient independent analyses for IBMA fitting and resampling operations to fail early instead of producing invalid results.

New Features:
- Reject image-based meta-analyses that have fewer than two usable analyses, with errors identifying dropped analyses and missing required inputs.
- Expose minimum-analysis requirements to diagnostics so jackknife and resampling workflows reject insufficient replicates.

Bug Fixes:
- Prevent IBMA fits with a single analysis from silently returning all-NaN maps.

Enhancements:
- Keep CBMA estimators exempt from the minimum-analysis check while enforcing estimator-specific floors for IBMA diagnostics.

Documentation:
- Document the new minimum-analysis behavior for IBMA fitting, jackknife, and resampled stability diagnostics.

Tests:
- Add coverage for insufficient IBMA fits, dropped required inputs, valid two-analysis fits, jackknife requirements, resampling floors, and unaffected CBMA jackknives.

### DIFF
--- a/nimare/diagnostics.py
+++ b/nimare/diagnostics.py
@@ -44,6 +44,29 @@ POSTAIL_LBL = "PositiveTail"  # Label assigned to positive tail clusters
 NEGTAIL_LBL = "NegativeTail"  # Label assigned to negative tail clusters
 
 
+def _estimator_min_analyses(estimator):
+    """Return the fewest analyses ``estimator`` will fit, or None if it declares no floor.
+
+    Read off the estimator so a diagnostic refitting over subsets cannot drift from it.
+    CBMA estimators declare none, which turns off every check that consults this.
+    """
+    floor = getattr(estimator, "_min_analyses", None)
+    return None if floor is None else int(floor)
+
+
+def _refit_group_sizes(estimator):
+    """Return the size of each set of analyses a diagnostic would refit over.
+
+    One entry for a single-sample estimator, two for a pairwise one, which renames
+    ``inputs_["id"]`` to ``id1`` and ``id2`` as it collects each group.
+    """
+    return [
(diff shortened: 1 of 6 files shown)


jdkent (commented):
LGTM, it's difficult to prospect how many analyses will be available per bag, so some bags may fail, but others will succeed.

PR #1113, opened 2026-08-23
Raise on an id that Studyset.slice cannot resolve

jdkent:
slice() returned an empty studyset for an id that named nothing, so a typo, an id from a different studyset and an id already sliced away were all indistinguishable from "these analyses genuinely have no data". The partially-resolvable case was worse than the wholly unresolvable one: a list where one id is wrong came back with a plausible-looking subset and nothing said an analysis had gone missing. Selecting is a request for the analyses named, so any id naming none of them now raises ValueError, and the error names the ids that failed rather than just the count.

Resolution is checked against the studyset in hand, not the store behind it, so slicing a sub-studyset with an id belonging to a sibling raises too. Excluding is left alone -- exclude_study_ids() and View.drop_keys() still ignore an id that matches nothing, because what they were asked to remove is absent from the result either way.

Two ways the short analysis id was unreachable, both from deriving it as full_id.rsplit("-", 1)[-1] instead of reading the analysis_key the store already carries:

- a hyphenated analysis id: "mixed-a-b" derived to "b", so the analysis could not be selected by the id it declares, and the short ids handed back by get_analyses_by_label/_mask/_coordinate could not be fed back into slice(). Left unfixed, raising would have turned a silent miss into a confident error on an id that is genuinely there.
- the positional ids generated for a document that declares none: "analysis-0" derived to "0".

Resolving a short id also kept only the first matching row. NIMADS does not require an analysis id to be unique across studies -- the bundled pain studyset calls every analysis "1" -- so selecting on one silently dropped every match after the first, and an accurate report of which ids resolved needs all of them anyway. Range lookup replaces the first-match one, matching the stance already documented on note_row_resolver.

Closes # .

-
-

## Summary by Sourcery

Enforce accurate, studyset-local identifier resolution when selecting analyses and studies.

New Features:
- Make analysis and study selection reject identifiers that do not resolve within the current Studyset, with errors identifying missing ids.
- Support selecting all analyses that share a short analysis id, including hyphenated ids and ids returned by query helpers.

Bug Fixes:
- Correct short-id resolution for hyphenated analysis identifiers and documents without declared analysis ids.
- Prevent silently dropping duplicate short-id matches during selection.

Enhancements:
- Preserve permissive behavior for exclusion operations while improving missing-name formatting and lookup utilities.

Documentation:
- Document strict selection behavior, supported identifier forms, duplicate short-id matching, and exclusion semantics.

Tests:
- Add coverage for unresolved analysis and study ids, sliced-away ids, long missing-id lists, hyphenated ids, and repeated short ids.

### DIFF
--- a/nimare/studyset/studyset.py
+++ b/nimare/studyset/studyset.py
@@ -19,7 +19,7 @@ import pandas as pd
 
 from nimare.studyset import blocks as _blocks
 from nimare.studyset import edit, requirements
-from nimare.studyset.columns import ID_COLS
+from nimare.studyset.columns import ID_COLS, join_names, missing_names
 from nimare.studyset.io import from_nimads, from_parquet, to_nimads_dict, write_nimads
 from nimare.studyset.io.parquet import write_parquet
 from nimare.studyset.layout import harmonize_space
@@ -394,10 +394,9 @@ class Studyset:
 
     def _label_block_for(self, labels, annotation=None):
         block = _blocks.label_block_for(self._view, annotation)
-        known = set(block.labels.tolist())
-        missing = [label for label in labels if label not in known]
+        missing = missing_names(labels, set(block.labels.tolist()))
         if missing:
-            raise ValueError(f"Missing label(s): {', '.join(map(str, missing))}")
+            raise ValueError(f"Missing label(s): {join_names(missing)}")
         return block
 
     def get_labels(self, ids=None):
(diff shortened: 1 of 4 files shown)


nimare/studyset/studyset.py
```diff
@@ -376,31 +412,32 @@ def get_labels(self, ids=None):
...
-            str(i).rsplit("-", 1)[-1]
-            for i in self.get_studies_by_label(labels, label_threshold, annotation)
-        ]
+        keep = self._label_masks(labels, label_threshold, annotation).all(axis=0)
+        return list(self._view.short_keys[keep].astype(str))
 
-    def get_studies_by_mask(self, mask):
-        """Full analysis ids with at least one focus inside ``mask``."""
+    def _analyses_in_mask(self, mask):
+        """This selection, narrowed to analyses with at least one focus in ``mask``."""
```
jdkent:
```suggestion
        """Studies filtered with at least one focus in ``mask``."""
```

jdkent (commented):
an edit to pass linting

jdkent (commented):
LGTM!

PR #1121, opened 2026-08-25
Compute the CBMR observed information in closed form

jdkent:
information_matrix differentiates the log-likelihood twice with torch.func.hessian, which is jacfwd(jacrev(.)). Forward mode carries one tangent per parameter through the likelihood at once, so the intermediate log-intensity tensor has shape (n_parameters, n_patterns, n_voxels). On a 21-group diagnosis model over 21,789 voxels that is 28.7 GB: the model cannot be fitted at all, and the size is driven by the group count, so a coarser grid does not avoid it.

All three distributions admit a closed form, and it is the same shape in each case. The predictor is linear in the coefficients, so the chain rule contributes no curvature term and the spatial block is sum_p L_pc L_pc' (B^T Sigma_p B), with Sigma_p the second differential in the log-intensity. Only Sigma_p differs: diagonal for Poisson and NegativeBinomial, diagonal plus rank one for ClusteredNegativeBinomial, whose spatial coefficients enter only through the scalar sum_v exp(S_pv). Each is n_patterns rank-updates of width n_bases, so nothing of size n_parameters is formed. Automatic differentiation stays as the fallback for a distribution added without a derivation.

That exposes the next bottleneck: covariance() spent 163 s on an SVD for a diagnostic condition number against 14 s for the inverse it was diagnosing. The same L_pc L_pc' factor says which coefficients can covary, independently of the distribution, so the information matrix is exactly block diagonal for a cell-means factor and bordered block diagonal once moderators are added. covariance() now dispatches over that structure, and takes the condition number from a symmetric eigensolver rather than an SVD -- exact, since a symmetric matrix's singular values are its absolute eigenvalues. Every rung returns the same matrix and reports the same condition number; only the work differs.

marginal_by_pattern keeps a sparse foci sparse, since densifying it inside the routine meant to avoid large allocations would defeat the purpose.

Measured on the real 21-group pool, complete run including 20 one-versus-rest contrasts: 769 s and 5.49 GB peak before, 44 s and 2.67 GB after.

Every identity is proved symbolically at https://github.com/jdkent/cbmr-proofs (59 claims, each a residual simplified to exactly zero) and checked numerically against torch.func.hessian in the new tests.

Closes # .

-
-

## Summary by Sourcery

Replace CBMR Hessian-based information and dense covariance workflows with closed-form, structure-aware computations that substantially reduce runtime and peak memory usage.

New Features:
- Compute observed Fisher information in closed form for Poisson, Negative Binomial, and Clustered Negative Binomial CBMR models, while retaining automatic differentiation as a fallback for unsupported distributions.
- Exploit block and bordered-block structure to accelerate Fisher covariance inversion and condition-number evaluation.
- Cache covariance matrices across repeated hypothesis tests for a fitted model.
- Preserve sparse foci inputs during pattern marginalization and experiment-total calculations.

Bug Fixes:
- Avoid memory-intensive Hessian intermediates and dense sparse-foci allocations that could prevent large CBMR models from fitting.

Enhancements:
- Add symbolic and numerical validation of information derivations, covariance routes, sparse/dense input equivalence, singular designs, and cache behavior.

CI:
- Relax benchmark regression detection to flag runs that are at least 20% slower than the baseline.

Tests:
- Add comprehensive tests comparing closed-form information with automatic differentiation across distributions and model designs.

Chores:
- Update CBMR benchmarks to reuse cached setup data and remove the obsolete diagnostics benchmark.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -20,4 +20,7 @@ jobs:
       - run: git fetch origin $GITHUB_BASE_REF:base $GITHUB_REF:pr
       - run: pip install asv virtualenv
       - run: asv machine --yes
-      - run: asv continuous base pr
+      # 1.2 rather than asv's default 1.1. The CBMR benchmarks run in single-digit
+      # milliseconds, and run-to-run spread on a shared runner is wider than 10% at that
+      # size, so 1.1 reports noise as a regression.
+      - run: asv continuous --factor 1.2 base pr
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -20,7 +20,5 @@ jobs:
       - run: git fetch origin $GITHUB_BASE_REF:base $GITHUB_REF:pr
       - run: pip install asv virtualenv
       - run: asv machine --yes
-      # 1.2 rather than asv's default 1.1. The CBMR benchmarks run in single-digit
-      # milliseconds, and run-to-run spread on a shared runner is wider than 10% at that
-      # size, so 1.1 reports noise as a regression.
+      # 1.2 means benches must be 20% slower to be flagged.
       - run: asv continuous --factor 1.2 base pr
--- a/benchmarks/bench_cbma.py
+++ b/benchmarks/bench_cbma.py
@@ -13,12 +13,8 @@ class TimeCBMA:
     """Time CBMA estimators."""
 
     def setup_cache(self):
-        """Build the datasets once per process, not once per timed sample.
+        """Build the datasets once per process.
 
-        asv re-runs ``setup`` before every sample. Building these costs seconds while the
-        estimators they feed run in tens of milliseconds, so setup dominated the suite's wall
-        clock without appearing in any reported number. The fixtures pickle to about 4.5 MB and
-        load in a few milliseconds, and fitting does not modify them.
         """
         dataset = nimare.dataset.Dataset(
             os.path.join(get_test_data_path(), "test_pain_dataset.json")
--- a/benchmarks/bench_cbma.py
+++ b/benchmarks/bench_cbma.py
@@ -12,25 +12,38 @@ from nimare.tests.utils import get_test_data_path
 class TimeCBMA:
     """Time CBMA estimators."""
 
-    def setup(self):
-        """
-        Setup the data.
+    def setup_cache(self):
+        """Build the datasets once per process, not once per timed sample.
 
-        Loads the dataset required for the benchmarks.
+        asv re-runs ``setup`` before every sample. Building these costs seconds while the
+        estimators they feed run in tens of milliseconds, so setup dominated the suite's wall
+        clock without appearing in any reported number. The fixtures pickle to about 4.5 MB and
+        load in a few milliseconds, and fitting does not modify them.
         """
-        self.dataset = nimare.dataset.Dataset(
+        dataset = nimare.dataset.Dataset(
             os.path.join(get_test_data_path(), "test_pain_dataset.json")
         )
-        _, self.dataset_dense = create_coordinate_dataset(
+        _, dataset_dense = create_coordinate_dataset(
             foci=60,
             n_studies=100,
             foci_percentage="100%",
             seed=123,
         )
-        self.studyset = Studyset.from_dataset(self.dataset)
-        self.studyset_dense = Studyset.from_dataset(self.dataset_dense)
+        return {
+            "dataset": dataset,
+            "dataset_dense": dataset_dense,
+            "studyset": Studyset.from_dataset(dataset),
+            "studyset_dense": Studyset.from_dataset(dataset_dense),
+        }
 
-    def time_ale(self):
+    def setup(self, data):
+        """Take the datasets from the cache."""
+        self.dataset = data["dataset"]
+        self.dataset_dense = data["dataset_dense"]
+        self.studyset = data["studyset"]
+        self.studyset_dense = data["studyset_dense"]
+
+    def time_ale(self, data):
         """
         Time the ALE estimator.
 
@@ -39,7 +52,7 @@ class TimeCBMA:
         meta = ALE()
         meta.fit(self.dataset)
 
-    def time_mkdadensity(self):
+    def time_mkdadensity(self, data):
         """
         Time the MKDADensity estimator.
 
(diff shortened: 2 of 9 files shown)


nimare/meta/cbmr/information.py
```diff
@@ -0,0 +1,310 @@
...
+    H_bb[(c,k),(c',k')] = sum_p L_pc L_pc' (B^T Sigma_p B)[k,k']
+
+Only ``Sigma_p``, the second derivative of the pattern's term in the log-intensity, differs
+between them. Each is ``n_patterns`` rank-updates of width ``n_bases``, so nothing of size
+``n_parameters`` is built.
+
+Notes
+-----
+Every identity here is derived and proved at https://github.com/jdkent/cbmr-proofs, and checked
+against ``torch.func.hessian`` in ``nimare/tests/test_meta_cbmr_information.py``.
```
jdkent:
```suggestion
```

nimare/meta/cbmr/information.py
```diff
@@ -0,0 +1,310 @@
...
+
+from nimare.meta.cbmr.distributions import (
+    ClusteredNegativeBinomial,
+    NegativeBinomial,
+    Poisson,
+)
+
+
+def _fitted_pieces(model):
+    """Return the fitted quantities all three closed forms share."""
```
jdkent:
```suggestion
    """Return the fitted quantities all closed forms share."""
```

nimare/meta/cbmr/information.py
```diff
@@ -0,0 +1,310 @@
...
+def poisson_information_matrix(model, foci=None):
+    """Return the observed Fisher information of a fitted Poisson model.
+
+    Parameters
+    ----------
+    model : :class:`~nimare.meta.cbmr.model.CBMRModel`
+        Fitted model.
+    foci : optional
+        Ignored. Under a log link the Poisson information depends only on the fitted
+        coefficients, so the counts drop out.
```
jdkent: should the foci argument be deleted if it's ignored? (it does not need to be kept for backward compatibility)

jdkent (commented):
there may be some lingering logic, deadends for code.

.github/workflows/benchmark.yml
```diff
@@ -20,4 +20,7 @@ jobs:
       - run: git fetch origin $GITHUB_BASE_REF:base $GITHUB_REF:pr
       - run: pip install asv virtualenv
       - run: asv machine --yes
-      - run: asv continuous base pr
+      # 1.2 rather than asv's default 1.1. The CBMR benchmarks run in single-digit
+      # milliseconds, and run-to-run spread on a shared runner is wider than 10% at that
+      # size, so 1.1 reports noise as a regression.
```
jdkent:
```suggestion
      # 1.2 means benches must be 20% slower to be flagged.
```

benchmarks/bench_cbma.py
```diff
@@ -12,25 +12,38 @@
 class TimeCBMA:
     """Time CBMA estimators."""
 
-    def setup(self):
-        """
-        Setup the data.
+    def setup_cache(self):
+        """Build the datasets once per process, not once per timed sample.
```
jdkent:
```suggestion
        """Build the datasets once per process.
```

benchmarks/bench_cbma.py
```diff
@@ -12,25 +12,38 @@
...
 
-        Loads the dataset required for the benchmarks.
+        asv re-runs ``setup`` before every sample. Building these costs seconds while the
+        estimators they feed run in tens of milliseconds, so setup dominated the suite's wall
+        clock without appearing in any reported number. The fixtures pickle to about 4.5 MB and
+        load in a few milliseconds, and fitting does not modify them.
```
jdkent:
```suggestion
```

benchmarks/benchmark_cbmr.py
```diff
@@ -119,9 +119,19 @@ def _legacy_terms(formula):
 class _CBMRBenchmarkMixin:
     """Shared setup for CBMR benchmarks."""
 
-    def setup(self):
-        """Simulate the data the benchmarks fit."""
-        self.studyset = _make_studyset()
+    def setup_cache(self):
+        """Simulate the studyset once per process, not once per timed sample.
```
jdkent:
```suggestion
        """Simulate the studyset once per process.
```

benchmarks/benchmark_cbmr.py
```diff
@@ -119,9 +119,19 @@ def _legacy_terms(formula):
...
+
+        asv re-runs ``setup`` before every sample, and simulating the studyset costs seconds
+        while the benchmarks it feeds run in milliseconds. ``setup_cache`` runs once and its
+        result is pickled and handed to each sample, which is about 1.9 MB and a hundredth of a
+        second here.
```
jdkent:
```suggestion
```

nimare/meta/cbmr/model.py
```diff
@@ -135,7 +139,17 @@ def information_matrix(self, foci):
         One matrix over all coefficients, so the cross blocks between terms are present rather
         than assumed away. Nuisance parameters are held fixed at their fitted values, matching
         how CBMR has always reported regression standard errors.
+
+        Computed in closed form for every distribution NiMARE ships; see
+        :mod:`nimare.meta.cbmr.information`. Automatic differentiation is the fallback for one
+        added without a derivation.
```
jdkent:
```suggestion
       than assumed away. Nuisance parameters are held fixed at their fitted values.

        Computed in closed form for every distribution with automatic differentiation as the fallback for
        distributions added without a derivation.
```

nimare/meta/cbmr/model.py
```diff
@@ -161,35 +175,35 @@ def covariance(self, foci, cov_type="fisher", meat="cluster", correction="hc1",
...
+
+        Notes
+        -----
+        The result is cached. It is a function of the converged coefficients and the foci, so
+        every hypothesis tested against one fit asks for the same matrix, and
+        :meth:`~nimare.meta.cbmr.results.CBMRResult.test` would otherwise rebuild it each time.
+        The cache holds one matrix, is keyed on the foci and the covariance options, and is
+        cleared by :meth:`fit`. At many groups that matrix is large -- 0.64 GB at 8,947
+        coefficients -- but it is shared by every result derived from the fit, since
+        ``CBMRResult.copy`` passes the estimator by reference.
```
jdkent:
```suggestion
        cleared by :meth:`fit`. At many groups that matrix is large; could be well over 500mb, but it is shared by every result derived from the fit, since
        ``CBMRResult.copy`` passes the estimator by reference.
```

jdkent (commented):
some wordings need to be fixed.

jdkent:
Since this pull request means the difference between being able to run CBMR on my laptop versus not, and that it's numerically equivalent with the old version for the models that were tested, and there is some evidence that the algebra was worked through to be equivalent, I'm going to merge this in.

(the benchmark has some noisy measures; could be due to import time)

PR #1126, opened 2026-08-27
Store foci on CBMRModel

### LINKED ISSUE #1118: add foci as an attribute to CBMRModel
reading this and covariance, I'm becoming convinced that foci should be an attribute on CBMRModel instead of being an argument passed around. the foci should never be different than what was fit with the estimator so there is one, and only one, correct value for the argument foci.

_Originally posted by @jdkent in https://github.com/neurostuff/NiMARE/pull/1080#discussion_r3847068017_

yifan0330:
## Summary
- store the fitted foci matrix on CBMRModel when fit is called
- remove repeated foci arguments from model likelihood, information, covariance, standard-error, and contrast paths
- update CBMR tests for the stored-foci API and covariance cache behavior

## Tests
- python -m pytest nimare/tests/test_meta_cbmr_model.py -q
- python -m pytest nimare/tests/test_meta_cbmr_covariance.py -q
- python -m pytest nimare/tests/test_meta_cbmr_information.py -q
- python -m pytest nimare/tests/test_meta_cbmr_contrasts.py -q
- python -m pytest nimare/tests/test_meta_cbmr_exposure.py nimare/tests/test_meta_cbmr_formula.py -q
- git diff --check

Closes neurostuff/NiMARE#1118

## Summary by Sourcery

Make CBMRModel retain the foci used for fitting and use that state throughout fitted-model inference.

New Features:
- Store an immutable copy of the fitted foci on CBMRModel for reuse by downstream fitted-model operations.

Bug Fixes:
- Preserve the previous fitted state, coefficients, foci, and covariance cache when a refit fails.

Enhancements:
- Simplify likelihood, information, covariance, standard-error, and contrast APIs by deriving foci from the fitted model.
- Tie covariance caching to covariance options and invalidate it only after a successful refit.

Tests:
- Update CBMR model, information, covariance, contrast, exposure, and formula tests for the stored-foci API and refit/cache behavior.

### DIFF
--- a/nimare/meta/cbmr/contrasts.py
+++ b/nimare/meta/cbmr/contrasts.py
@@ -276,7 +276,7 @@ def _statistics(model, block, contrast, constants, covariance_block, bases):
 
 
 def evaluate_hypotheses(
-    model, hypotheses=None, foci=None, name=None, term=None, method=None, **covariance_kwargs
+    model, hypotheses=None, name=None, term=None, method=None, **covariance_kwargs
 ):
     """Test hypotheses about the fitted coefficients.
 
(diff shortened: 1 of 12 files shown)


jdkent (approved):
LGTM!

API looks cleaner without having to pass around the same foci from function to function.

Issue #912, opened 2025-02-19
Improve documentation on why the "generic" (non-Monte Carlo) MCC methods are suboptimal


tsalo:
The section on [multiple comparisons correction](https://nimare.readthedocs.io/en/stable/cbma.html#multiple-comparisons-correction) in the documentation basically just says "We do not recommend using the generic methods." It would be great to go into a bit more detail on why that is, even if the focus is on ALE (e.g., by citing [Eickhoff et al. (2016) 1](https://doi.org/10.1016/j.neuroimage.2016.04.072) and [Frahm et al. (2022)](https://doi.org/10.1002/hbm.25898)).

This issue is somewhat motivated by [this Neurostars post](https://neurostars.org/t/multiple-comparisons-correction-for-ale-subtraction-in-nimare/31789).

Achintyasingh412:
"Hi! I'd love to work on this documentation update. I'll open a PR soon!"

PR #1127, opened 2026-09-01
DOC: expand explanation on why generic MCC methods are suboptimal (Is…

Achintyasingh412:
Screenshot 2026-09-01 072018
Screenshot 2026-09-01 072018
…sue #912)

Closes # .

-
-

## Summary by Sourcery

Documentation:
- Expand the CBMA documentation with additional explanation of why generic multivariate coordinate-based meta-analysis methods may be suboptimal.

### DIFF
--- a/docs/cbma.rst
+++ b/docs/cbma.rst
@@ -94,7 +94,7 @@ One example of a generic method is the "bonferroni" method for the FWECorrector.
 Generic multiple-comparison correction methods fail to properly account for
 the spatial smoothness, correlation, and null distribution characteristics
 unique to coordinate-based meta-analyses like ALE. As outlined by
-Eickhoff et al. (2016) and Frahm et al. :footcite:p:`frahm2022evaluation`,
+Eickhoff et al. :footcite:p:`eickhoff2016behavior` and Frahm et al. :footcite:p:`frahm2022evaluation`,
 using these generic approaches results in suboptimal statistical control,
 making estimator-specific or permutation/Monte Carlo-based corrections
 necessary for valid inferences.
--- a/docs/cbma.rst
+++ b/docs/cbma.rst
@@ -94,9 +94,10 @@ One example of a generic method is the "bonferroni" method for the FWECorrector.
 Generic multiple-comparison correction methods fail to properly account for
 the spatial smoothness, correlation, and null distribution characteristics
 unique to coordinate-based meta-analyses like ALE. As outlined by
-Eickhoff et al. (2016) and Frahm et al. (2022), using these generic approaches
-results in suboptimal statistical control, making estimator-specific or
-permutation/Monte Carlo-based corrections necessary for valid inferences.
+Eickhoff et al. (2016) and Frahm et al. :footcite:p:`frahm2022evaluation`,
+using these generic approaches results in suboptimal statistical control,
+making estimator-specific or permutation/Monte Carlo-based corrections
+necessary for valid inferences.
 
 Estimator-specific methods are approaches that are implemented within the Estimator as class methods
 that are then called by the Corrector.
--- a/docs/cbma.rst
+++ b/docs/cbma.rst
@@ -91,6 +91,12 @@ One example of a generic method is the "bonferroni" method for the FWECorrector.
 
 .. tip::
     We do not recommend using the generic methods.
+Generic multiple-comparison correction methods fail to properly account for
+the spatial smoothness, correlation, and null distribution characteristics
+unique to coordinate-based meta-analyses like ALE. As outlined by
+Eickhoff et al. (2016) and Frahm et al. (2022), using these generic approaches
+results in suboptimal statistical control, making estimator-specific or
+permutation/Monte Carlo-based corrections necessary for valid inferences.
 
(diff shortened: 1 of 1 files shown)


docs/cbma.rst
jdkent, line 174: here is an example of a citation.

jdkent (changes requested):
Thanks for the contribution, since you are referencing papers, we can cite them using restructured text, you will need to either add/use the reference in this file: https://github.com/neurostuff/NiMARE/blob/main/nimare/resources/references.bib

and cite it like :footcite:p:`smith2009threshold` in the document.

docs/cbma.rst
Achintyasingh412: "Hi @jdkent, thank you for the helpful review and the citation example! I'll update the formatting right away and push the fix shortly."

Achintyasingh412:
Hi @jdkent, thank you for the guidance! I've updated the Frahm citation to use the :footcite:p: format as requested Please let me know if it looks good to go now.

docs/cbma.rst
```diff
@@ -91,6 +91,13 @@ One example of a generic method is the "bonferroni" method for the FWECorrector.
 
 .. tip::
     We do not recommend using the generic methods.
+Generic multiple-comparison correction methods fail to properly account for
+the spatial smoothness, correlation, and null distribution characteristics
+unique to coordinate-based meta-analyses like ALE. As outlined by
+Eickhoff et al. (2016) and Frahm et al. :footcite:p:`frahm2022evaluation`,
```
jdkent: please add the citation for Eickoff et al. 2016 as well.

jdkent (changes requested):
thanks for the change, please add a citeation for Eickhoff as well.

Achintyasingh412:
Hi @jdkent, thanks for catching that! I've added the Sphinx footcite tag for Eickhoff et al. (2016) right alongside Frahm. Please let me know if everything looks good to go now

Achintyasingh412:
Hey @jdkent, following up on your suggestion to integrate Jackknife directly into the Pair wise CBMA Workflow—I tested it out, but I'm running into a technical limitation when the HTML report gets generated.

During run_reports, it throws the following error:
*******************************************************************************
ValueError: The condensed distance matrix must contain only finite values
*******************************************************************************
This happens inside SciPy's linkage function when building the contribution heatmap. It appears that for a subtraction meta-analysis, certain studies dropped during the jackknife procedure result in non-finite values or edge-case matrices that break the hierarchical clustering reordering.

Since you've recently optimized Jackknife, I wanted to flag this error in case it's something we can handle upstream in the diagnostic or report generation logic.

In the meantime, what do you think we should do for this example? Would falling back to FocusCounter here be the right move, or is there another way you'd prefer to handle subtraction diagnostics while this is looked into?

Achintyasingh412:
Hey @jdkent, just dropping a friendly ping on this PR whenever you have a spare moment to review. Let me know if any adjustments or further clarifications are needed!

docs/cbma.rst
```diff
@@ -91,6 +91,13 @@ One example of a generic method is the "bonferroni" method for the FWECorrector.
 
 .. tip::
     We do not recommend using the generic methods.
+Generic multiple-comparison correction methods fail to properly account for
+the spatial smoothness, correlation, and null distribution characteristics
+unique to coordinate-based meta-analyses like ALE. As outlined by
+Eickhoff et al. :footcite:p:`eickhoff2016behavior` and Frahm et al. :footcite:p:`frahm2022evaluation`,
+using these generic approaches results in suboptimal statistical control,
+making estimator-specific or permutation/Monte Carlo-based corrections
+necessary for valid inferences.
```
jdkent, line 100: thanks for the formatting, sorry I haven't had time to review the content much, your addition repeats the same sentiment as the paragraph above the "tip", so I do not think you need to add another paragraph. Honestly, this whole section could be re-written, but that's out of scope.

What is comes down to is that the FWECorrector using method="montecarlo", is the correct choice for statistical inference, based on Eickhoff's work.  (at least for ALE kernels). 

There has not been a systematic comparison for MKDA as far as I know.

So my recommendation is to embellish this sentence:

```
Generic methods rely on internal implementations of common correction approaches to correct the results as an array,
without accounting for any of the idiosyncrasies of neuroimaging data, such as the smoothness of the data.
```

as opposed to adding a new paragraph. The estimater-specific or permutation/montecarlo based correction could be simplified to only mentioning montecarlo based corrector, since estimator-specific and permutation methods are just another way to refer to montecarlo corrections. Users do not need to know the implementational details of montecarlo corrections (since they are implemented per estimator); the main message is that montecarlo corrections are the most validated for ALE methods.

And it could be made clearer that every other method, e.g., "bonferroni and both FDR methods" do not take into account the idiosyncracies of neuroimaging data, or to state the other way, montecarlo correction is the only method to take into account neuroimaging idiosyncracies.

jdkent (changes requested):
thanks for the wording change, I left some feedback; it's been a while since this section's been touched, so even outside this pull request I would want to make changes.

Achintyasingh412:
I'll take a look at the documentation feedback as well, clean up that section around the Monte Carlo corrections, and get it sorted out soon.

PR #1129, opened 2026-09-02
DOC: Add robustness analysis 

### LINKED ISSUE #927: Missing robustness analysis code and plots in two-sample ALE example documentation
## Summary

The example script for the two-sample ALE meta-analysis ([CBMA subtraction & conjunction analysis](https://nimare.readthedocs.io/en/latest/auto_examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.html)) does not appear to include any code or plots demonstrating the robustness analysis (e.g., visualization of robust activation areas), even though such plots are shown in the blog article referenced.

## Additional details

- NiMARE version:v0.5.1

I attempted to replicate the robustness analysis plot shown in the blog-cited paper (Alexander Enge, Rasha Abdel Rahman, and Michael A Skeide. A meta-analysis of fMRI studies of semantic cognition in children. NeuroImage, 241:118436, 2021. URL: https://doi.org/10.1016/j.neuroimage.2021.118436, doi:10.1016/j.neuroimage.2021.118436), using coordinate-based data and the example subtraction analysis script from the official NiMARE documentation.

I expected the example script to include or reference the code that generated the robustness visualization — i.e., the plot showing brain regions with robustness above a specific threshold, similar to the figure shown in the paper.

### What actually happened?
The script executed successfully for the two-sample ALE and subtraction analyses, but it did not include any robustness evaluation or plotting code. The robustness results/visualizations shown in the blog post are not reproducible based on the current example script.

Achintyasingh412:
This is a suggested pull request template for NiMARE.
It's designed to capture information we've found to be useful in reviewing pull requests.

If there is other information that would be helpful to include, please don't hesitate to add it!

Please also label your pull request with the relevant tags.
See here for more information and a list of available options:
https://github.com/neurostuff/NiMARE/blob/main/CONTRIBUTING.md#pull-requests
-->

Closes #927 

-
-

## Summary by Sourcery

Document study-level robustness of subtraction meta-analysis results in the two-sample CBMA example.

Enhancements:
- Add a Jackknife robustness analysis and contribution table to the two-sample subtraction CBMA example.

Documentation:
- Document robustness analysis of subtraction results in the two-sample CBMA example.

### DIFF
--- a/examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.py
+++ b/examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.py
@@ -192,7 +192,31 @@ run_reports(res_sub, html_dir)
 #
 #     <iframe src="./08_subtraction/report.html" style="border:none;" seamless="seamless" \
 #        width="100%" height="1000px"></iframe>
+###############################################################################
+###############################################################################
+# Robustness Analysis of Subtraction Results
+# -----------------------------------------------------------------------------
+# To evaluate the stability and robustness of the subtraction clusters against
+# individual study dropouts, we can run a Jackknife diagnostic analysis
+# on the subtraction MetaResult.
+from nimare.diagnostics import Jackknife
 
+jackknife = Jackknife(
+    target_image="z_desc-group1MinusGroup2Size_level-cluster_corr-FWE_method-montecarlo",
+    voxel_thresh=None,
+    display_second_group=True,
+)
+sub_diagnostic_results = jackknife.transform(res_sub)
+
+###############################################################################
+# Subtraction Jackknife Table
+# -----------------------------------------------------------------------------
+# We can inspect the contribution table to see how individual studies from
+# both groups influence the resulting subtraction clusters.
+sub_jackknife_table = sub_diagnostic_results.tables[
+    "z_desc-group1MinusGroup2Size_level-cluster_corr-FWE_method-montecarlo_diag-Jackknife_tab-counts_tail-positive"
+]
+sub_jackknife_table.head(10)
 ###############################################################################
 # Conjunction analysis
 # -----------------------------------------------------------------------------


Achintyasingh412:
Hi ,

This PR resolves Issue #927 by integrating the missing Jackknife robustness analysis directly into the two-sample ALE subtraction example script. 

To achieve this, we added the `Jackknife` diagnostic configured with the correct subtraction map name (`z_desc-group1MinusGroup2Size_level-cluster_corr-FWE_method-montecarlo`) to evaluate study stability across both groups. 

We thoroughly verified the workflow by executing the script locally from top to bottom, confirming that all Monte Carlo permutations, diagnostic tables, and conjunction plots complete successfully without errors. 

You will notice we did not add a new external citation to the bibliography at the bottom; this is because `Jackknife` is already a core, built-in diagnostic class in NiMARE rather than introducing a brand-new external paper. 

Looking forward to your feedback!

examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.py
jdkent, line 174: this is where the diagnostic is specified, we historically did not not run jackknife in the example because it took so long, I've since made some optimizations, so perhaps jackknife could be added here, but perhaps a note signalling jacnkknife could be used here would suffice.

jdkent (changes requested):
thanks @Achintyasingh412! please open/reference an existing issue before making a pull request. That allows us get alinged on what the issue is exactly before we try to fix it. That way we don't spend extra effort where it's not necessary. For example, I think this pull request could just be a comment on the relevant line that specifies the diagnostic, not a new section in this notebook.

Achintyasingh412:
Hi @jdkent, thank you for the guidance! I completely understand about linking the issue first—I'll make sure to do that for all future PRs. Regarding the diagnostic, thanks for pointing out that Pairwise CBMA Workflow handles it via the diagnostics parameter. Should I update the workflow to use Jackknife directly there now that it's optimized, or would you prefer a comment/note indicating where it can be swapped?

jdkent:
responding to this comment in what I assume to be relevant to this issue: https://github.com/neurostuff/NiMARE/pull/1127#issuecomment-5515225548

I would keep the diagnostic as focuscounter, there may be some cases that I don't cover with jackknife when the number of studies is low (like in this example), if you don't mind opening an issue, then someone (you if you want) can work on characterizing how that error arises and what a better guard/work around/error message would be. Thanks for looking into this!

Achintyasingh412:
Done! Opened tracking issue #1130 to document the Jackknife distance matrix error during subtraction report generation. Thanks again for pointing it out

PR #1131, opened 2026-09-04
Fix Jackknife non-finite values in pairwise subtraction and add test …

### LINKED ISSUE #1130: Bug: Jackknife diagnostic raises non-finite distance matrix error during subtraction meta-analysis report generation
## Summary
Jackknife diagnostic raises a ValueError ("The condensed distance matrix must contain only finite values") during HTML report generation when integrated into a two-sample ALE subtraction workflow.

## Additional details

- NiMARE version: Main / Latest

I was trying to integrate the `Jackknife` diagnostic directly into a `PairwiseCBMAWorkflow` for a two-sample ALE subtraction analysis and generate an HTML report using `run_reports`.

I expected the workflow and report generation to complete successfully, computing leave-one-out study contributions for the subtraction results.

### What actually happened?
The meta-analysis and workflow executed, but `run_reports()` crashed at the final step during hierarchical clustering with the following traceback:

This appears to happen because certain studies dropped during the jackknife procedure in a subtraction setup produce non-finite values or edge-case matrices that break SciPy's `linkage` function (especially when study counts are low).

## Reproducing the bug

<!--Please include any code snippets.
Enclose them in triple back-ticks (```)
Like this:
from nimare.workflows import PairwiseCBMAWorkflow
from nimare.diagnostics import Jackknife
from nimare.reports.base import run_reports

workflow = PairwiseCBMAWorkflow(
    estimator=subtraction_estimator,
    corrector=subtraction_corrector,
    diagnostics=Jackknife(
        target_image="z_desc-group1MinusGroup2Size_level-cluster_corr-FWE_method-montecarlo",
        voxel_thresh=None,
        display_second_group=True,
    ),
    voxel_thresh=None,
)
res_sub = workflow.fit(knowledge_studyset, related_studyset)

run_reports(res_sub, html_dir)

Achintyasingh412:
Closes #1130.

Added a robust guardrail to the Jackknife diagnostic's cluster summarization function (_summarize_cluster_values) to detect and sanitize non-finite (NaN and infinite) values before they reach SciPy's hierarchical clustering functions.

Implemented a safe fallback mechanism using np.nanmedian and np.nan_to_num to prevent ValueError crashes during HTML report generation in two-sample subtraction workflows.

Added comprehensive unit test coverage (test_jackknife_non_finite_guardrail) in test_diagnostics.py to prevent regressions.

Verified end-to-end functionality using validation scripts (arun_test.py and ana_test.py), confirming that PairwiseCBMAWorkflow, ALESubtraction, and run_reports() complete successfully without errors.

## Summary by Sourcery

Harden Jackknife diagnostics against non-finite cluster contributions in pairwise subtraction workflows.

Bug Fixes:
- Prevent Jackknife cluster summarization from failing when pairwise subtraction produces NaN or infinite values.
- Update the pairwise subtraction example to use Jackknife diagnostics with the subtraction target image.

Enhancements:
- Add a safe fallback that replaces non-finite cluster contributions with finite baseline values before hierarchical clustering.

Tests:
- Add regression coverage confirming non-finite Jackknife cluster values are sanitized successfully.

### DIFF
--- a/examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.py
+++ b/examples/02_meta-analyses/08_plot_cbma_subtraction_conjunction.py
@@ -171,7 +171,12 @@ subtraction_corrector = FWECorrector(
 workflow = PairwiseCBMAWorkflow(
     estimator=subtraction_estimator,
     corrector=subtraction_corrector,
-    diagnostics=FocusCounter(voxel_thresh=0.01, display_second_group=True),
+    diagnostics=Jackknife(
+        target_image="z_desc-group1MinusGroup2Size_level-cluster_corr-FWE_method-montecarlo",
+        voxel_thresh=None,
+        display_second_group=True,
+    ),
+    voxel_thresh=None,
 )
 res_sub = workflow.fit(knowledge_studyset, related_studyset)
 
--- a/nimare/diagnostics.py
+++ b/nimare/diagnostics.py
@@ -399,17 +399,29 @@ def _build_cluster_summary_context(masker, label_map, label_vector, cluster_ids)
 def _summarize_cluster_values(values, masker, cluster_summary_context):
     """Reduce per-feature values to cluster-level means."""
     if cluster_summary_context["mode"] == "masked_array":
-        return np.array(
+        raw_means = np.array(
             [
-                np.mean(values[cluster_idx])
+                np.mean(values[cluster_idx]) if cluster_idx.size > 0 else np.nan
                 for cluster_idx in cluster_summary_context["cluster_indices"]
             ],
             dtype=DEFAULT_FLOAT_DTYPE,
         )
+    else:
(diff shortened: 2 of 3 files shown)


jdkent (changes requested):
Thanks for the fix! This fixes a symptom of the problem, I want to know how the nans are being generated in the first place, and have the values be appropriately set there (check the `_transform` function in `Jackknife`, that should be place where the values are getting wonky).

In general, we will not impute median/mean values in NiMARE; if we don't have a usable number, we will drop it from the computation or fix it.

Achintyasingh412:
Thanks for the feedback! I completely agree that addressing the root cause in _transform and avoiding imputation is the right approach. I will look into how the non-finite values are being generated there, drop or fix them properly, and update the PR as soon as possible. @jdkent

PR #1135, opened 2026-09-08
Reject coordinates outside mask image bounds

### LINKED ISSUE #1134: Mask queries can include coordinates outside the image volume
## Summary

Studyset mask queries can treat a coordinate outside the image volume as inside the mask. `View.points_in_mask()` clips voxel indices to the image boundaries before looking up mask values. An out of bounds coordinate is therefore evaluated at the nearest boundary voxel.

## Additional details

NiMARE version: current `main` at commit `872486e`

Identify analyses with coordinates inside a mask.

Coordinates outside the mask image volume should return `False` and should not cause their analyses to appear in mask query results.

### What actually happened?

An out of volume coordinate returned `True` when the boundary voxel it was clipped onto was true.

## Reproducing the bug

```python
import numpy as np

from nimare.studyset import Studyset

studyset = Studyset(
    {
        "id": "example",
        "studies": [
            {
                "id": "study",
                "analyses": [
                    {
                        "id": "analysis",
                        "points": [
                            {"space": "MNI", "coordinates": [100, 0, 0]}
                        ],
                    }
                ],
            }
        ],
    },
    target=None,
)

mask = np.zeros((3, 3, 3), dtype=bool)
mask[2, 0, 0] = True

print(studyset.view.points_in_mask(mask, np.eye(4)).tolist())
```

This prints `[True]`. The coordinate maps to voxel index `[100, 0, 0]`, which is outside the image and should produce `[False]`.

SahilKumar75:
Closes #1134.

- Check transformed voxel indices against the mask shape before indexing the mask.
- Return `False` for coordinates outside the image volume instead of clipping them onto boundary voxels.
- Add a regression test with an out of volume focus and a true boundary voxel.

Testing:

- `python -m pytest nimare/tests/test_studyset_store.py -k points_outside_mask_volume -q`: 1 passed, 113 deselected.
- `black --check nimare/studyset/view.py nimare/tests/test_studyset_store.py`: passed.
- `isort --check-only nimare/studyset/view.py nimare/tests/test_studyset_store.py`: passed.
- `flake8 nimare/studyset/view.py nimare/tests/test_studyset_store.py`: passed.

## Summary by Sourcery

Reject coordinates outside image volumes instead of treating them as boundary-voxel coordinates.

Bug Fixes:
- Prevent out-of-volume coordinates from inheriting mask values from clipped boundary voxels.
- Ensure diagnostic and masking workflows safely handle voxel indices outside image bounds.

Enhancements:
- Add a shared voxel lookup utility that reports both retrieved values and whether indices are in bounds.

Tests:
- Add regression coverage for out-of-bounds voxel lookups and mask boundary behavior.


nimare/studyset/view.py
```diff
@@ -290,8 +290,13 @@ def points_in_mask(self, mask_data, affine):
...
         with np.errstate(invalid="ignore"):
             ijk = (xyz @ inv[:3, :3].T + inv[:3, 3]).astype(np.int32)
-        ijk = np.clip(ijk, 0, np.asarray(mask_data.shape) - 1)
-        return mask_data[ijk[:, 0], ijk[:, 1], ijk[:, 2]] > 0
+        shape = np.asarray(mask_data.shape)
+        in_bounds = np.all((ijk >= 0) & (ijk < shape), axis=1)
+        inside = np.zeros(len(ijk), dtype=bool)
+        if in_bounds.any():
+            valid = ijk[in_bounds]
+            inside[in_bounds] = mask_data[valid[:, 0], valid[:, 1], valid[:, 2]] > 0
```
jdkent:
```suggestion
flat_idx = np.ravel_multi_index((ijk[:, 0], ijk[:, 1], ijk[:, 2]), mask_data.shape, mode="clip")
return (mask_data.reshape(-1)[flat_idx] > 0) & in_bounds
```

after some benchmarking, this solution is approximately 15-30% faster, although, this is likely not a bottleneck in processing.

jdkent (changes requested):
Thanks for the change. 

it looks good! I discovered a faster implementation (look at the comment below.

Would you mind making this a utility function that is imported/used here, in the FocusFilter `transform` function, and in `_count_foci_per_cluster`?

I see the same/similar code is being reused and it could be abstracted away into a single function instead of being rewritten (sometimes incorrectly) multiple times.

Thank you!

SahilKumar75:
Thanks, @jdkent. I added a shared _get_voxel_values utility and now use it in View.points_in_mask, FocusFilter.transform, and _count_foci_per_cluster. The helper uses int64 indices and the suggested flattened lookup with np.ravel_multi_index. I also expanded the regression coverage to verify a true boundary voxel, a large out of bounds coordinate, and the helper behavior directly. Validation completed with 126 tests passed and 13 skipped across the utility and studyset modules, plus all focused diagnostics tests passing. Black, isort, and flake8 also pass on the modified files.

jdkent (approved):
LGTM, thank you for the PR!

PR #1137, opened 2026-09-10
Fix inclusive thresholds in decoder preprocessing

### LINKED ISSUE #1136: Inconsistent frequency_threshold comparisons in Decoder preprocessing and downstream decoding
## Summary

`NeurosynthDecoder` (and any `Decoder` subclass using the inherited preprocessing) filters features using `> frequency_threshold`, while `neurosynth_decode()` uses `>= frequency_threshold`. Consequently, features whose maximum value equals the threshold are removed before decoding.

https://github.com/neurostuff/NiMARE/blob/872486ee47f6e0939796ef7ff71f8b1810f7847f/nimare/decode/base.py#L53
https://github.com/neurostuff/NiMARE/blob/872486ee47f6e0939796ef7ff71f8b1810f7847f/nimare/decode/discrete.py#L289

Document: https://nimare.readthedocs.io/en/stable/generated/nimare.decode.discrete.NeurosynthDecoder.html#nimare.decode.discrete.NeurosynthDecoder

> Values greater than or equal to the threshold as assigned as label+, while values below the threshold are considered label-.

## Additional details

- NiMARE version: 0.21.0

Decode count-based annotations with `frequency_threshold=1`.

A feature with values `[1, 0]` should be retained, with the first analysis classified as positive, consistent with the documented `>=` behavior.

### What actually happened?

The feature is removed during preprocessing. If no features remain, fitting raises an exception; otherwise, affected features are omitted from decoding.

## Reproducing the bug

```python
from nimare.nimads import Studyset
from nimare.decode.discrete import NeurosynthDecoder

studyset = Studyset({
    "id": "example",
    "studies": [
        {
            "id": f"s{i}",
            "analyses": [{
                "id": f"a{i}",
                "name": "example",
                "annotations": {"term": count},
                "points": [{
                    "coordinates": [0, 0, 0],
                    "space": "MNI",
                }],
            }],
        }
        for i, count in enumerate([1, 0])
    ],
})

decoder = NeurosynthDecoder(features=["term"], frequency_threshold=1)
decoder.fit(studyset)
```

Expected: fitting succeeds and retains `term`, since one analysis meets `>= 1`.

The preprocessing code instead uses `> 1`, removing the only feature and raising:
`Exception: No features identified in the input Studyset/Dataset collection!`

const7:
Closes #1136.

- Treat values equal to `frequency_threshold` as positive in decoder preprocessing and pairwise group checks.
- Preserve ROI association's exclusion of features without positive annotations.


jdkent:
That was a clear inconsistency, thanks for fixing this!

jdkent (approved):
LGTM, The raising of an error for no features is new behavior, but I think that is a clearer message.

PR #1140, opened 2026-09-11
[FIX] ensure traversing slices of a studyset maintains the view

### LINKED ISSUE #1133: Filtered Studysets return inconsistent data across access methods
## Summary

Filtered Studysets return inconsistent data: `.analyses` and `.coordinates` reflect the selection, but nested traversal through `.studies → .analyses → .points` still includes excluded analyses and points.

## Additional details

- NiMARE version: 0.21.0

Iterate over analyses and points after filtering a Studyset.

All access paths should return only the selected analysis and point.

### What actually happened?
Nested traversal returns both analyses and all three points. This also affects Dataset conversion and Sleuth export. Analysis slicing preserved the nested selection in 0.20.0.

It appears that nested accessors retain the store but lose the View's selection. Would preserving that selection throughout nested access be the intended behavior?

## Reproducing the bug

```python
from nimare.studyset import Studyset

ss = Studyset({
    "id": "example",
    "studies": [{
        "id": "S1",
        "analyses": [
            {"id": "A1", "points": [
                {"space": "MNI", "coordinates": [1, 2, 3]},
                {"space": "MNI", "coordinates": [4, 5, 6]},
            ]},
            {"id": "A2", "points": [
                {"space": "MNI", "coordinates": [7, 8, 9]},
            ]},
        ],
    }],
}, target=None)

sub = ss.slice(["S1-A1"]).select_points([True, False, False])

print(list(sub.ids))                   # ['S1-A1']
print(sub.coordinates[["x", "y", "z"]]) # Only [1, 2, 3]

for study in sub.studies:
    for analysis in study.analyses:
        print(analysis.id, [p.coordinates for p in analysis.points])
# A1: [[1, 2, 3], [4, 5, 6]]
# A2: [[7, 8, 9]]
```

`sub.analyses[0].points` also ignores the point filter, and reverse navigation through `analysis.study.analyses` restores excluded analyses.

jdkent:
Closes #1133 .

- keep the filtered view when accessing child/parent objects. (Studysets can be different views of the same store; the storage is immutable, and the studysets are immutable as well, but multiple studysets can reference the same store)

## Summary by Sourcery

Keep Studyset selections intact across nested traversal and all supported export paths.

Bug Fixes:
- Preserve analysis and focus selections when traversing nested Studyset accessors, including parent and child relationships.
- Ensure coordinates, image references, and exported representations remain consistent with the active Studyset view.

Enhancements:
- Propagate view context through nested accessors and validate point-selection masks against the full store.
- Optimize nested selection lookups and parent-study discovery for filtered views.

Documentation:
- Clarify that nested accessors remain confined to the originating Studyset view and document point-mask alignment.

Tests:
- Add coverage for filtered nested traversal, view-specific coordinate spaces, resolved image references, exports, and Sleuth conversion.

### DIFF
--- a/nimare/studyset/__init__.py
+++ b/nimare/studyset/__init__.py
@@ -22,7 +22,9 @@ The layering, which the module boundaries enforce:
 
 ``nested``
     read-only ``Study``/``Analysis``/``Point``/``Image`` accessors over the
-    columns, for callers that want to walk the studyset as objects.
+    columns. A walk is
+    confined to the view it started from, so it reports what that view's frames
+    and blocks report.
 
 Data flows store -> view -> block and never back.
 
--- a/nimare/studyset/nested.py
+++ b/nimare/studyset/nested.py
@@ -1,10 +1,13 @@
 """Read-only nested accessors over the columns.
 
 Convenient for inspection and for code that reads a studyset the way the document
-is shaped. These are *views*, not storage: they hold a store and a row, and every
-attribute is a column lookup. The previous implementation kept an equivalent
-object graph as a second copy of the data, which is what the revision counters
-and mutation tracking existed to keep in step.
+is shaped. An accessor is a view and a row, and
+every attribute is a column lookup.
+
+The view travels with the accessor rather than being applied once at the top: an
+analysis reached from a sliced studyset shows only its selected foci, the study
+reached back from that analysis hides the same analyses the slice did, and both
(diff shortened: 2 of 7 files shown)


jdkent:
the benchmark is still a bit noisey, there's nothing that really impacts mkdachi2, and even if there was a slight regression in speed, it's worth it for correctness.

PR #1141, opened 2026-09-11
[ENH] mkda sample size weighting

jdkent:
Based on the matlab implementation that already takes into account sample size, and this makes MKDA more directly comparable to ALE.

Closes # .

-
-

## Summary by Sourcery

Enable optional sample-size weighting in MKDA and provide statistically consistent null distributions, configuration APIs, documentation, and regression coverage.

New Features:
- Add opt-in sample-size weighting to MKDA using square-root sample-size scaling, with support for explicit weights and fixed-effects discounts.
- Expose configurable handling of missing weights, transformations, reductions, and weighting metadata through the new StudyWeights API.

Bug Fixes:
- Correct summary-statistic threshold derivation from approximate and Monte Carlo null histograms by using upper-tail accumulation.

Enhancements:
- Implement weighted MKDA null distributions for continuous weighted statistics while preserving exact integer behavior for uniform weights.
- Renormalize weights over analyzed subsets and include weighting details and effective sample size in generated method descriptions.
- Optimize histogram bin lookup and weighted permutation histogram generation.

Documentation:
- Document MKDA sample-size weighting and add an example demonstrating its effects and configuration options.
- Add the supporting Wager et al. weighting reference.

Tests:
- Add comprehensive tests for weighting formulas, metadata handling, normalization, fixed-effects discounts, imputation, weighted null distributions, subset alignment, Monte Carlo behavior, and threshold monotonicity.

### DIFF
--- a/docs/dev/mkda_sample_size_weighting_design.md
+++ b/docs/dev/mkda_sample_size_weighting_design.md
@@ -1,0 +1,689 @@
+:orphan:
+
+# Sample-size weighting for MKDA
+
+Design for weighting MKDA density results by study sample size, per
+Wager et al. (2009), NeuroImage 45(1 Suppl):S210-21, doi:10.1016/j.neuroimage.2008.10.061
+(PMC3318962).
+
+## 0. Decisions
+
+Settled by the primary sources and by measurement. The evidence column points at the
+section that carries it.
+
+| # | Decision | Evidence |
+|---|---|---|
+| D1 | **Weight is `delta_c * sqrt(N_c)`, renormalised across the contrasts in the analysis.** This is Wager 2009's equation verbatim, and it is what the CANlab MATLAB reference computes. | §1.1, §1.2 |
+| D2 | **Normalise to sum to `k`, not to 1.** CANlab's weights sum to 1, making the statistic a proportion. NiMARE's MKDA statistic is already `k x` that proportion. Keeping the `k` scaling makes the unweighted path bit-identical and the weighted statistic monotone-equivalent to CANlab's. | §2.1, §3.1 |
+| D3 | **Weighting is opt-in via a `weighting=` argument that defaults to `None`.** Turning it on by default would silently change every existing MKDA result. | §3.2 |
+| D4 | **The approximate null must become a weighted Poisson-binomial.** Today's exact convolution assumes unit weights. Plugging non-uniform weights into the shipped code and leaving the null alone gives Dice 0.80 against the correct answer at z>1.65, and misses a third of the suprathreshold volume. This is the substantive part of the work, not the weight formula. | §2.2 |
+| D5 | **`CBMAEstimator._p_to_summarystat`'s montecarlo branch is broken and must be fixed first.** It compares raw counts to a p-value without normalising or accumulating, so it returns the same threshold for p=0.01 and p=0.001 (6, where the correct answers are 2 and 3). It is currently conservative on integer bins; on the fine grid D4 needs, it flips to returning ~0. | §2.3 |
+| D6 | **`N` for a contrast is the *mean* over the analysis's `sample_sizes` entries**, matching what the ALE kernel does with the same field. Every NiMARE converter writes `sample_sizes` as a one-element list per contrast, and the Sleuth exporter reduces it with `min` -- the field is not a per-group decomposition to sum. | §3.3 |
+| D7 | **The fixed-effects discount is opt-in and label-driven; never inferred.** NIMADS has no fixed/random field and NiMARE has no such convention. Default `delta = 1` for every contrast; the user names a metadata field to activate `delta = 0.75`. | §1.3, §3.3 |
+| D8 | **Missing, zero or NaN `N` imputes the mean weight and warns.** This is CANlab's behaviour. Dropping those studies would silently change the study set relative to the unweighted analysis. | §1.2, §3.5 |
+| D9 | **Weights align to MA-map rows by study id, with an assertion — not by positional coincidence.** The current code relies on `groupby("id").first()` happening to match the kernel's `np.unique(id)` ordering. It does today; nothing enforces it. | §2.4 |
+| D10 | **`MKDAChi2` and `KDA` are out of scope.** Wager 2009 applies the weights to the density statistic only; its chi-square tests run on unweighted counts. | §1.4, §6 |
+| D11 | **Replace the body of `MKDADensity._compute_weights` rather than extend it.** It reads `inference` and `sample_size` columns that `_preprocess_input` never creates for an MKDA kernel, so it has always returned uniform weights. It is dead code with a live-looking `TODO`. | §2.1 |
+
+## 1. Primary sources
+
+### 1.1 Wager et al. (2009), the weighting equation
+
+> "Rather than weighting Z-scores, the current version of MKDA weights by the square root
+> of the sample size for each SCM. In addition, we down-weight studies using fixed effects
+> analyses by a factor of 0.75."
+
+    P = sum_c I_c ( delta_c sqrt(N_c) / sum_c delta_c sqrt(N_c) )
+
+`c` indexes comparison (contrast) maps, `I_c` is the indicator that the voxel lies within
+`r` mm of a peak reported by contrast `c`, `delta_c` is the fixed-effects discount, `N_c`
+is the sample size. On the scope of the scheme:
+
+> "This weighting scheme could be used to weight by other study quality measures developed
+> by the analyst, such as diagnostic criteria or sample-matching procedures employed in
+> studies of psychiatric or medical populations."
+
+Note what `sqrt(N)` is and is not. It is not inverse-variance weighting — that would be
+`~N` for a standardised mean difference. It is a weighting on the *evidence* scale: a
+study's Z-statistic grows like `sqrt(N) * d`, and `I_c` is a thresholded statistic rather
+than an effect size. The choice is a defensible heuristic, not a derived optimum, and the
+0.75 discount is arbitrary by the authors' own presentation. §4.1 returns to this.
+
+### 1.2 CANlab reference implementation
+
+`densityUtility3/Meta_Setup.m`, the canonical MATLAB implementation:
+
+```matlab
+DB.rootn = sqrt(DB.Subjects);
+[DB.connumbers,DB.pointind] = unique(DB.Contrast);
+DB.rootn = DB.rootn(DB.pointind);
+...
+if isfield(DB,'SubjectiveWeights')
+    w = DB.rootn .* DB.SubjectiveWeights(DB.pointind);
+else
+    w = DB.rootn;
+end
+
+% make sure no NaNs or zeros; impute mean
+whbad = find(isnan(w) | w == 0);
+if ~isempty(whbad)
+    disp(['Warning! ' num2str(length(whbad)) ' contrasts have bad or missing weights. Imputing mean. ']);
+    wok = w; wok(whbad)=[];
+    w(whbad) = mean(wok);
+end
+
+% these must sum to 1 !
+DB.studyweight = w ./ sum(w);
+```
+
+Three things follow. The weight is *per contrast*, taken once from the contrast's first
+peak. The fixed-effects discount is not automatic — despite what `read_database.m`'s
+header says, no shipped CANlab code derives `SubjectiveWeights` from `FixedRandom`; the
+analyst supplies it. And bad weights are imputed, not dropped.
+
+`Support_functions_density/meta_prob_activation.m` applies them:
+
+```matlab
+for i=1:nc, ivectors(:,i) = ivectors(:,i) .* wts(i); end
+activation_proportions = sum(ivectors, 2);
+```
+
+`meta_stochastic_activation_blobs.m` — the Monte Carlo null — keeps each weight attached
+to its own contrast while the contrast's blobs move (`indic = indic .* MC_Setup.wts(c)`),
+and errors outright on NaN weights. `Meta_Select_Contrasts.m` recomputes `w ./ sum(w)`
+over the retained subset, which is what D2 and §3.6 have to reproduce for leave-one-out.
+
+For between-condition contrasts, `meta_apply_contrast.m` rescales to mean-1 weights
+(`wts = wts ./ mean(wts)`) and runs weighted OLS on the condition indicator matrix, so
+the betas are weighted proportions *within* condition. That is the shape a weighted
+two-sample MKDA would take; see §6.
+
+### 1.3 Whether the fixed-effects discount is empirically supported
+
+Bossier et al. (2018), "The Influence of Study-Level Inference Models and Study Set Size
+on Coordinate-Based fMRI Meta-Analyses" (PMC5778144), concludes that "the best balance
+between TPR and FPR detection is observed when using mixed effects group level models
+together with a fixed or random effects meta-analysis," and reports fixed-effects
+study-level models performing consistently worse in AUC. That supports the *direction* of
+a discount. It supports no particular magnitude, and 0.75 remains a convention.
+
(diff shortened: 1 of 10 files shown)


docs/dev/mkda_sample_size_weighting_design.md
```diff
@@ -0,0 +1,689 @@
...
+class StudyWeights(NiMAREBase):
+    def __init__(
+        self,
+        source="sample_size",          # "sample_size" | "uniform" | mapping | array
+        transform="sqrt",              # "sqrt" | "linear" | "none"
+        reduce="sum",                  # collapse per-group sample_sizes: "sum" | "mean"
+        inference_field=None,          # metadata field naming the study-level model
+        fixed_effects_discount=0.75,
+        fixed_effects_labels=("fixed", "ffx", "fixed-effects", "fe"),
+        on_missing="impute",           # "impute" | "raise"
```
jdkent: I would prefer to raise over impute missing sample size data.

docs/dev/mkda_sample_size_weighting_design.md
jdkent, line 1: this whole file should not be a part of the commit, if there are things that should be stated, it should go into the official documentation.

nimare/meta/cbma/mkda.py
```diff
@@ -31,11 +32,114 @@
...
+    return current
+
 
 class MKDADensity(CBMAEstimator):
     r"""Multilevel kernel density analysis- Density analysis.
 
     The MKDA density method was originally introduced in :footcite:t:`wager2007meta`.
+    Sample-size weighting follows :footcite:t:`wager2009evaluating`.
+
+    .. versionchanged:: 0.5.0
```
jdkent:
```suggestion
    .. versionchanged:: 0.22.0
```

nimare/meta/cbma/mkda.py
```diff
@@ -194,35 +330,84 @@ def _generate_description(self):
...
-        inf = ids_df["inference"].map({"ffx": 0.75, "rfx": 1.0}).values
+        if self._raw_weights_ is None:
+            raw = np.ones(n_exp, dtype=np.float64)
+        else:
+            aligned = self._raw_weights_.reindex(study_ids)
+            if aligned.isna().any():
+                missing = list(aligned.index[aligned.isna()])
+                raise ValueError(
+                    f"No weight was computed for {len(missing)} of the {n_exp} contrasts "
+                    f"being analysed: {', '.join(str(i) for i in missing[:5])}."
```
jdkent: how are weights being passed in so they have to be aligned? the sample size should be accessible from within the studyset and should already be aligned.

jdkent (commented):
some comments on the PR

docs/dev/mkda_sample_size_weighting_design.md
jdkent: Done in 5b1591fc — `on_missing` now defaults to `"raise"`. `"impute"` is still there for anyone who wants CANlab's behaviour, but it has to be asked for, and the error message names the option.

Agreed on the reasoning: an imputed weight is indistinguishable from a real one once it is in the map, so the failure is silent exactly where it matters most.

docs/dev/mkda_sample_size_weighting_design.md
jdkent: Removed in 5b1591fc.

The user-facing part is now a "Study weights" section in `docs/cbma.rst`, under Estimators next to Null methods: the weight itself, how to turn it on, the `StudyWeights` escape hatch for non-sample-size weights, the sum-to-*n* normalisation convention, and the sample-size requirement (including that most corpora do not meet it). The rest of that file was working notes and review scratch, which is your point — it does not belong in the repo.

nimare/meta/cbma/mkda.py
jdkent: Applied in 5b1591fc, and I did the same for `versionadded` in `weights.py`, which had the same wrong number.

nimare/meta/cbma/mkda.py
jdkent: You are right, and I have removed it in 5b1591fc.

To answer the question directly: nothing is passed in. The weights are read from the collection in `_preprocess_input` against `inputs_["id"]`, and the kernel emits one MA-map row per study in `np.unique(coordinates["id"])` order. Those two are the same sequence — `inputs_["id"]` comes from `narrowed.ids`, which is sorted — so the rows and the weights already line up. I checked rather than assumed: reversing the study order in the source `Studyset` still yields sorted `ids`, and `test_MKDADensity_weighting_matches_published_formula` already covers the reversed-order case. So the `reindex` plus the "no weight was computed" branch were dead code in the normal path.

The one place the IDs did real work is the subsample path, where `_approximate_z_from_ma` analyses a subset of rows and the weights have to renormalise over just those. That is why it was re-deriving the IDs from the coordinates at all — `_prepare_subsample_null` filtered `inputs_["coordinates"]` first so the re-derivation would pick up the subset.

That indirection is gone. `_prepare_subsample_null` now passes `subset_study_ids` — which its caller already has — straight through to `_compute_weights`, so:

- the full-sample case indexes nothing and just uses the collected weights in order;
- the subsample case does one `.loc` on the ID-indexed Series;
- `_prepare_subsample_null` no longer mutates `inputs_["coordinates"]`, which also removes the reason `_approximate_z_from_ma` had to hand the temporary estimator its own `inputs_` dict.

Net effect is 13 fewer lines. `test_MKDADensity_weights_renormalise_over_a_subset` covers the subset renormalisation and `test_diagnostics.py` (41 tests) covers the path that lost the mutation; both pass.

examples/02_meta-analyses/17_plot_mkda_sample_size_weighting.py
```diff
@@ -0,0 +1,263 @@
...
+    "pain_09.nidm-1": 0.0420535511130868,
+    "pain_10.nidm-1": 0.0420535511130868,
+    "pain_11.nidm-1": 0.0420535511130868,
+    "pain_12.nidm-1": 0.0437707237544620,
+    "pain_13.nidm-1": 0.0686731613994095,
+    "pain_14.nidm-1": 0.0594727023300776,
+    "pain_15.nidm-1": 0.0454230267018595,
+    "pain_16.nidm-1": 0.0454230267018595,
+    "pain_17.nidm-1": 0.0420535511130868,
+    "pain_18.nidm-1": 0.0420535511130868,
```
jdkent: the canlab comparison should be in benchmarks, not a notebook

docs/cbma.rst
```diff
@@ -40,6 +40,60 @@ The Estimator classes take a kernel object as a parameter, and use that kernel t
...
+either way. Dividing by the contrast count recovers the weighted *proportion* of
+:footcite:t:`wager2009evaluating`.
+
+.. important::
+    Weighting requires a sample size for every contrast, and most coordinate corpora
+    record one for only a small fraction of their analyses. A contrast with a missing or
+    non-positive sample size raises by default; pass
+    ``StudyWeights(on_missing="impute")`` to give those contrasts the mean weight of the
+    rest instead, which is what the CANlab MATLAB implementation does.
+
```
jdkent: delete this section, neurostore has a lot of sample sizes reported.

examples/02_meta-analyses/17_plot_mkda_sample_size_weighting.py
```diff
@@ -0,0 +1,263 @@
...
+        cmap="RdBu_r",
+        symmetric_cbar=True,
+        threshold=1.65,
+        title=name,
+    )
+
+###############################################################################
+# The two maps are highly correlated -- weighting rescales the same contrasts rather
+# than changing which ones enter -- but the ranking of voxels does move, and any
+# threshold acts on that ranking.
```
jdkent:
```suggestion
# than changing which ones enter -- but the ranking of voxels does move, and 
# as a consequence different voxels may survive the threshold
```

examples/02_meta-analyses/17_plot_mkda_sample_size_weighting.py
```diff
@@ -0,0 +1,263 @@
...
+#     DB.studyweight = w ./ sum(w);                        % sums to 1
+#
+# NiMARE, :func:`~nimare.meta.cbma.weights.normalize_weights`::
+#
+#     raw = discount * transform(sample_size)              # delta_c * sqrt(N_c)
+#     weight_vec = n_studies * raw / raw.sum()             # sums to n_studies
+#
+# So CanLab reports a weighted proportion and NiMARE a weighted sum. If the relative
+# weights agree, ``nimare_weight / n_contrasts`` should equal CanLab's ``studyweight``.
+# The values below were produced by running ``Meta_Setup`` on this same Studyset under
```
jdkent: delete the entire comparison section, this can be test that is used for reference.

examples/02_meta-analyses/17_plot_mkda_sample_size_weighting.py
```diff
@@ -0,0 +1,263 @@
...
+# implementations still rasterise their spheres slightly differently, so the maps are
+# not bit-identical even though the weights are.
+
+###############################################################################
+# When sample sizes are missing
+# -----------------------------------------------------------------------------
+# Most coordinate corpora record a sample size for only a small fraction of their
+# analyses, so this is the common case rather than an edge case. A contrast with a
+# missing or non-positive sample size raises by default, because a substituted weight
+# is invisible in the output map.
```
jdkent:
```suggestion
# A contrast with a missing or non-positive sample size raises
# by default, because a substituted weight is invisible in the output map.
```

nimare/meta/cbma/base.py
```diff
@@ -42,6 +42,32 @@
 _UNSET = object()
 
 
+def _nullhist_to_summarystat(hist_weights, bin_centers, p):
```
jdkent, line 57: I see this function is an abstraction over it being written several times, that's an okay abstraction, please profile this function.

nimare/meta/cbma/mkda.py
```diff
@@ -72,6 +176,25 @@ class MKDADensity(CBMAEstimator):
         Number of iterations to use to define the null distribution.
         This is only used if ``null_method=="montecarlo"``.
         Default is 5000.
+    weighting : None, {"sample_size", "uniform"}, or \
```
jdkent, line 180: Does None, also mean uniform? it's not well described in the description, please change (without making the description longer; it's already too long)

jdkent (commented):
there are a few documentation changes and I want this code to be profiled/optimized before merging in.

nimare/meta/cbma/base.py
jdkent: Profiled in 98da34f.

It is not a bottleneck: `_p_to_summarystat` calls it once per correction, and at the worst case this PR can produce (a 100k-bin weighted grid) it took 0.81 ms. So the abstraction costs nothing measurable.

It was doing more passes over the histogram than it needed, though, so I tightened it anyway — normalise after the accumulation instead of before, and drop the `np.max`/`np.squeeze`, since the first entry of an upper-tail cumulative sum *is* the total. 0.81 ms → 0.58 ms, same output (`test_nullhist_to_summarystat_accumulates_tail` unchanged and passing).

The profiling did turn up two real ones nearby; details in the PR comment below.

---

examples/02_meta-analyses/17_plot_mkda_sample_size_weighting.py
jdkent: Removed in 98da34f. The CanLab `studyweight` values and the comparison are now `test_weights_match_the_canlab_reference_implementation` in `nimare/tests/test_meta_weights.py`, which asserts the same `nimare_weight / n_contrasts == canlab_studyweight` agreement to 1e-12. The example no longer mentions MATLAB at all.

---

jdkent:
Pushed 98da34f (on top of a merge of current `main`, which carries the nearest-bin histogram fixes from #1144/#1147 — the MKDA override already used the same rounding convention, so nothing changed there).

## Review comments

- `docs/cbma.rst`: the "most corpora have no sample size" block is gone. All that survives is one sentence saying a missing or non-positive sample size raises and that `on_missing="impute"` is the way out. Same claim removed from the example's intro and its missing-data section.
- Example: CanLab comparison deleted, both wording suggestions applied, the trailing "ALE and SDM also need sample size" note dropped.
- `MKDADensity.weighting`: now says `None` *and* `"uniform"` both mean equal weights, and the ALE contrast is one clause instead of a paragraph — the entry is three lines shorter than before.

## Profiling

Setup: `MKDADensity(weighting="sample_size")` on the NIDM pain Studyset, 228k-voxel mask, and a synthetic 400-contrast Studyset for scaling.

**Price of weighting**, after the changes below:

| null | unweighted | weighted |
|---|---|---|
| `approximate` | 0.043 s | 0.054 s |
| `montecarlo`, 300 iters | 0.711 s | 0.965 s |

The `montecarlo` gap is inherent: a continuous statistic needs a 100k-bin grid, so each permutation bins a full brain map onto 100k bins instead of 22. `_uniform_bin_counts` is already ~7x faster than the `np.histogram` it replaced, and I could not beat it — I tried reusing scratch buffers across permutations, measured it in the real loop, and it was **not** faster (numpy already recycles those allocations), so it is not in the commit.

**Two things were worth fixing:**

1. `_compute_weights` looked its subset weights up through a pandas label index (`Series.loc[list_of_ids]`) — **630 µs per call**. That path runs once per leave-one-out replicate *and* once per stability iteration, so on a default `ResampledStability` run it was several seconds of pure lookup. Keeping the weights as an array plus an id→position dict does the same work in **8 µs** (16.7x on the whole `_compute_weights` call). It also gives a real error message when a subset names a study the fit never weighted.

2. `_get_last_bin` (in `nimare/meta/utils.py`) built the index of *every* non-zero bin just to read the last one. That was fine at 22 bins; at 100k, once per permutation, it cost 0.33 ms. Scanning from the end short-circuits instead: **0.10 ms**, and the Monte Carlo null went **1.06 s → 0.94 s per 300 iterations (11%)**. This helps ALE and KDA too, which call it on the same shaped array. Identical results, with a test for the all-zero and empty cases.

3. `_nullhist_to_summarystat`, per your comment above — profiled, not a bottleneck, tightened anyway. 0.81 ms → 0.58 ms.

Remaining costs, for the record: `_weighted_null_histogram` is O(contrasts × bins) — 2 ms at 21 contrasts, 47 ms at 400 (14% of that fit). `_approximate_z_from_ma` already reuses a precomputed null across replicates, so it runs once, not once per replicate.

## One thing for you to decide

`n_histogram_bins=100_000` looks 10x larger than it needs to be. Comparing z-maps against a 4M-bin reference:

| `n_histogram_bins` | max \|Δz\| | mean \|Δz\| |
|---|---|---|
| 1 000 | 6.8e-02 | 8.6e-04 |
| 10 000 | 5.0e-02 | 1.1e-03 |
| 100 000 | 5.0e-02 | 9.9e-04 |
| 1 000 000 | 1.9e-02 | 3.1e-05 |

10k and 100k are indistinguishable, because at that point the error is the observed statistic being snapped to a bin by `nullhist_to_p`, not the convolution. Dropping the default to 10k would make `_weighted_null_histogram` ~10x cheaper (47 ms → ~5 ms at 400 contrasts), cut the Monte Carlo null ~8%, and use 10x less memory for the null arrays — for no measurable change in the map. I did not change it, because it is a statistical default and the accuracy claim in the docstring ("better than 2e-10", which I did reproduce — 2.2e-11 absolute, against brute-force enumeration over all 2^16 outcomes) is yours to move. Say the word and it is a one-line change.

Tests: 219 passed across `test_meta_mkda`, `test_meta_weights`, `test_meta_ale`, `test_diagnostics`, `test_utils`, `test_stats`. `flake8 nimare/` clean.

---

jdkent (commented):
same formulations, and numerically identical output, LGTM.

PR #1143, opened 2026-09-12
[FIX] Round to the nearest bin in the ALE approximate null

jdkent:
Found while cross-checking NiMARE against GingerALE for #998.

`_study_ma_histogram` and `_update_ale_histogram` in `nimare/meta/cbma/null_utils.py` address the histogram with `int(value * inv_step_size)`, which truncates. The grid those indices address is defined by bin *centres* — `_determine_histogram_bins` says so: *"bins are -.0005-.0005, .0005-.0015, etc."* — so the correct lookup is `round`.

Truncating moves probability mass down by up to a full bin **once per study in the convolution**, so the null drifts low in proportion to the number of studies and uncorrected p-values come out too small.

**Validated against an independent computation.** The same convolution at the same resolution, with correct rounding, recomputed from the same MA maps; it is converged (going from 1e-5 to 2e-6 moves tail p by under 0.6%). GingerALE 3.0.2 matches that reference to its float32 output precision.

pain21, at the peak ALE value:

| | p |
|---|---|
| NiMARE, as shipped | 2.4635e-12 |
| NiMARE, with this fix | 2.6883e-12 |
| GingerALE 3.0.2 | 2.6812e-12 |

**The bias grows with corpus size** (ratio of shipped to corrected p, 99th percentile): 0.965 at 19 studies, 0.920 at 50, 0.843 at 100, **0.703 at 200**.

**Downstream**, pain21 at 1000 permutations with the same mask and threshold, NiMARE declared 666 voxels significant at voxel-level FWE against GingerALE's 566 — 17.7% more, in the direction the bias predicts.

Three reference implementations in `test_meta_ale.py` mirrored the flooring they were checking and are corrected alongside; `_dense_ale_reference` also built its per-study histogram edges from the bin centres rather than half a step either side, which floors by a different route.

`test_meta_ale.py` passes (48).

## Summary by Sourcery

Correct approximate ALE null binning to use nearest bin centres and produce unbiased null distributions and p-values.

Bug Fixes:
- Correct approximate ALE null histogram binning so values and convolution scores are assigned to their nearest bin centre, preventing study-count-dependent downward bias in null distributions and p-values.

Enhancements:
- Align dense and reference ALE implementations with centre-based histogram binning and update edge-case coverage.

Tests:
- Update ALE histogram reference tests and add coverage confirming values just below a bin centre remain in that centre's bin.

### DIFF
--- a/nimare/meta/cbma/null_utils.py
+++ b/nimare/meta/cbma/null_utils.py
@@ -121,7 +121,7 @@ def _study_ma_histogram(study_ma_values, n_zero_voxels, mask_voxel_recip, inv_st
     """Bin one study's nonzero ALE values onto the fixed approximate-null grid."""
     exp_hist = np.zeros(n_bins, dtype=np.float64)
     for i_val in range(study_ma_values.shape[0]):
-        idx = int(study_ma_values[i_val] * inv_step_size)
+        idx = int(round(study_ma_values[i_val] * inv_step_size))
         if idx < 0:
             idx = 0
         elif idx >= n_bins:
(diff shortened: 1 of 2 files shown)


jdkent:
closed in favor of #1147

PR #1146, opened 2026-09-13
Add CBES: coordinate-based effect-size meta-analysis estimator

jdkent:
Closes # .

`CBES` is a coordinate-based **effect-size** estimator in `nimare/meta/cbma/effectsize.py`, exported from `nimare.meta.cbma`. Where ALE and (M)KDA ask *where* studies agree, this asks *how big* the effect is — and it gets there from the coordinates' silence rather than from their reported heights.

**This description was rewritten from scratch after a redesign.** An earlier version of this PR described a different estimator: peak heights pooled as effect sizes, a spatial kernel, a `peak_bias` correction, `g_relative`/`g_absolute` maps, and an arrangement null that shuffled effect sizes among the foci of an analysis. All of that is gone. Anything you may have read about `peak_bias`, `fwhm`, `stat_column`, `use_images` or the arrangement null no longer applies.

## What it does

A reported peak height is a poor estimate of the effect at its own location: it was selected for being large and it sits wherever noise pushed the maximum. Regressing a held-out truth on the effect size a focus's own table reports gives a slope of 0.08 to 0.18 — one tabulated coordinate explains 5% to 9% of the variance in the effect at its own location. So the heights are not read.

The same selection makes *silence* informative. If a study reported nothing near a voxel, its effect there did not clear that study's reporting threshold, which is a genuine upper bound. Across a literature those bounds constrain the magnitude.

- **Images carry the magnitude.** Each voxel pools the studies supplying `g`/`g_var` maps by local inverse-variance weighting with a per-voxel DerSimonian–Laird `tau2`. A collection must supply at least one such study; one is enough to run, and a fit without any is refused rather than returned as zeros.
- **Coordinates carry only a reporting indicator, never a magnitude.** Three zones per study per voxel: **−1** within `report_radius` of one of that study's foci, **+1** beyond `coverage_radius` from all of them (silent), and **0** in the ring between, which says nothing either way. Both limbs are needed — silences alone leave the model reading the silence fraction against a denominator that excludes every study that reported. The labels are set, never accumulated, so a study with two nearby foci still contributes exactly one −1 there.
- **The report extent adapts to the collection.** How far a report should assert its lower bound depends on how badly the reports are outnumbered, and that spans two orders of magnitude across real studysets: a reported voxel carries 5 silences per report on NIDM pain's nine tables and 605 on a 1,443-study NeuroStore corpus, where both `g` and `prevalence` collapse for want of anything holding them up. `report_radius="adaptive"` measures that ratio and picks the named voxel below 50 or 4 mm above it, recording the choice on `report_radius_`. `silence_to_report_ratio` is public, since it is computable before fitting and says whether a coordinate corpus can bound the effect from below at all.
- **Non-reporting is censoring, not imputation.** A zero-inflated censored likelihood, fitted by EM, separates *how big* the effect is where present (`mu`, reported as `g`) from *how often* it is present (`pi`, reported as `prevalence`), and reports their product as `g_marginal`. Nothing is imputed.
- **The reporting threshold is supplied or assumed, not inferred.** Pass `threshold` a float, or a metadata field name for per-study values. `clamp_threshold` lowers a study's assumed cut to its own smallest reported statistic when the table contradicts the assumption — a hard upper bound, never below the truth, and bit-identical where it should do nothing. `reporting_cutoff_to_g` is the only place a reported statistic's scale enters the model.
- **Inference is a within-study image permutation.** `permute-images` scrambles each image study's values among that study's own voxels and holds the silence pattern fixed, so the censoring term cancels between the observed fit and every permutation. `correct_fwe_montecarlo` adds voxel- and cluster-level FWE with a generalized-Pareto tail.
- **`interval="profile"`** additionally emits `g_lower`/`g_upper` from the profile likelihood, which inverts no matrix and needs no `dof`.
- **`analysis_mask`** lets an ROI or partial-coverage study declare which voxels it examined, so its silence elsewhere is not read as evidence.
- **`nimare/generate.py`** gains `n_image_studies` and `image_dir` on `create_effect_size_coordinate_studyset`, which simulates each study's whole statistic field and reports the local maxima that clear its own threshold — without which there is no way to build a valid CBES input.

Maps: `g`, `se`, `z`, `p`, `logp`, `tau2`, `n_studies`, `n_eff`, `dof`; plus `prevalence`, `g_marginal`, `se_marginal` and `coordinate_share` under the zero-inflated model; plus `g_lower`/`g_upper` under `interval="profile"`.

## Which map to read

An image-based meta-analysis pools per-study effect maps, so where a fraction `pi` of studies carry an effect `mu` it targets the population mean `pi*mu`. **That makes `g_marginal` the IBMA-comparable estimand, and `g` a quantity no image-only method estimates**: the mean effect among the studies whose effect at that voxel is non-null.

The choice is not clean, because `pi` is the weakest part of the fit. On the NIDM pain collection, scored against a pool of held-out studies — itself a `pi*mu` quantity, so `g_marginal` is the estimand-matched arm — `g_marginal` wins on rmse (0.184 against 0.230) but undershoots by 0.291 at the strongest voxels where `g` sits at +0.005, because the fitted prevalence is biased low.

Read `g_marginal` when comparing against an IBMA or when studies genuinely differ in whether they carry the effect. Read `g` when the question is how large the effect is where present. **To power a new study, use `g`** — power is defined conditional on the alternative being true.

## What it recovers

On the 21-study NIDM pain collection, split in half so the reference comes from studies the coordinates never touched, eight paired splits, using **the coordinates the papers actually printed** (the collection's own 267 transcribed peaks):

| estimate | bias | at top | rmse | rank r | AUC |
| --- | --- | --- | --- | --- | --- |
| images only, pooled | +0.136 | +0.137 | 0.269 | 0.484 | 0.893 |
| CBES with the silence off | +0.142 | +0.155 | 0.272 | 0.499 | 0.899 |
| **CBES `g`** | **+0.069** | **+0.005** | **0.230** | 0.480 | 0.886 |
| CBES `g_marginal` | −0.064 | −0.291 | 0.184 | 0.483 | 0.887 |

rmse falls 14% against pooling the images alone (paired p = 0.0024) and the mean bias 49% (p = 0.0001), while the +0.137 overestimate at the strongest voxels becomes +0.005 — the best-centred top stratum measured anywhere in this work. The ordering is unmoved (rank r −0.004 at p = 0.84, AUC −0.006 at p = 0.30) and the Pearson cost is not significant (0.030, p = 0.075). **The coordinate channel corrects the level and leaves the pattern alone.**

Also run on tables re-extracted from the same studies' maps. The two are not interchangeable: the cluster scheme recovers only 23% of pain's published peaks within 8 mm and finds 119 peaks where the papers printed 267, so it reads more voxels as silent than the literature does. That arm gives a larger rmse gain (22%) and a worse-centred top (−0.065), consistent with more shrinkage. The published row is the one to quote.

Against SDM-PSI at one image, CBES leads on pattern on every dataset tried — HCP `EMOTION_FACES` r +0.882 against +0.482 and AUC 0.984 against 0.737; a NeuroStore working-memory studyset r +0.328 against +0.075.

## How many images before the coordinates start hurting

The coordinate channel carries a bias that does not shrink as coordinate studies accumulate, while the image channel carries noise that does. Working out when the combination still beats the images alone, the weight cancels and what is left is

```
coordinates help   <=>   b^2  <  1/(m*I_c) + 1/(k*n)
```

squared channel bias below the sum of the two channels' variances — verified symbolically, and against simulation where the predicted crossing matched the measured one in every scenario. So there is a crossing, and **on pain it is 3 images**: paired over the same 8 splits with published tables, `g` beats an image pool at 1–2 images (rmse −0.092 and −0.040, p ≤ 0.002), ties at 3, and is significantly worse from 5 (+0.027, p = 0.001). Bias at the top decile crosses later, around 6 images.

Held at a fixed wall of coordinate studies while images are added one at a time — 1,443 NeuroStore pain studies and 41,874 peaks against NIDM pain images — the same condition predicts and finds two things. At one image the wall is the estimator's best configuration, rmse 0.236 against the pool's 0.366. But once `1/(m*I_c)` vanishes only the bias is left, so the wall protects it: the estimate moves just 0.236 to 0.224 across 1 to 11 images while the pool falls to 0.138, and at 11 images more coordinate studies is strictly worse (0.141 at 25 tables, 0.163 at 200, 0.224 at 1,443).

The reason is that a large topic-wide corpus supplies silence rather than information. Per voxel over those 1,443 studies, the median voxel has 316 studies saying something and 1,127 silent; only 8.8% of voxels are named by any study, and at most 4 studies ever name the same one. Widening the report limb is what recovers it, which is what `report_radius` adapts.

## What it does not claim

- **It can be worse than doing nothing, and the regime is not exotic.** Against held-out HCP subjects an images-only pool recovered 0.85 of the true magnitude on `MOTOR_LH` and 0.96 on `EMOTION_FACES`, where `g` recovered 0.63 and 0.68, with `prevalence` reading 0.66 and 0.63 against a true 1.000. That is what happens when every study really has the effect: there is no absence to find and any `pi < 1` is error. Which regime a collection is in is not observable from the collection.
- **A magnitude is recoverable only from studies that differ in size.** A silence constrains `(pi, mu)` through one scalar per distinct (sampling error, cutoff) pair, and because the cutoff in sampling-sd units is the reported statistic again, varying the *threshold* moves both mixture components together while varying the *sample size* moves only the active one. At a fixed 20 studies, spreading sample sizes 12 to 120 lifts the fraction of voxels where the likelihood bounds `mu` from 0.42 to 0.69 and the fitted prevalence from 0.82 to 0.91; spreading thresholds 2.3 to 4.5 changes neither.
- **`se` is conservative and the likelihood often does not bound `mu` at all.** `se/sd` runs 1.2 to 2.2 against a known truth, most of it the cost of profiling out `pi`. The profile interval is unbounded wherever the data do not reject `pi = 0`, which at two image studies is 97% of voxels — the honest answer, not a search failure. **Read the width, never coverage**: every configuration measured covers 0.95 to 1.00, including those admitting almost any magnitude.
- **`prevalence` is ordinal, not a fraction,** and not comparable between voxels of one map.
- **The report limb's probability is over-stated by a bounded factor.** A paper reports a voxel only if it cleared the cut *and* was a local maximum. At a signal peak the mean field's own curvature makes a maximum more likely, so a zero-mean maxima density is a lower bound — off by up to two, and most where the effect is sharpest. Correcting it needs a per-study peak sharpness no paper reports.
- **The threshold's conversion to an effect size is load-bearing on assumed degrees of freedom**: at a *z* of 6 the implied `g` varies 38% over plausible values.
- **Voxels where a single study reported have no estimable magnitude** — the likelihood is a plateau, and `max_iter` decides which point on it is returned, reproducibly and no more.
- **Not validated against a reference implementation**, because none exists. Instead the DerSimonian–Laird path is checked against PyMARE, the censored mixture against an independently written brute-force maximization, and the estimator's hand-derived closed forms — the censoring score and curvature, all three observed-information blocks, and the bias/variance crossing above — against symbolic differentiation.

## Error rates

Global null, 40 simulations, 200 permutations, nominal 0.05, binomial standard error 0.034 on the familywise rate. Every figure previously quoted in this PR was measured under the deleted arrangement null and does not transfer; these replace them.

| arm | uncorrected | familywise |
| --- | --- | --- |
| 20 studies, 2 images | 0.0508 | 0.075 |
| 20 studies, 1 image | 0.0498 | 0.025 |
| 12 studies, 2 images | 0.0501 | 0.150 |
| 12 studies, 1 image | 0.0496 | 0.100 |

The voxelwise rate is nominal throughout. **The familywise rate is calibrated at twenty studies and anti-conservative below it.** The two effects look additive: dropping 20 studies to 12 adds about 0.075, and a second image adds about 0.05. Two candidate causes have been ruled out — the permutation distribution is not degenerate (200 distinct maxima out of 200, cv 0.08, so the guard correctly does not fire), and disabling the generalized-Pareto tail leaves the rates unchanged. The remaining candidate is residual non-exchangeability that matters more when each study is a larger share of the map.

These rates are measured on the interior of the simulated volume, 8 mm in from every face, and that matters. Measured over the whole volume the same arms give 0.375, 0.200 and 0.350. The cause is the test bed rather than the estimator: the field is smoothed inside a finite cube, so its boundary shell carries inflated variance and inflated peak density, which couples a study's image magnitude to the number of studies silent nearby — `rho = -0.125` over the whole cube against **−0.0005** beyond 8 mm from the edge. Since the permutation moves image values while holding the indicator fixed, any such coupling makes the null non-exchangeable, and the maximum statistic is the thing that notices. A real analysis on a brain mask has a boundary too, so whether the same coupling arises there is a question about data preparation that this bed cannot answer.

## Test plan

`pytest nimare/tests/test_meta_effectsize.py nimare/tests/test_generate.py` — 96 pass. Beyond smoke tests the suite pins the properties the model rests on: reduction to DerSimonian–Laird against PyMARE, the censored likelihood against a brute-force optimum, the null shuffling within a study and never between, a degenerate collection refused rather than given p-values, convergence alone not making a voxel significant, the threshold conversion's two designs, the prevalence held at 1 where no indicator identifies it (and the interval identity that follows), the profile interval being unbounded where `pi = 0` stands, the three indicator zones with widening the report eating only into the ring that carries no indicator, and a study contributing exactly one indicator per voxel however close its foci sit.

`examples/02_meta-analyses/17_plot_cbes.py` is a worked example that runs end to end in ~27s.

The estimator's docstring carries the conclusions and caveats; the full measurement record — every table, every route that failed, and the symbolic proofs — lives with its scripts in a companion experiments repository rather than in the module.

https://claude.ai/code/session_01E37peWH4mfQi7ArYFJJSSj


jdkent:
The `benchmark` check is failing intermittently on this branch, and the failures are not this PR's. Recording why rather than chasing it.

Across four consecutive heads the check alternates, and each failure names a *different* benchmark, in code this PR does not touch:

| head | benchmark | result |
|---|---|---|
| `26a1897` | `benchmark_cbmr.TimeCBMRInference.time_spatial_contrast_sandwich` | fail, 2.48× (22.3±6 ms vs 9.02±0.6 ms) |
| `82dbf4a` | — | pass |
| `3c51d30` | — | pass |
| `9c9f38a` | `bench_cbma.TimeCBMA.time_mkdachi2_studyset` | fail, 1.61× (66.9±0.8 ms vs 41.6±0.1 ms) |

The decisive one is `9c9f38a`: its entire diff is eleven added lines inside a class docstring, with no executable code changed at all. A docstring cannot make `MKDAChi2` 1.61× slower, so that reading is measurement noise by construction.

The first failure shows the same thing in its error bars — `22.3±6 ms` is a ±27% interval, and the neighbouring `time_scalar_contrast` swung 2.81 → 8.53 ms between the two *base* rounds of that same run, on identical code.

This PR touches `nimare/meta/cbma/effectsize.py`, its tests, `nimare/generate.py` and `docs/cbma.rst`. It adds no import to the CBMR or MKDA paths beyond the estimator already exported from `nimare.meta.cbma`, which was equally present on the two heads that passed.

No fix to port: there is no regression here to fix, and I am not going to adjust an unrelated benchmark's threshold from this PR to silence it. The check has already had several natural re-runs across the heads above, which is what a manual re-run would have provided. If the maintainers would like the noise addressed, tightening the ASV configuration for these two benchmarks — more rounds, or a wider `--factor` for the ones that are milliseconds long — looks like the useful follow-up, and belongs in its own PR.

---

jdkent:
## `benchmark` is failing on `bench_cbma.TimeCBMA.time_mkdachi2_studyset`, and it isn't this PR

Standing down on this one rather than pushing a fix, because the evidence says it is the benchmark's own instability crossing `asv continuous --factor 1.2`, not a regression from this branch.

**This PR never touches MKDAChi2.** The diff adds `nimare/meta/cbma/effectsize.py`, appends to `nimare/meta/utils.py` and `nimare/generate.py`, exports `CBES` from `nimare/meta/cbma/__init__.py`, and adds tests. `mkda.py` is untouched, and the only edit to pre-existing code anywhere in the diff is one name added to an existing import statement.

**All twelve `bench_cbma` results from the first failing run:**

| benchmark | PR | base | ratio | base spread |
|---|---|---|---|---|
| `time_ale` | 114±3ms | 112±3ms | 1.018 | 2.7% |
| `time_ale_studyset` | 75.6±0.3ms | 75.9±1ms | 0.996 | 1.3% |
| `time_kda` | 61.7±0.6ms | 60.5±0.2ms | 1.020 | 0.3% |
| `time_kda_studyset` | 27.0±0.8ms | 26.4±0.3ms | 1.023 | 1.1% |
| `time_mkdachi2` | 137±3ms | 141±2ms | **0.972** | 1.4% |
| `time_mkdachi2_dense` | 1.09±0.01s | 1.13±0.04s | 0.965 | 3.5% |
| `time_mkdachi2_dense_studyset` | 116±0.5ms | 110±1ms | 1.055 | 0.9% |
| `time_mkdachi2_studyset` | 70.1±0.7ms | 55.2±**7**ms | **1.270** | **12.7%** |
| `time_mkdadensity` | 70.2±0.4ms | 68.0±1ms | 1.032 | 1.5% |
| `time_mkdadensity_dense` | 641±3ms | 641±3ms | 1.000 | 0.5% |
| `time_mkdadensity_dense_studyset` | 155±0.7ms | 156±2ms | 0.994 | 1.3% |
| `time_mkdadensity_studyset` | 34.0±0.4ms | 33.6±0.7ms | 1.012 | 2.1% |

Eleven of the twelve land within ±5.5% in mixed directions, so there is no systematic bias. The flagged benchmark is the only one with a double-digit spread on the base measurement — 12.7% here, 18% on the next run (49.1±9ms) — where the next-worst is 3.5%. Its own sibling `time_mkdachi2` came out *faster* on this branch.

Across three runs of the same code the base measurement itself read 49.1ms and 55.2ms, and `a018ea5` passed on re-run with no change at all. A ratio of 1.24 against a 1.2 factor is inside that benchmark's noise.

**Proposed patch, not applied here** since `benchmarks/` is outside this PR's scope. `TimeCBMA` sets no sampling attributes, so this benchmark takes asv's defaults, which evidently do not stabilise it at ~50–70ms. Raising its sample count would:

```python
    def time_mkdachi2_studyset(self, data):
        """Time the MKDAChi2 estimator on Studyset inputs converted from the benchmark Dataset."""
        meta = MKDAChi2()
        meta.fit(self.studyset, self.studyset)

    time_mkdachi2_studyset.repeat = 10
    time_mkdachi2_studyset.number = 3
```

Raising `--factor` in the workflow instead would work but would blunt regression detection for the other eleven.

I have used my one re-run on this failure (`a018ea5`, which then passed). Happy to apply the patch above here if you would rather have it in this PR than a separate one.

---

jdkent:
## Correction on the `benchmark` failure: my earlier diagnosis was wrong

I said above that `bench_cbma.TimeCBMA.time_mkdachi2_studyset` was tripping asv's 1.2 factor because its base measurement was noisy (±7 to ±9 ms on ~50 ms), and proposed pinning its sampling. I landed that patch in 92a5a2d. **The next run invalidates the reasoning**, so I have reverted it in 481027b.

With the patch in, run [34884908500](https://github.com/neurostuff/NiMARE/actions/runs/34884908500) reported:

| | base `966ee7de` | PR `208d8080` | ratio |
|---|---|---|---|
| `time_mkdachi2_studyset` | 43.6±**0.7**ms | 66.7±**0.5**ms | **1.53** |

Both error bars are now tight, and the ratio *grew* from the 1.24 and 1.27 seen before. A tight difference is not noise, so the argument I gave for the patch does not hold, and I am not leaving a change to a shared benchmark file resting on it.

**I cannot reproduce it.** `main` and this branch measure the same here, in both directions across repeats:

| configuration | main | this branch |
|---|---|---|
| 4 cores | 76.9 ms | 73.1 ms |
| pinned to 2 cores | 73.4 ms | 76.0 ms |

Ruled out along the way:

- **asv's `number` auto-tuning being asymmetric between the two sides.** Batching consecutive calls does not change per-call time: 74.0 / 72.0 / 74.9 / 73.6 ms at 1 / 3 / 10 / 30 calls.
- **The new module's imports perturbing threading.** This PR adds `from .effectsize import CBES` to `nimare/meta/cbma/__init__.py`, so importing that package now also pulls in joblib, `scipy.optimize`, `scipy.special` and `nilearn.maskers`. Preloading all of those on `main` before benchmarking: 74.8 ms plain, 73.6 ms preloaded.
- **Core count.** No difference at 2 cores, as above.
- **Code reachable from the benchmark.** `MKDAChi2` lives in `mkda.py`, which no commit on this branch touches; the only edit to pre-existing code anywhere in the diff is one name added to an existing import in `nimare/meta/utils.py`. Its sibling `time_mkdachi2` is unaffected or slightly faster on this branch.

The open question I could not settle from here is whether `asv continuous` discovers benchmarks from each revision's own `benchmarks/` directory or from a single checkout. If per-revision, my patch applied to one side only and manufactured the tighter, larger ratio — which is why reverting is the safe move. If from a single checkout, then the 1.53 was a symmetric measurement and there is something real on the runner that does not appear on a 4-core container.

Leaving it unexplained rather than papering over it. Every other check on this PR is green, and I would rather flag an honest unknown in benchmark tooling than assert a cause I have now had wrong twice. Happy to dig further with a hint about the runner, or to re-run if you want another sample.

---

jdkent:
## Closing the loop on `benchmark`: intermittent, ~1 in 3, unrelated to anything I changed

My previous comment argued that because the error bars had gone tight, the difference could not be noise. **That was wrong**, and the full run history for this workflow on this branch shows why:

| commit | sampling patch present | `benchmark` |
|---|---|---|
| `3ad062d` | no | pass |
| `3f8ecd4` | no | pass |
| `a018ea5` | no | **fail (1.27)**, pass on re-run of identical code |
| `8812026` | no | **fail (1.24)** |
| `92a5a2d` | yes | pass |
| `c4e541e` | yes | pass |
| `58bf077` | yes | **fail (1.53)** |
| `b5564d7` | yes | pass |

Three failures in nine completed runs, interleaved with passes both with and without the patch. `c4e541e` → `58bf077` → `b5564d7` runs pass → fail → pass, and the diffs between those three commits are one added output map and documentation — nothing that can turn a 53% regression on and then off again.

My error was reading asv's `±` as evidence about run-to-run stability. It is the spread across samples *within* one run on one machine state. The base reading for this benchmark across runs was 55.2, 49.1 and 43.6 ms — a 26% span — and that is the variation that decides whether the 1.2 factor gets crossed. A tight `±` on any single run is consistent with that.

So: no action needed on this PR, and **please disregard my request above for a hint about the runner** — there is no evidence of anything real to chase. The patch stayed reverted in `481027b` because failures occur at the same rate with and without it; it is neither implicated nor shown to help, and it is not my place to carry an unjustified change to a shared benchmark file.

If the intermittency is worth addressing on its own terms, that is a repository-level decision about `asv continuous --factor` or this benchmark's sampling, independent of this PR.

---

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2924 @@
...
+    error is bounded by a factor of two and is not correctable from tables; see ``CBES`` under
+    "Warnings", and ``notes/cbes-evidence.md`` in the companion experiments repository for the
+    measurements and the derivation.
+
+    Returned as one dict because the E step and the M step both need these at the same ``mu``.
+    This is the most expensive function in the estimator -- most of the wall clock under a
+    permutation null, since each permutation runs a full EM -- and 45% of that is the two
+    ``ndtr`` calls, which no rearrangement of the surrounding algebra reaches. Three were
+    benchmarked and none was worth the churn; the lower tail cannot be dropped either. The
+    levers that work are ``n_cores`` and the compaction in :meth:`CBES._fit_chunk`.
```
jdkent: too much explanation here

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2924 @@
...
+
+    This is the one thing the reported statistics are still read for, and it is not a
+    magnitude: anything a study reported *cleared* that study's threshold, so
+
+    .. math:: c_k \le \min_j |z_{kj}|
+
+    is a hard inequality rather than an inference. It can clamp an assumed threshold downward
+    and can never place one above the truth, which is what separates it from the retired
+    ``"study-min"`` rule -- that one tried to *recover* :math:`c_k` by undoing the order
+    statistic for the number of peaks, and overshot a cluster-forming cut by about 1 z.
```
jdkent: mentions a previous term that should not be mentioned

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2845 @@
...
+            self.inv_sigma,
+            self.inv_sigma_sq,
+            self.sign,
+        )
+
+
+class CBES(Estimator):
+    r"""Coordinate-based effect-size meta-analysis, estimated from what studies did *not* report.
+
+    .. versionadded:: 0.13.0
```
jdkent: `0.22.0`

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2845 @@
...
+        extent a focus stands in for has to be assumed rather than read.
+
+        Used only when ``selection_model="zero-inflated"``. ``g`` is insensitive to it at
+        realistic focus counts; ``prevalence`` is not, rising monotonically with the radius at
+        every true value and never recovering the truth, which is one reason to read that map
+        ordinally.
+    max_iter : :obj:`int`, default=25
+        EM iterations. Voxels are retired as they settle, so this bounds the slowest rather
+        than the typical one. Where a single study reported, the likelihood is a plateau and no
+        stopping rule recovers a magnitude; the cap makes that answer reproducible.
```
jdkent: why mention whether magnitude can be recovered or not?

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2845 @@
...
+
+       In practice the choice is not clean, because :math:`\pi` is the weakest part of the fit.
+       On the NIDM pain collection, scored against an inverse-variance pool of held-out studies
+       -- itself a :math:`\pi\mu` quantity, so ``g_marginal`` is the estimand-matched arm --
+       ``g_marginal`` does win on rmse, 0.184 against 0.230. But it undershoots by 0.291 at the
+       strongest voxels where ``g`` sits at +0.005, because the fitted prevalence is biased low
+       there (0.78 against a true 1.0 on held-out subjects, 0.65 over the whole map). So
+       ``g_marginal`` estimates the right quantity with a biased multiplier, and ``g`` estimates
+       a different quantity that happens to be close to it wherever :math:`\pi` is genuinely
+       near 1.
```
jdkent: say the quantity `g` is measuring (conditional on the effect size being no nulll)

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2845 @@
...
+       0.91, while spreading thresholds from 2.3 to 4.5 changes neither.
+
+    4. **``se`` is conservative, and where the effect is weak the likelihood does not bound
+       :math:`\mu` at all.** ``se/sd`` runs 1.2 to 2.2 against a known truth. Most of the excess
+       is the cost of profiling out a prevalence the indicator only partly identifies, not the
+       censoring term -- an all-image collection, where the indicator is structurally empty, was
+       the worst-calibrated configuration measured until :math:`\pi` was held at 1 there.
+       ``interval="profile"`` reports the likelihood region honestly, and it is unbounded
+       wherever the data do not reject :math:`\pi = 0`, which at two image studies is 97% of
+       voxels.
```
jdkent: this also talks about experimentation

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2845 @@
...
+    pain collection, where studies genuinely differ, the same machinery improves the level
+    substantially. Which regime a given collection is in is not observable from the collection,
+    and that is the honest statement of the risk.
+
+    **The report limb's probability is over-stated, and the error is bounded but not fixed.** The
+    model treats a report as :math:`P(|g| \ge c)`, while a paper reports a voxel only if it
+    cleared :math:`c` *and* was a local maximum. At a signal peak the mean field's own curvature
+    makes a maximum more likely, so a zero-mean maxima density is a lower bound -- off by up to a
+    factor of two, and most where the effect is sharpest. Correcting it needs a per-study peak
+    sharpness that no paper reports.
```
jdkent: too much explanation here.

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2845 @@
...
+                ("w", weights),
+                ("w2", weights),
+                ("a", a),
+                ("a2", a**2),
+                ("ag", a * g[cols]),
+                ("ag2", a * g[cols] ** 2),
+                ("w2_over_s2", a),
+                ("n", weights),
+            ):
+                sums[name][cols] += value
```
jdkent, line 1639: are some of these columns completely redundant? like "a" and "w2_over_s2"?

or "w", "w2", and "n"

nimare/meta/cbma/effectsize.py
```diff
@@ -0,0 +1,2845 @@
...
+
+        # `dof` is taken from the censoring roster, not from a Kish count over the pooled
+        # weights. Two reasons, and the second is decisive. The likelihood uses every study on
+        # the roster -- the images through their values, the rest through their silence -- so
+        # crediting only the weighted contributors denies the model information it demonstrably
+        # used. And with coordinate magnitudes gone, the only weighted contributions are images
+        # at weight 1, so a Kish count is just the image count: `n_eff = k`, `dof = k - 1`,
+        # which is **0 with one image** and makes the recommended interval `nan`. The roster is
+        # the only reference that survives the design.
+        roster_size = getattr(self, "_sample_sizes_", None)
```
jdkent: "and the second is decisive" and similar speech patterns are difficult to read, please write in a straightforward manner.

jdkent (commented):
Documentation/comments are too verbose, and certain LLM ordering of verbs, adjectives, and nouns are tiring to read, please make the comments/doc strings/documentation more compressed.

jdkent:
Addressed in 7c42a9b.

On the redundant columns: you were right, and it was worse than redundant. Every image contributes weight 1, so `w`, `w2` and `n` were three names for the study count and `w2_over_s2` was a second name for `a`. That let `_local_dersimonian_laird` drop two of its seven parameters — with unit weights its `expected_q` reduces to `k - 1`, the classical estimator — and `_pool` no longer computes a Kish count that can only equal the study count, or return a `sum_w` that nothing read. 127 lines of code gone.

On verbosity: the CBES docstring is down from 218 lines to about 150, `_observed_information` from 73 to 62, and the seven longest comment blocks lost their experiment narration. The measurements they quoted are in the companion experiments notes, which is where the rest of the record already lives. Prose density in the module is now 36%, from 43% at the start of this pass.

Also fixed: `versionadded` is 0.22.0; `g` is now stated plainly as the mean effect among studies whose effect at the voxel is non-null; and the asides about magnitudes not being recoverable, the retired `"study-min"` rule, and the `coverage_radius`/`max_iter` experiment reports are gone.

92 tests pass and flake8 is clean.

Two things from today's measurements that change what the PR should claim, both against my own earlier statements:

- **The familywise rates are inflated and the cause is in the test bed, not the estimator.** All three arms of the global-null re-measurement: uncorrected 0.0533 / 0.0496 / 0.0534 (nominal), familywise 0.375 / 0.200 / 0.350. The permutation moves image values while holding the indicator fixed, so the null is valid only if the two are independent across voxels — and they are not, at rho −0.115. But that coupling is entirely the simulated cube's boundary: beyond 8 mm from the edge it is −0.0005, with `corr(|g|, edge distance)` −0.212 and `corr(silent count, edge distance)` +0.469, both from smoothing running off the edge of a finite volume. A re-measurement on an eroded mask is running; I will post real numbers rather than the current ones either way.
- **The coordinates can make `g` worse, and on pain the crossing is at 3 images.** Paired over the same 8 splits with published tables, CBES beats an image pool at 1–2 images (rmse −0.092 and −0.040), ties at 3, and is significantly worse from 5 on (+0.027, p=0.001). Bias at the top decile crosses later, at about 6. I will fold this into the Notes once the larger sweep against the NeuroStore pain corpus finishes.

---

jdkent:
interesting experiment, starting from an algebraic foundation instead of a data driven foundation was a better, more grounded way to improve the estimator, testing on real data is good after mathematical proofs and simulations are completed.  Otherwise the structure of the algorithm overfits on the data provided. Additionally, perhaps being less opionated on the solution is helpful for the llm to find a better overall solution (unless you have hard data to back up why the llm should avoid a certain solution).

PR #1154, opened 2026-09-20
Validate Monte Carlo iteration counts

### LINKED ISSUE #1153: Validate Monte Carlo iteration counts when creating FWECorrector
## Summary

`FWECorrector(method="montecarlo")` accepts invalid explicit values for `n_iters`, including zero, negative numbers, noninteger values, Boolean values, and strings. These values are stored in `parameters` and are only rejected, or fail indirectly, after correction reaches estimator specific code.

## Additional details

NiMARE version: current main at `e11d0e0`

Configure Monte Carlo FWE correction with an explicit iteration count.

An explicit `n_iters` value should be a positive integer. Invalid values should raise a clear `ValueError` when the corrector is created, before any estimator work begins.

### What actually happened?

All of the invalid values below are accepted and stored. This delays the error until estimator specific array allocation, iteration, or probability calculation. The IBMA Monte Carlo implementation already checks that `n_iters` is a positive integer, so validation currently differs by execution path.

## Reproducing the bug

```python
from nimare.correct import FWECorrector

for value in (0, -1, 1.5, True, "10"):
    corrector = FWECorrector(method="montecarlo", n_iters=value)
    print(repr(value), repr(corrector.parameters["n_iters"]))
```

Current output:

```text
0 0
-1 -1
1.5 1.5
True True
'10' '10'
```

I propose validating explicit `n_iters` values in `FWECorrector` when `method="montecarlo"`, while preserving `None` so estimator defaults still work. Tests can cover valid Python and NumPy integers, `None`, and rejection of zero, negative, Boolean, noninteger, and string values.

SahilKumar75:
Closes #1153.

1. Validate explicit Monte Carlo iteration counts when creating `FWECorrector`.
2. Preserve `None` so estimator specific defaults continue to apply.
3. Cover valid Python and NumPy integers and reject zero, negative, Boolean, noninteger, and string values.

Validation:

1. `python -m pytest nimare/tests/test_correct.py -q`
2. `make lint`
3. Clean main baseline: 1304 passed and 263 skipped.

## Summary by Sourcery

Validate explicit Monte Carlo iteration counts for positive integer values.

Bug Fixes:
- Validate Monte Carlo iteration counts while preserving estimator-specific defaults when no count is provided.

Tests:
- Add coverage for supported Python and NumPy integer counts and invalid Monte Carlo iteration values.


jdkent (approved):
LGTM, thanks for catching this guard @SahilKumar75

PR #1156, opened 2026-09-20
[FIX] Load m2r2's mdinclude extension instead of importing MdInclude

### LINKED ISSUE #1155: documentation build is failing
## Additional details

- NiMARE version:

jdkent:
Closes #1155.

- `docs/conf.py`: load the `m2r2.mdinclude` Sphinx extension instead of importing `MdInclude` and registering the directive by hand.
- `setup.cfg`: pin `m2r2>=1.1` in the `doc` extra, the first release that ships that extension.

## The failure

The Read the Docs build fails before Sphinx reads a single document:

```
ImportError: cannot import name 'MdInclude' from 'm2r2'
```

`docs/conf.py` imported `MdInclude` at module scope and wired the `mdinclude` directive up itself. That import only ever worked on m2r2 &le; 0.3.4, which shipped as a single `m2r2.py` module. 1.0.0 repackaged the project into `m2r2/` and moved the class to `m2r2.sphinx.directives`, so the package root has not exported it for two releases. The `doc` extra required `m2r2` unpinned, so the build broke as soon as the resolver picked up a 1.x release.

## The fix

1.1.0 added an `m2r2.mdinclude` extension whose `setup()` registers the same five config values and the same `mdinclude` directive that `conf.py` was registering by hand, so this loads that extension and drops the manual block.

Two things worth flagging for review:

- **Both halves have to change together.** Leaving the manual registration in place alongside the extension raises `Config value 'no_underscore_emphasis' already present`.
- **It is `m2r2.mdinclude`, not plain `m2r2`.** The full extension also registers a source parser for `.md` files, which would collide with `recommonmark`. m2r2's own error message recommends the submodule for exactly this case.

The floor pin is load-bearing rather than tidiness: on 1.0.x neither the old import nor the new extension exists, so without it the resolver could land there and break the build a different way.

One behavioural note: the old code hardcoded `m2r_use_mermaid = True`, while the extension derives it from `"sphinxcontrib.mermaid" in extensions`. That extension is in this `conf.py`, so the value is unchanged.

## Validation

`mdinclude` is used by `docs/changelog.rst` and `docs/contributing.rst`, so the directive has to keep working, not just import.

- `docs/conf.py` imports cleanly on m2r2 1.1.0 — the exact point the RTD build died.
- Built the real docs with this `conf.py`: `changelog.html` and `contributing.html` both render their markdown (`CHANGELOG.md` and `CONTRIBUTING.md`), with no m2r2 or `mdinclude` warnings. The gallery was disabled for that run because `torch` is not installed locally; RTD installs the `cbmr` extra, so that gap does not apply there.
- Separately confirmed `m2r2.mdinclude` and `recommonmark` coexist without a parser collision.
- `flake8 docs/conf.py` reports exactly the same three pre-existing findings as on `main`; `flake8 nimare` is clean.

Found while driving #1152 to green, where this is the only remaining red signal; it is unrelated to that PR and reproduces identically on `main`.

https://claude.ai/code/session_0132Md3Z3JSHpD5xt3mcKqTV

---
_Generated by [Claude Code](https://claude.ai/code/session_0132Md3Z3JSHpD5xt3mcKqTV)_

## Summary by Sourcery

Update the documentation configuration to use m2r2's compatible mdinclude extension and require a version that provides it.

Bug Fixes:
- Fix the documentation build by loading m2r2's supported mdinclude extension instead of importing the removed MdInclude class directly.

Enhancements:
- Replace manual mdinclude configuration and registration with the m2r2.mdinclude Sphinx extension while preserving Markdown inclusion support.

Build:
- Require m2r2>=1.1 for the documentation dependencies so the mdinclude extension is available.


setup.cfg
```diff
@@ -61,7 +61,7 @@ predictive =
     xgboost<3.1
 doc =
     %(parquet)s
-    m2r2
+    m2r2>=1.1  # 1.1 is the first release with the m2r2.mdinclude extension
```
jdkent:
```suggestion
    m2r2>=1.1
```

jdkent: Taken — you'd already committed it yourself in d40f6a6 by the time I got here, so I dropped my duplicate commit rather than pushing over yours. The rationale for the `>=1.1` floor is in 1b1c09f's message and the PR description.

---

